PR #50, opened 2014-10-20
Fix bug where dates are several days in the future

rculbertson:
Docker returns timestamps with nanosecond precision (e.g.
2014-10-17T21:22:56.949763914Z), but the Date object only supports
milliseconds. Creating a Date from the nanosecond timestamp results
in the date being set to several days after what date should be.
This class converts the timestamp from nanoseconds to milliseconds
by removing the last six digits of the timestamp, so we can generate
a Date with the correct value (albeit with less precision).

Note: a more complete solution would be to introduce a custom date
type which can store the nanosecond value in an additional field, so
users can access the complete value. Or just use Java 8 which has
date objects with nanosecond support.

### DIFF
--- a/findbugs-exclude.xml
+++ b/findbugs-exclude.xml
@@ -5,4 +5,10 @@
     <Bug pattern="SE_BAD_FIELD"/>
   </Match>
 
+  <Match>
+    <Class name="com.spotify.docker.client.DockerDateFormat" />
+    <Method name="clone" />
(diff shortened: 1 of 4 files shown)


KatalinaKhalifa:
I have a friend who added a set of songs around 2 years ago: 
starred_9_years. 
They are still showing up as if they were added 9 years into the future!
Also - he sent one or two of these songs to my inbox, everyday they are showing up as just sent today. This causes another glitch where when he does send music to my inbox - I get no alert. Frustrating!

PR #103, opened 2014-12-12
Fix #98: Use jnr-unixsocket instead of junixsocket.

### LINKED ISSUE #98: Usage of jnr-unixsocket for unix socket support
I would like to propose changing docker-client's unix socket support from using unix-socket-factory (junix-socket) to jnr-unixsocket [1]. The latter is more actively developed and also used in JRuby.

Currently it seems docker-client requires de.gesellix:unix-socket-factory but only uses a single bundled class (org.newsclub.net.unix.AFUNIXSocket) from junixsocket (likely since this provides the classes through a mavenized resource). The version that is bundled is listed as 1.4 [2], but the project's latest release was 1.3 [3] so it's unclear how to reproduce it.

NOTE: There are some additional changes that I would need to propose and push into jnr-unixsocket to support some necessary socket options. Once they're completed, I could create a pull request.

The changes can be found at :
https://github.com/rgrunber/docker-client
https://github.com/rgrunber/jnr-unixsocket

[1] https://github.com/jnr/jnr-unixsocket
[2] https://github.com/gesellix/unix-socket-factory/tree/master/libs
[3] https://code.google.com/p/junixsocket/

rgrunber:
A few notes :

- I've run the tests and noticed intermitent failure of testSsl locally for me with or without this change.
- On the test instance, testContainerWithHostConfig also seemed to have failed for me, but passed on a re-trigger.
- commons-lang:commons-lang:2.4 was already an indirect dependency brought in by the unix-socket-factory artifact and it is in fact a build-time dependency for DefaultDockerClientTest so it should be listed.
- All of the classes marked unimplemented do not seem to be used by the underlying API currently.

### DIFF
--- a/src/main/java/com/spotify/docker/client/ApacheUnixSocket.java
+++ b/src/main/java/com/spotify/docker/client/ApacheUnixSocket.java
@@ -21,10 +21,6 @@
 
 package com.spotify.docker.client;
 
-import com.google.common.collect.Queues;
-
-import org.newsclub.net.unix.AFUNIXSocket;
-
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -32,11 +28,17 @@ import java.net.InetAddress;
 import java.net.Socket;
 import java.net.SocketAddress;
 import java.net.SocketException;
+import java.nio.channels.Channels;
 import java.nio.channels.SocketChannel;
 import java.util.Queue;
 
+import jnr.unixsocket.UnixSocketAddress;
+import jnr.unixsocket.UnixSocketChannel;
+
+import com.google.common.collect.Queues;
+
 /**
- * Provides a socket that wraps an org.newsclub.net.unix.AFUNIXSocket and delays setting options
+ * Provides a socket that wraps an jnr.unixsocket.UnixSocketChannel and delays setting options
  * until the socket is connected. This is necessary because the Apache HTTP client attempts to
  * set options prior to connecting the socket, which doesn't work for Unix sockets since options
  * are being set on the underlying file descriptor. Until the socket is connected, the file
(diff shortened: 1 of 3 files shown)


src/main/java/com/spotify/docker/client/ApacheUnixSocket.java
```diff
@@ -47,75 +49,78 @@
...
-    setAllSocketOptions();
+    if (endpoint instanceof UnixSocketAddress) {
+      inner.connect((UnixSocketAddress) endpoint);
+      setAllSocketOptions();
+    }
   }
 
   @Override
   public void connect(final SocketAddress endpoint, final int timeout) throws IOException {
-    inner.connect(endpoint, timeout);
```
davidxia: @rohansingh Did we accidentally stop setting `timeout` with this PR change?

rgrunber: Looks like it's definitely not being set on the initial connection. Looking at it now, we could certainly call a setSoTimeout(int timeout) assuming that would do the job as expected. In fact, looking in jnr-unixsocket, It seems I added support for this ( https://github.com/jnr/jnr-unixsocket/commit/7f3ee30ff762d7bad5a441e0461bf74d494880e2 ), presumably so that this class would be able to continue making those calls.

Despite the call not being done on creation, I'm pretty sure that from my testing of pull #916 , both TCP and UNIX sockets would stop working when the read timeout was exceeded so it seems they're respecting it now. Although I also recall times in the past when timeouts were not respected at all on the UNIX socket.

Is there test case / set of calls that show UNIX sockets still don't fully respect the timeout ?

rohansingh: @davidxia Sorry it's way too long ago, I barely remember the context.

davidxia: @rgrunber I'm not really that familiar with jnr. I just saw that the `timeout` arg wasn't being used at all in this method. Does the timeout get magically set somewhere else?

rgrunber: Looks like I'm confusing connection timeout and socket timeout. Connection timeout is the maximum amount of time to wait for a connection to be established with the daemon. Socket timeout is the maximum amount of inactivity between 2 packets.

The timeout passed in to the connect() call is for connection timeout (value was 5000 ms), and you're right, it wouldn't be set anywhere else. I wonder if we could just call setSoTimeout initially with that value and have the socket behave the same. Eventually there are other calls to setSoTimeout done by Apache HttpComponents itself based on the RequestConfig values set in DefaultDockerClient. All of those calls are either for a timeout of 0 (no timeout client) or 30000 (30s timeout client).

davidxia: Is this issue now moot with these changes? https://github.com/spotify/docker-client/pull/918

rgrunber: It's definitely a step in the right direction. Assuming all tests are working as expected with that change, then yes, that commit should resolve the issue.

PR #106, opened 2014-12-16
docker exec command

relgames:
Implemented exec command introduced in Docker 1.3

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -880,8 +880,8 @@ public class DefaultDockerClient implements DockerClient, Closeable {
 
     String response;
     try {
-      response = request(POST, String.class, resource, resource
-              .request(APPLICATION_JSON_TYPE),
+      response = request(POST, String.class, resource,
+              resource.request(APPLICATION_JSON_TYPE),
               Entity.json(writer.toString()));
     } catch (DockerRequestException e) {
       switch (e.status()) {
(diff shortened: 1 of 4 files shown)


rohansingh:
@relgames I think it's fine to have the test aware of the what execution driver is being used. In fact, it's a good signal for anybody who uses this method that they need to do the same thing and check the exec driver before assuming that exec will work.

relgames:
Added API version and driver checks. API at least 1.15 is required and "native" driver.

davidxia:
@relgames 3 tests failed. The 2 tests that are not testExec will be fixed (hopefully) if you rebase onto the latest master.

davidxia:
@relgames Hmm, maybe skip testExec() if we detect that the host doesn't have the native libcontainer driver? Otherwise the test will [fail here](https://github.com/spotify/docker-client/pull/106/files#diff-8b9a5e049d8e8be5e08ee5cc7526354fR1081).

JWGmeligMeyling:
Isnt it so that Assume statements (as suggested) ignore the test if it fails? (Instead of Asserts)

relgames:
Rebased, changed assertThat into assumeThat.

relgames:
Not sure why it still fails:
````
Tests in error: 
  testContainerWithHostConfig(com.spotify.docker.client.DefaultDockerClientTest): 
Request error: POST unix://localhost:80/v1.12/containers/0480cce9f36ae7279db70c4db014c31c1dd84a1c3c9857b74d277c8940ea71ce/kill: 500
````
But I did not touch that part.

relgames:
Well after minor cosmetic change it is fixed. I guess, something was/is wrong with your CI environment...

davidxia:
:+1: CircleCI fails from time to time...

PR #111, opened 2014-12-21
Added checkstyle rules to better conform to google style

rculbertson:


rculbertson:
@davidxia yea a checkstyle error will cause the build to fail

PR #118, opened 2015-01-29
Allow user to set auth headers per pull

davidxia:


src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -657,6 +657,35 @@ public void pull(final String image, final ProgressHandler handler)
   }
 
   @Override
+  public void pull(String image, AuthConfig authConfig)
```
rohansingh: params should be `final`

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -657,6 +657,35 @@ public void pull(final String image, final ProgressHandler handler)
...
+      throws DockerException, InterruptedException {
+    pull(image, authConfig, new LoggingPullHandler(image));
+  }
+
+  @Override
+  public void pull(String image, AuthConfig authConfig, ProgressHandler handler)
```
rohansingh: params here should be `final` too

rohansingh:
:+1: *

*minor comments inline

davidxia:
@rohansingh yay tests finally passed, try locally too for good measure?

rohansingh:
All tests pass locally for me.

Works On My Machine award

PR #123, opened 2015-02-16
Modified pom.xml file to allow for separate shaded and non-shaded artifacts

matthewwerny:
1. Added remarks to README.md to indicate the change in the dependency structure and artifact contents to be generated.
2. Altered pom.xml to allow for separate shaded and non-shaded artifacts

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,11 +100,28 @@ is exhausted and it takes too long to acquire a new connection for a request, we
 Maven
 -----
 
+Please note that in releases 2.7.6 and earlier, the default artifact was the shaded version.  When upgrading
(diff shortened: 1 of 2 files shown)


davidxia:
This seems like a reasonable thing. @rohansingh?

matthewwerny:
One minor note to consider.  When I created the PR, I used version 2.7.7 in the readme.md.  It'll be good to make sure that if any releases go out prior to this PR getting approved that that readme.md is updated to be correct.  I will happily amend this PR should that eventuality and need arise.

rohansingh:
This looks good. One thing I'd like to point out to users is that they will need the shaded version to avoid dependency conflicts if they are using Jersey 1.x. Let's just merge this for now though and I can update the README later.

Issue #124, opened 2015-02-18
API for monitoring Docker events


diptanu:
Does this client provide an API to monitor docker events? If not, I think we could do an efficient implementation with disruptor, where the user supplies the client with an event handler and on receiving an event from docker the client invokes the hander.

davidxia:
It doesn't have support for events but we'd definitely be open to a pull request :)

diptanu:
@davidxia Sounds good, let me see if i can pull some time out to look into this. 

Do you guys have an API that you would prefer for this? I guess we should talk about that first.

davidxia:
@diptanu Do you mean which Docker API version you should use?

davidxia:
@diptanu Is this still an issue or can I close it?

davidxia:
closing due to inactivity. We can reopen if necessary.

Issue #114, opened 2014-12-23
event support


mainiak:
Hi.

I have question regarding support for events inside docker host.

Is there or will be any support for docker events?
https://docs.docker.com/reference/api/docker_remote_api_v1.16/#monitor-dockers-events

I was looking into DockerClient API but I didn't seen any.

Thank you in advance for your answer.

mainiak:
@tedyoung - could you, please, paste here link to your implementation, so I can look into it?

tedyoung:
Oops, just realized that I totally dropped the ball on this. I'll sanitize my code and link to it, and then see what a PR might look like.

mattnworb:
done in #358

PR #127, opened 2015-03-01
Adds support for event callbacks to the docker client

colinmorelli:
This commit adds support for streaming events from the Docker API.

A few notes about the PR that kind of suck:

 - I wasn't able to find a way to abort a request in the Jersey Client, which makes it impossible to close a streaming request to the Docker API (the stream is terminated by the client, and simply closing the InputStream from Jersey does nothing as it internally just waits for an EOF). Long story short: I had to use the Apache HttpClient directly
 - The events API can either be streaming or not, depending on whether or not the "until" param is set (and, if it's set, what value it's set to)
 - The tests could be flaky, and may need to be addressed (although they're reliable on my machine, they do include dates which I don't like)

Open to suggestions on ways to clean this up

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -40,9 +40,6 @@ import com.spotify.docker.client.messages.Info;
 import com.spotify.docker.client.messages.ProgressMessage;
 import com.spotify.docker.client.messages.RemovedImage;
 import com.spotify.docker.client.messages.Version;
-import org.apache.http.HttpResponse;
-import org.apache.http.client.ClientProtocolException;
-import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
@@ -65,16 +62,6 @@ import org.glassfish.jersey.jackson.JacksonFeature;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.ws.rs.ProcessingException;
-import javax.ws.rs.WebApplicationException;
-import javax.ws.rs.client.Client;
-import javax.ws.rs.client.ClientBuilder;
-import javax.ws.rs.client.Entity;
-import javax.ws.rs.client.Invocation;
-import javax.ws.rs.client.ResponseProcessingException;
-import javax.ws.rs.client.WebTarget;
-import javax.ws.rs.core.GenericType;
-import javax.ws.rs.core.Response;
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -21,13 +21,12 @@
```
colinmorelli: I'll clean up the imports in this file back to what they were

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -865,6 +863,49 @@ public LogStream logs(final String containerId, final LogsParameter... params)
...
+        // We must URL encode the string, otherwise Jersey chokes on the double-quotes in the json.
+        final String encoded = URLEncoder.encode(writer.toString(), UTF_8.name());
+        resource = resource.queryParam("filters", encoded);
+      }
+    } catch (IOException exception) {
+      throw new DockerException(exception);
+    }
+
+    try {
+      CloseableHttpClient client = (CloseableHttpClient) ApacheConnectorProvider.getHttpClient(noTimeoutClient);
```
colinmorelli: This cast is safe. For whatever reason, ApacheConnectorProvider.getHttpClient downcasts to HttpClient despite its own internal property being a ClosableHttpClient

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -915,6 +916,73 @@ public void testContainerWithHostConfig() throws Exception {
...
+  }
+
+  @Test(timeout = 5000)
+  public void testEventStreamWithUntilTime() throws Exception {
+    EventStream eventStream = sut.events(DockerClient.EventsParam.until((new Date().getTime() + 2000) / 1000));
+
+    while (eventStream.hasNext()) {
+      eventStream.next();
+    }
+  }
```
colinmorelli: Since it isn't 100% clear what this test does, I'll explain here:

Basically, the Docker events API provides a way to end the streaming at a given timestamp. This method calls the streaming events API with a timestamp 2 seconds from now, but sets the test to timeout in 5 seconds. Basically, if the client doesn't pass the parameter to the API, the test will fail with a timeout (because it'll still be waiting for events to stream after 5 seconds). But if all is well, the Docker daemon should close the stream in ~2 seconds.

colinmorelli:
Addresses #124 #114

JWGmeligMeyling:
I see that one test is still failing. Could this be related to this: https://github.com/spotify/docker-client/pull/106 ? Also sometimes the CI just fails. Do your tests pass locally?

The issue with the Jersey client is related to [JERSEY-2162](https://java.net/jira/browse/JERSEY-2162)

colinmorelli:
The tests do pass locally most of the time (I have seen the occasional failure). However, the failures occur in more than just my added tests. Every time it fails, it's due to a 404 that is raised when attempting to kill a container (during cleanup), which doesn't make sense given that the cleanup method only kills containers that were just returned from the listContainers call. The tests themselves consistently run and pass.

I spent a good amount of time trying to figure out the Jersey client issue, but despite the fact that JERSEY-2162 appears to have been closed, I still don't believe this works correctly (or I'm just not figuring it out). Happy to change it to use Jersey if I can figure it out, though.

JWGmeligMeyling:
I also can't find a proof of it being fixed, it seems more like it's set to wontfix. But I'm not to much into Jersey client either.

mattnworb:
Sorry for the slow response @JWGmeligMeyling , but this change looks good to me - looks like it follows the same pattern as `LogStream` and friends and I don't see a huge issue with having to access the underlying Apache `CloseableHttpResponse` if Jersey gives you no better option.

Do you mind rebasing this on top of master to resolve the merge conflicts?

triggetry:
Any chance that this will make it into master any time soon? :-)

davidxia:
replaced by #358

PR #128, opened 2015-03-03
update copyright header

noaresare:
The previous copyright header was wrong, copied from some project covered by the Apache Foundation CLA

### DIFF
--- a/java.header
+++ b/java.header
@@ -1,13 +1,9 @@
 /*
  * Copyright (c) 2014 Spotify AB.
  *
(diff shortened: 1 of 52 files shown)


rohansingh:
Could you please also update the `java.header` file in the root directory? This is what's used by checkstyle.

noaresare:
I did, but due to a bug in my magic script I incidentally added a newline to the end of each fixed file. An updated file coming up

davidxia:
:+1:

PR #131, opened 2015-03-09
Appease findbugs - do not have an empty catch clause

davidxia:
Some versions of findbugs error out if we don't
do anything with this exception.


rohansingh:
:+1:

PR #133, opened 2015-03-10
Adding getters on Container.PortMapping

xcoulon:
Adding getters on the Container.PortMapping attributes. This is particularly interesting when clients need to consume the Port Mappings. In the current code, the only mean to extract the IP address, private/public ports and type are by parsing the result of the call to the Container.PortMapping#toString() method, which is a bit of a hack (and maybe not reliable in the long term) :-/

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/Container.java
+++ b/src/main/java/com/spotify/docker/client/messages/Container.java
@@ -155,6 +155,22 @@ public class Container {
     @JsonProperty("Type") private String type;
     @JsonProperty("IP") private String ip;
 
+    public String getIp() {
+      return ip;
+    }
+    
+    public int getPrivatePort() {
+      return privatePort;
+    }
+    
+    public int getPublicPort() {
+      return publicPort;
+    }
+    
+    public String getType() {
+      return type;
+    }
+    
     @Override
     public boolean equals(final Object o) {
       if (this == o) {


rohansingh:
Are you sure you need to consume the port mappings, and not the realized bindings for a running container? See #41 for example.

xcoulon:
@rohansingh 

Thanks for your reply. In our tooling, we want to display the list of all containers in table, in a similar fashion to what a user sees when he runs the 'docker ps -a' command.
Amongst all the containers I have on my docker host, one of them as for example the following port bindings (as retrieved via the 'docker inspect' command):
"Ports": {
            "27017/tcp": [
                {
                    "HostIp": "0.0.0.0",
                    "HostPort": "27017"
                }
            ]
        }  

But this is similar to what the PortMapping will give me when calling the Client#listContainers() method: PortMapping{privatePort=27017, publicPort=27017, type=tcp, ip=0.0.0.0}
Furthermore, I would prefer not having to send 'n + 1' queries to retrieve the list of containers and then inspect each one of them.

Am I missing something else ?
Thanks !

rohansingh:
Makes sense. Would you mind fixing the spacing to match our conventions and the rest of the file (Google Java code style) & squash? I'll go ahead and merge it.

xcoulon:
@rohansingh,
great ! I fixed the spacing in all the body of all the method I added and I rebased my branch. 
Sorry, I did not realize the initial additions did not match the coding convention :-/

xcoulon:
thanks, @rohansingh !

rohansingh:
@xcoulon Sure thing. Thank you for the contribution!

PR #134, opened 2015-03-12
Add operating system to Info message

rohansingh:
You can now do this:

    dockerClient.info().operatingSystem()


davidxia:
Let's also check this in the test. See https://github.com/spotify/docker-client/pull/135

rohansingh:
So the reason I didn't add a test for it is because this isn't returned until v1.2.

davidxia:
oh in that case :+1: 

On an unrelated note, should we state what versions of docker we actually test against?

rohansingh:
Yeah, we should and I think we should probably increase the minimum Docker version we support (at which point we could add that test). I'm guessing that we'd break against Docker v1.0 in actuality, but we claim to support it.

rohansingh:
By "claim to support it" I mean that the remote API version string makes it seem like we want to talk the Docker 1.0 API, even though we don't.

PR #140, opened 2015-03-17
Adding missing attributes on System-wide info from API 1.17

xcoulon:


src/main/java/com/spotify/docker/client/messages/Info.java
```diff
@@ -26,17 +26,27 @@
   @JsonProperty("Containers") private int containers;
   @JsonProperty("Images") private int images;
   @JsonProperty("Driver") private String storageDriver;
+  @JsonProperty("DriverStatus") private List<String> driverStatus;
   @JsonProperty("ExecutionDriver") private String executionDriver;
   @JsonProperty("KernelVersion") private String kernelVersion;
+  @JsonProperty("NCPU") private int cpu;
```
rohansingh: Let's call this `cpus` to match what we do for `NGoroutines` (`goroutines`).

src/main/java/com/spotify/docker/client/messages/Info.java
```diff
@@ -26,17 +26,27 @@
...
+  @JsonProperty("MemTotal") private long memtotal;
```
rohansingh: Should be `memTotal`.

rohansingh:
Couple nitpicks, other than that looks great. Thanks!

xcoulon:
@rohansingh thanks for the quick review ! I updated the PR after your comments.

xcoulon:
Thanks, @rohansingh !

PR #144, opened 2015-03-20
Get rid of "sockets" in Info

rohansingh:
Maybe `docker info` returned a list of sockets in the distant past, but it
doesn't seem to be a thing anymore. Can't find it documented anywhere either.


davidxia:
:+1:

PR #149, opened 2015-03-26
Add ability to force re-tagging of images

dflemstr:
This change should be fairly self-descriptive.


rohansingh:
I think it would be simpler if instead of the whole `TagParameter` thing, we just have an overload that accepts `boolean force`. There really aren't any other parameters anyway.

Also could you please add a test for this to the test suite? Thanks.

dflemstr:
@rohansingh, I rewrote the commits with the changes you suggested.

PR #159, opened 2015-04-22
Dockerignore support

dflemstr:
Please review the commits individually.

This adds support for `.dockerignore` files when running the `build` command.

As indicated in the `CompressedDirectoryMatchFilepathTest.java` file, we are not 100% compatible with the Go path matcher, but the exception cases are so rare that it shouldn't matter, and we only fail in negative cases (i.e. when the Go implementation would have thrown an exception, we don't).

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -21,6 +21,7 @@ package com.spotify.docker.client;
 
 import com.google.common.io.CharStreams;
 import com.google.common.net.HostAndPort;
+
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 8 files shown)


dflemstr:
ITs fail due to a refactoring accident, will fix.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -806,11 +806,11 @@ public String build(final Path directory, final String name, final ProgressHandl
      resource = resource.queryParam("t", name);
     }
 
-    final File compressedDirectory = CompressedDirectory.create(directory);
-
-    try (ProgressStream build = request(POST, ProgressStream.class, resource,
+    try (CompressedDirectory compressedDirectory = CompressedDirectory.create(directory);
+         InputStream fileStream = Files.newInputStream(compressedDirectory.file());
```
rohansingh: `final InputStream fileStream`

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -806,11 +806,11 @@ public String build(final Path directory, final String name, final ProgressHandl
...
+         ProgressStream build = request(POST, ProgressStream.class, resource,
```
rohansingh: `final ProgressStream build`

rohansingh:
Reviewed the rest of the commits. Looks great. Some minor nits inline, otherwise :+1: 

Also, want to squash your last two commits (spacing + tar) into the previous ones?

dflemstr:
@rohansingh rebased.

PR #161, opened 2015-04-23
Fix tests

mbruggmann:


rohansingh:
The reason for the parallelism is that in one instance we test via TCP and the other we test Unix sockets.

mbruggmann:
Reverted the parallelism change now, it wasn't the cause for image layer issue.

For future reference, here is the error:
```
com.spotify.docker.client.ImagePullFailedException: Image pull failed: dxia2/scratch-private:latest: ProgressMessage{id=null, status=null, stream=null, error=Error pulling image (latest) from dxia2/scratch-private, Driver btrfs failed to create image rootfs 511136ea3c5a64f264b78b5433614aec563103b4d4702f3ba7d4d2698e22c158: Failed to create btrfs subvolume: file exists, progress=null, progressDetail=null}
```

It's likely related to removing an image and later pulling the same image again on the the same docker daemon but I couldn't track it down further.

PR #163, opened 2015-04-24
Fix testRemoveImage & don't run it on CircleCI

rohansingh:
Before we were pointlessly trying to verify Docker internals.

Now just check that when we remove an image, it's untagged and no longer exists.


mbruggmann:
:+1:

PR #169, opened 2015-05-06
Add Container.portsAsString()

davidxia:
This method shows port information similar to the output of docker ps.


rohansingh:
:+1:

PR #170, opened 2015-05-06
Adding missing attributes to ContainerState

davidxia:
These attributes are available as of Docker 1.5, remote API 1.17.
We include them here to prepare for future upgrades.


davidxia:
@rculbertson

davidxia:
@rohansingh

rohansingh:
:+1:

Issue #157, opened 2015-04-17
X-Registry-Auth deprecated


kevtaylor:
As far as I can make out, the api version being called from the spotify client is V1.12

1).
In the docker server.go software the X-Registry-Auth appears to only be used prior to V1.9
as: 

// This block can be removed when API versions prior to 1.9 are deprecated.
// Both headers will be parsed and sent along to the daemon, but if a non-empty
// ConfigFile is present, any value provided as an AuthConfig directly will
// be overridden. See BuildFile::CmdFrom for details.
if version.LessThan("1.9") && authEncoded != "" {

it has been superseded by X-Registry-Config

2).
Also, the structure of X-Registry-Auth appears to be this:

Username      string `json:"username,omitempty"`
Password      string `json:"password,omitempty"`
Auth          string `json:"auth"`
Email         string `json:"email"`
ServerAddress string `json:"serveraddress,omitempty"`

There doesn't appear to be a way of adding the "auth" element to the spotify client

3).
The /build element of the rest api requires authentication on a private repo 
I tried adding the header .header("X-Registry-Auth", authHeader(authConfig)) to the build method
but I cannot get this build to work without falling over with a 403 error - the credentials don't seem to be passed correctly

Can anyone shed any light on these?

davidxia:
@kevtaylor 

1. The version in the URL this docker-client makes requests to is "v1.12" but the docker daemon just ignores this. We've been testing and building docker-client against docker 1.3 with api version 1.15 for several months now. Soon we'll make it explicit in the docs and code what version we've been testing on.

2. I think the value of X-Registry-Auth should be 

```
{"Username": "foo", "Password": "pass", "Email": "foo@bar.com"[, "ServerAddress": "example.com"]}
```

3 Where are you trying to add `.header("X-Registry-Auth", authHeader(authConfig))`? Isn't this already in  `DefaultDockerClient.push()` and `pull()`?

kevtaylor:
The problem we are facing is actually build Docker images through the API (indirectly through the maven plugin). So build() is the method being called here. This explicitly requires X-Registry-Config in versions after 1.8 (I have actually added X-Registry-Auth) to this method and changed the API version to 1.8 which then made the code work. 
When the build api is called it will attempt to get https://registry/v1/someapp/images and over SSL this requires authentication which is missing from the spotify docker client code

kevtaylor:
@rohansingh Thanks for merging. This issue is resolved by the PR

PR #171, opened 2015-05-11
Add authentication X-Registry-Config to build api for SSL support on private repository

kevtaylor:
Introduce a wrapper class to pass auth values and return formatted X-Registry-Config string.
The user/password/email and repository are base64 encoded and passed on the request header.

Notes - in order to get a local build, a couple of issues were found: 
a). Found that the ant version 1.8.2 is transitively pulled which causes findbugs to fail with a class not found exception - explicitly added 1.9.4 in the pom to correct this
b). Although exec has been added to the API in 1.15, exec inspect didn't get added until 1.16 so this causes failures in docker 1.3.2 (API V1.15)

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -27,6 +27,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.spotify.docker.client.messages.AuthConfig;
+import com.spotify.docker.client.messages.AuthRegistryConfig;
 import com.spotify.docker.client.messages.Container;
 import com.spotify.docker.client.messages.ContainerConfig;
 import com.spotify.docker.client.messages.ContainerCreation;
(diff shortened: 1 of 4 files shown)


kevtaylor:
Fixes issue: https://github.com/spotify/docker-client/issues/157

src/main/java/com/spotify/docker/client/messages/AuthRegistryConfig.java
```diff
@@ -0,0 +1,137 @@
...
+   * @return String
+   * {"configs":{"myrepository":{
+   *   "username":"myusername",
+   *   "password":"mypassword",
+   *   "auth":"",
+   *   "email":"myemail",
+   *   "serveraddress":"myserverAddress"}}}
+  */
+  public String toConfig() {
+    StringBuilder sb = new StringBuilder();
```
rohansingh: We should use the appropriate jackson stuff rather than building JSON strings by concatenation. The latter could end poorly depending on the values supplied by the user.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -807,11 +808,33 @@ public String build(final Path directory, final String name, final ProgressHandl
      resource = resource.queryParam("t", name);
     }
 
+    log.debug("Auth Config {}", authConfig);
+
+    // Convert auth to X-Registry-Config format
+    AuthRegistryConfig registryConfig = null;
+    if (authConfig == null) {
+      registryConfig = new AuthRegistryConfig("", "", "", "", "", "");
```
rohansingh: Would be better to have an `AuthRegistryConfig.EMPTY` static variable rather than constructing a new empty instance each time.

src/main/java/com/spotify/docker/client/messages/AuthRegistryConfig.java
```diff
@@ -0,0 +1,137 @@
...
+   * @param username    User name
+   * @param password    Password for authentication
+   * @param auth      Not used but must be supplied
+   * @param email     EMAIL address of authenticated user
+   * @param serverAddress The address of the repository
+  */
+  public AuthRegistryConfig(String repository, 
+                            String username, 
+                            String password, 
+                            String auth,
```
rohansingh: Maybe have an overload that doesn't require this since it isn't used.

rohansingh:
Thanks for the PR! I made some comments inline. Could you please also squash your commits into either a single commit, or rebase into a single commit per semantic change.

For example, you could just have one commit that has the same title as this PR. Alternatively, you could have 3 commits, one for adding the auth stuff, one for ant stuff, and one for fixing the exec test.

Either of these approaches help keep the project's commit history clean & useful.

src/main/java/com/spotify/docker/client/messages/AuthRegistryConfig.java
kevtaylor: @rohansingh Thanks for your comments and I'll have a look at applying them. I don't quite follow this comment about an overload though. Please could you elaborate?

kevtaylor: @rohansingh Scrub that last comment. I get it - you are referring to the unused auth property. Fine no problems

kevtaylor:
@rohansingh All changes applied as discussed.

kevtaylor:
Would somebody be able to update us on whether this pull request can be approved and merged?
We are currently building a local patched version and there are a couple of other interested parties from the plugin forum who would benefit from this fix, so it would be nice to get it released.
Thanks for any input. Kevin

davidxia:
@rohansingh Other than your comment above, is this good to go? I can review it as well if you'd like.

kevtaylor:
@davidxia @rohansingh Hi. Do you have any update on our pull request?

rohansingh:
Apologies guys, dropped the ball on this. @kevtaylor If you want to rebase on latest master, I'm happy to merge.

kevtaylor:
@rohansingh @davidxia  Rebase completed - tests all good.

PR #173, opened 2015-05-18
Use Docker API v1.18 corresponding to Docker 1.6.0

davidxia:


rohansingh:
:+1:

kevtaylor:
Have you tried this on older docker versions? If I try and access the docker api directly on 1.3.2 using /v18/ it complains about the api version only supporting version 15 so I would assume (but I haven't tested it), that this would also fail.

davidxia:
@kevtaylor Yes, this wouldn't work for older versions. Personally, I'd like to have docker-client not stray too far away from the latest docker release and document clearly what API version each docker-client release is explicitly using. Users should try to upgrade docker so they can use newer versions of docker-client.

Older releases of docker-client would just stay frozen. We could come up with a versioning scheme that allows for bug fixes (or even feature development through PRs?) for older versions. I'm not sure how external people are using docker-client right now and how easy it is for them to upgrade docker.

davidxia:
@rohansingh One downside of this is that circleCI doesn't have docker 1.6.x :(

kevtaylor:
@davidxia Although we want to move up to docker 1.6 we are currently constrained by UNIX kernel versions (we are using centos 6.5) which cause failures in the docker versions. it is quite a large upgrade project to get it to the new version. So we are using the docker-client indirectly via the maven plugin. What I thought might be possible would be to examine the API version directly but set a max version so that you could then dynamically work out the API version, but not allow it to stray over the latest tested codebase.

davidxia:
@kevtaylor Ah I see. Knowing the docker runtimes users have is very helpful information for us. Your idea sounds reasonable.

davidxia:
@kevtaylor We've decided to remove the version altogether which will default to using the latest docker API version in the installed docker runtime. https://github.com/spotify/docker-client/pull/176

kevtaylor:
:+1:

PR #178, opened 2015-05-21
Fix javadoc errors

davidxia:


davidxia:
Fixing because they seem to break our build now but not sure if it's the cause. https://jenkins-helios.spotify.net/job/docker-client/58/console

rculbertson:
:+1:

Issue #177, opened 2015-05-21
how to get the exited containers for java


qq1514730142:
like this ?    docker.listContainers(ListContainersParam.create("filters", "status[exited]"));
but is error,waiting for help,thinks

zanella:
for (Container c : docker.listContainers( DockerClient.ListContainersParam.create("all", "1") )) {
    System.out.println("(DBG) container: " + c.status());
}

zanella:
docker.listContainers(
                    DockerClient.ListContainersParam.allContainers(),
                    DockerClient.ListContainersParam.create("filters", URLEncoder.encode("{\"status\":[\"exited\"]}", "UTF-8"))
            ).stream()
                .forEach(c -> System.out.println("(DBG) status: " + c.status()));

qq1514730142:
Think you very much!

PR #179, opened 2015-05-21
Added listExitedContainers() method

zanella:


zanella:
#177

davidxia:
@zanella Thanks for this. Maybe instead of its own special method, we can add another static method to `class ListContainersParam` that will generate the key and value?

zanella:
@davidxia Absolutely, the "problem" is that the filter depends on "all=1", so I would have to return an Array/List<ListContainersParam>... am I wrong in my assumption ?

davidxia:
@zanella Hm it seems to work without the all. See https://github.com/spotify/docker-client/pull/181

zanella:
@davidxia Ah!, indeed, then #181

PR #184, opened 2015-05-27
add VirtualSize

crazywen:
docker now response have the virtualSize,it should be add.


davidxia:
@crazywen Thanks for this. I fixed a checkstyle error and updated a test here https://github.com/spotify/docker-client/pull/186

PR #185, opened 2015-05-27
Reinstate a forgotten test

davidxia:


davidxia:
@rculbertson @rohansingh Did we just forget to run this test or was there a reason it's ignored?

PR #187, opened 2015-05-29
Ignore failure to kill container in test tear down

davidxia:
Docker 1.6 sometimes fails to kill a container because it disappears.
https://github.com/docker/docker/issues/12738

Ignore these failures.


davidxia:
@rculbertson

gimaker:
:+1: are the left-behind ghost processes going to cause other funky issues?

gimaker:
(oops, fat-fingered the wrong button)

davidxia:
Not sure, but this would only screw up our test host so it wouldn't be that bad. The tests on my boot2docker almost never pass for my anymore :(

rohansingh:
:+1:

PR #190, opened 2015-06-04
Initial Support for Remote API v1.18

diptanu:
Added support for the following -

a. Labels
b. Mac Address
b. CPU, CPU Set, Memory, Swap, Parent cgroup

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
@@ -19,12 +19,14 @@ package com.spotify.docker.client.messages;
 
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
@@ -56,6 +58,8 @@ public class ContainerConfig {
   @JsonProperty("Entrypoint") private ImmutableList<String> entrypoint;
   @JsonProperty("NetworkDisabled") private Boolean networkDisabled;
   @JsonProperty("OnBuild") private ImmutableList<String> onBuild;
+  @JsonProperty("Labels") private ImmutableMap<String, String> labels;
+
 
   private ContainerConfig() {
   }
@@ -84,6 +88,7 @@ public class ContainerConfig {
     this.entrypoint = builder.entrypoint;
     this.networkDisabled = builder.networkDisabled;
     this.onBuild = builder.onBuild;
+    this.labels = builder.labels;
   }
 
   public String hostname() {
@@ -178,6 +183,8 @@ public class ContainerConfig {
     return onBuild;
   }
 
+  public Map<String, String> labels() { return labels; }
+
   @Override
   public boolean equals(final Object o) {
     if (this == o) {
@@ -264,6 +271,10 @@ public class ContainerConfig {
       return false;
     }
 
+    if (labels != null ? !labels.equals(config.labels) : config.labels != null) {
+      return false;
+    }
+
     return true;
   }
 
@@ -292,6 +303,7 @@ public class ContainerConfig {
     result = 31 * result + (entrypoint != null ? entrypoint.hashCode() : 0);
     result = 31 * result + (networkDisabled != null ? networkDisabled.hashCode() : 0);
     result = 31 * result + (onBuild != null ? onBuild.hashCode() : 0);
+    result = 31 * result + (labels != null ? labels.hashCode() : 0);
     return result;
   }
 
@@ -321,6 +333,7 @@ public class ContainerConfig {
         .add("entrypoint", entrypoint)
         .add("networkDisabled", networkDisabled)
         .add("onBuild", onBuild)
+        .add("labels", labels)
         .toString();
   }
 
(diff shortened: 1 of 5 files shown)


diptanu:
@davidxia @rohansingh  The test failed because we couldn't find Docker 1.6.2 I guess. Not sure how the test environment is setup. Any pointers?

davidxia:
We use circleci which last I checked only supports up to 1.5. Add an
'assumes'?

On Thursday, June 4, 2015, Diptanu Choudhury <notifications@github.com>
wrote:

> @davidxia <https://github.com/davidxia> @rohansingh
> <https://github.com/rohansingh> The test failed because we couldn't find
> Docker 1.6.2 I guess. Not sure how the test environment is setup. Any
> pointers?
>
> —
> <https://github.com/spotify/docker-client/pull/190#issuecomment-109029845>
> .
>


-- 
David Xia
Software Engineer
dxia@spotify.com

diptanu:
@davidxia @rohansingh  Not sure why rest of the tests are failing. They work locally. Can you guys take a look please? I have tested this out on our servers, works fine now. cpu.shares is getting set appropriately.  What are the next steps here?

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
```diff
@@ -629,6 +668,33 @@ public Builder onBuild(final String... onBuild) {
       return onBuild;
     }
 
+    public Builder labels(Map<String, String> labels) {
```
davidxia: Add a `final` here for consistency?

src/main/java/com/spotify/docker/client/messages/HostConfig.java
```diff
@@ -46,6 +46,11 @@
   @JsonProperty("VolumesFrom") private ImmutableList<String> volumesFrom;
   @JsonProperty("NetworkMode") private String networkMode;
   @JsonProperty("SecurityOpt") private ImmutableList<String> securityOpt;
+  @JsonProperty("Memory") private Long memory;
+  @JsonProperty("MemorySwap") private Long memorySwap;
+  @JsonProperty("CpuShares") private Long cpuShares;
+  @JsonProperty("CpusetCpus") private String cpuset;
```
davidxia: Any reason to use `cpuset` instead of `cpisetCpus`? The latter would be more consistent with the docker API.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -935,25 +938,31 @@ public void testInspectContainerWithSecurityOpts() throws Exception {
 
   @Test
   public void testContainerWithHostConfig() throws Exception {
+    assumeTrue("Docker API should be at least v1.18 to support Container Creation with HostConfig, got "
+            + sut.version().apiVersion(), versionCompare(sut.version().apiVersion(), "1.18") >= 0);
```
davidxia: nitpick: line up the `+` with `"` above.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -935,25 +938,31 @@ public void testInspectContainerWithSecurityOpts() throws Exception {
 
   @Test
   public void testContainerWithHostConfig() throws Exception {
+    assumeTrue("Docker API should be at least v1.18 to support Container Creation with HostConfig, got "
```
davidxia: style nitpick: don't exceed 100 chars on a line

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -935,25 +938,31 @@ public void testInspectContainerWithSecurityOpts() throws Exception {
 
   @Test
   public void testContainerWithHostConfig() throws Exception {
```
davidxia: Maybe we should add an extra test specifically for container creation where the HostConfig is in the ContainerConfig so the assumeTrue doesn't ignore this test for docker API versions less than 1.18.

davidxia: For reference, Docker API 1.15 is the last version where `/containers/(id)/start` took a HostConfig JSON in the POST data. https://docs.docker.com/reference/api/docker_remote_api_v1.15/#start-a-container

diptanu: @davidxia Yes, I noticed that. So are you guys willing to drop support for 1.15 entirely? I am happy to update the PR to make the API conform to only API version 1.18

davidxia: I'd prefer to keep this PR small. So maybe include an assumeTrue for this test for docker API <= 1.15 and rename it to something like `testContainerWithHostConfigOldApi`. Then add another `testContainerWithHostConfig` with your changes above?

@rohansingh thoughts?

diptanu:
@davidxia Addressed your review comments. Waiting to hear if you guys want to drop support for versions less than 1.15 completely since CPU shares, CPU Set etc aren't in ContainerConfig anymore after v1.15 and the container start api doesn't take HostConfig anymore. I can update the PR to reflect the new API

rohansingh:
@diptanu Just discussed it and we're OK with dropping support for older versions. We need to figure out how to annotate docker-client version <-> Docker version compatibility, but that's a separate issue.

diptanu:
@rohansingh Thanks, let me update this PR now.

diptanu:
@rohansingh @davidxia I removed the start method in DockerClient which took HostConfig. Please let me know if you guys have more comments?

I would like to get this PR merged and then work on the remaining items which would add support for more properties in the HostConfig like privileges, log driver configuration, RestartPolicy, ReadonlyRootfs, etc.

diptanu:
CircleCI shows Build Succeded and all tests are passing not sure why the PR is not updated with the success callback from Circle CI.

rohansingh:
Green light from CircleCI now. @davidxia I'll leave pushing the button up to you since you reviewed.

davidxia:
@diptanu Amazing! Thanks for this clean commit and the fast replies.

diptanu:
@davidxia @rohansingh Thanks! More commits coming your way to add support for rest of the properties in HostConfig

PR #194, opened 2015-06-09
CircleCI: Cache public Docker images

rohansingh:
Instead of always hitting the public Docker registry, run a mirror locally.

This should hopefully reduce the number of build failures we get due to
image pulls failing against the public Docker registry.


davidxia:
aw why did you close?

rohansingh:
Because it didn't work.

PR #196, opened 2015-06-12
CircleCI: Cache public Docker images

rohansingh:
Instead of always hitting the public Docker registry, run a mirror locally
that uses S3 in us-east-1 as its backing store.

This should hopefully reduce the number of build failures we get due to
image pulls failing against the public Docker registry. us-east-1 is the
same region that CircleCI workers are in.

### DIFF
--- a/circle.sh
+++ b/circle.sh
@@ -11,7 +11,7 @@ case "$1" in
     fi
 
     # have docker bind to both localhost and unix socket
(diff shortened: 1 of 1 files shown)


rohansingh:
@rculbertson

rculbertson:
:+1:

PR #197, opened 2015-06-15
Add support for one-time container stats

unguiculus:
Retrieves one-time stats for a specific container using
`/containers/<container>/stats?stream=0`. Docker 1.7.2-rc3 or higher
is required for this to work.


src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1002,6 +1003,24 @@ public ExecState execInspect(final String execId) throws DockerException, Interr
     }
   }
 
+  @Override
+  public ContainerStats stats(final String containerId) 
+           throws DockerException, InterruptedException {
+    WebTarget resource = resource().path("containers").path(containerId).path("stats")
```
rohansingh: Should be `final WebTarget resource`

rohansingh:
Looks great. One minor comment inline.

PR #202, opened 2015-06-27
added null-empty check on config builders

fbertola:
When using the configuration builders while loading properties from e.g. a map or a file, frequently a null or empty value is passed. While for _Strings_ or _Longs_ this is not a problem, _Maps_ and _Collections_ requires an external check thus breaking the fluency and overall being very annoying.

This PR implements those simple checks inside the builders.


davidxia:
@fbertola I've rebased and squashed here https://github.com/spotify/docker-client/pull/208

PR #204, opened 2015-07-05
Remove unused dependency jackson-dataformat-yaml

### LINKED ISSUE #201: jackson-dataformat-yaml does not appear to be used
[jackson-dataformat-yaml](https://github.com/spotify/docker-client/blob/v3.0.0/pom.xml#L100) is listed as a compile-time dependency but it doesn't appear to be used anywhere in the code. Running the build and tests with the dependency removed seems to work so I wonder if it's safe to remove, or is it satisfying something I've missed.

davidxia:
Fixes #201

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,11 +96,6 @@
       <version>2.2.3</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


gimaker:
:+1:

PR #209, opened 2015-07-16
Support Remote API v1.19.

rgrunber:
- Debug field in Info changed from an integer back to a boolean.

I needed this change to get docker-client working with the [v1.19 Remote API](https://docs.docker.com/reference/api/docker_remote_api_v1.19/#display-system-wide-information). Specifically, the Debug field was changed back to a boolean.

I'm not sure when the plan is for docker-client in terms of when to adopt the new API, but at least they're known.


rgrunber:
I would have expected this to break against the daemon (which I assumed was running 1.18).

PR #211, opened 2015-07-17
Add support for specifying a dockerfile when building.

nickethier:
Starting with api version [1.17](https://docs.docker.com/reference/api/docker_remote_api_v1.17/#build-image-from-a-dockerfile) docker has allowed you to specify the path to the dockerfile to build within the given build context.

This pull simply adds support for setting that parameter.


xcoulon:
@davidxia any chance this pull request could be merged, too ? 
Once it is released, we could leverage the dockerfile support in the Eclipse Docker tooling ;-)

davidxia:
@nickethier @xcoulon Thanks! I added a test here https://github.com/spotify/docker-client/pull/227

PR #218, opened 2015-08-04
Actually append apiversion to uripath

brujoand:
```
     * @param path the path, may contain URI template parameters.
     * @return a new target instance.
     * @throws NullPointerException if path is {@code null}.
     */
    public WebTarget path(String path);
```
WebTarget.path(String) returns a new instance, so the new instance should be returned.


davidxia:
doh, thanks!

PR #219, opened 2015-08-04
Adding support for loading authConfig from dockers config

brujoand:
My Java fu is a bit outdated, but this should get the job done.
Thanks

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/AuthConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/AuthConfig.java
@@ -17,11 +17,19 @@
 
 package com.spotify.docker.client.messages;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Objects;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.glassfish.jersey.internal.util.Base64;
+import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


brujoand:
Any feedback on this?

davidxia:
@brujoand Sorry for the delayed response. I just merged https://github.com/spotify/docker-client/pull/261 but haven't gotten a chance to see go over how yours differs from that one yet.

mattnworb:
Overall LGTM, but what happens if there are multiple servers in the `auths` element of `~/.docker/config.json`? Seems like only the first one is used. It might be best to attempt to match it up with a user-supplied URL.

src/main/java/com/spotify/docker/client/messages/AuthConfig.java
```diff
@@ -168,6 +178,69 @@ public String serverAddress() {
...
+    }
+
+    private JsonNode extractDockerJson(String configFile) throws IOException {
+      ObjectMapper mapper = new ObjectMapper();
+      JsonNode config = mapper.readTree(new File(configFile));
+
+      if (config.has("auths")) {
+        return config.get("auths");
+      }
+      return config;
```
mattnworb: I don't think there would be any value in returning the JsonNode if it doesn't have an `auths` field, as the calling code takes the first field in the return value as the servername

mattnworb:
Oh and thank you for adding tests with your change :wink:

brujoand:
Thanks for the feedback, I've cleaned up some of this on the branch we use at work. I'll push the changes when I get in tomorrow! :)

brujoand:
I moved this to #266
So I'll just close this one

PR #220, opened 2015-08-04
Added "--pull" and "--rm=true" build parameters

xcoulon:
Also:
- renamed the BuildParameter fields for clarity
- added a method to retrieve the actual value for an existing parameter (for external tools convinience)


davidxia:
Thanks!

PR #221, opened 2015-08-04
Add support for missing parameters for showing logs.

dennylin93:
Currently the parameters (1) tail *number* and (2) since *timestamp* aren't supported.

Query parameters for showing logs: https://docs.docker.com/reference/api/docker_remote_api_v1.19/#get-container-logs

### DIFF
--- a/src/main/java/com/spotify/docker/client/DockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DockerClient.java
@@ -511,7 +511,7 @@ public interface DockerClient extends Closeable {
    * @throws DockerException if a server error occurred (500)
    * @throws InterruptedException If the thread is interrupted
    */
-  LogStream logs(String containerId, LogsParameter... params)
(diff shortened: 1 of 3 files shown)


src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -604,13 +604,142 @@ public String getName() {
   void close();
 
   /**
-   * Parameters for {@link #logs(String, LogsParameter...)}
+   * Parameters for {@link #logs(String, LogsParam...)}
    */
-  public static enum LogsParameter {
```
davidxia: Ah I see you changed the class name here to LogsParam because can no longer be an enum. And enums in this interface are Parameters while classes are Param. :+1:

davidxia:
I've added tests and a note about how since is only available in Docker API >= 1.19 in #262 which will replace this PR.

Issue #146, opened 2015-03-24
ApacheUnixSocket throws Unimplemented instead of calling appropriate JNR-unixsocket methods


cholcombe973:
I'm running arch linux and I'm unable to get the example working.  I'm using docker version: 
Docker version 1.5.0, build a8a31ef

```
URI uri = URI.create(DefaultDockerClient.DEFAULT_UNIX_ENDPOINT);
DockerClient dockerClient = DefaultDockerClient.builder().uri(uri).build();

LOG.info("Ping: " + dockerClient.ping());
```

It seems to ignore this URI that I give it and tries to connect to localhost:80

```
[jersey-client-async-executor-0] DEBUG org.apache.http.impl.execchain.MainClientExec - Opening connection {}->unix://localhost:80
[jersey-client-async-executor-0] DEBUG org.apache.http.impl.conn.HttpClientConnectionOperator - Connecting to localhost/127.0.0.1:80
[jersey-client-async-executor-0] DEBUG org.apache.http.impl.conn.DefaultManagedHttpClientConnection - http-outgoing-0: Shutdown connection
[jersey-client-async-executor-0] DEBUG org.apache.http.impl.execchain.MainClientExec - Connection discarded
[jersey-client-async-executor-0] DEBUG org.apache.http.impl.conn.DefaultManagedHttpClientConnection - http-outgoing-0: Close connection
[jersey-client-async-executor-0] DEBUG org.apache.http.impl.conn.PoolingHttpClientConnectionManager - Connection released: [id: 0][route: {}->unix://localhost:80][total kept alive: 0; route allocated: 0 of 100; total allocated: 0 of 100]
```

Then it fails with this:

    com.spotify.docker.client.DockerException: java.util.concurrent.ExecutionException: javax.ws.rs.ProcessingException: java.lang.UnsupportedOperationException: Unimplemented

Any insights?  I went deep with the debugger but couldn't figure out why this was happening.

rohansingh:
`localhost:80` here is actually just the dummy host we use when using the Unix sockets stuff. This is because most HTTP-aware libraries (Apache, Jersey, etc.) expect there to be a host, even though that concept doesn't make sense for Unix sockets.

So that `Opening connection {}->unix://localhost:80` always gets logged, even when things aren't working. I guess the question is this `UnsupportedOperationException` — what is actually being thrown by? Do you have a stack trace?

cholcombe973:
Yeah here's the stack trace:
[...]

PR #222, opened 2015-08-05
Update Maven build configuration, dependencies and adapt code to them.

akurtakov:
* Update maven plugin versions.
* Update parent pom.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,8 +18,12 @@
   <parent>
     <groupId>org.sonatype.oss</groupId>
     <artifactId>oss-parent</artifactId>
-    <version>7</version>
+    <version>9</version>
   </parent>
+
+  <prerequisites>
+    <maven>3.0.4</maven>
+  </prerequisites>
 
   <licenses>
     <license>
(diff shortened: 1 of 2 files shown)


davidxia:
@akurtakov Thanks for this!

@rohansingh How does it look to you? Specifically this part https://github.com/spotify/docker-client/pull/222/files#diff-a479355e73762bc07b466a285d105088R273. Did we not need to do this before?

rgrunber:
I was just looking at @akurtakov's patch and noticed the SO_LINGER changes so I thought I'd clarify. As of Apache HttpCore 4.4.0, setSoLinger is now being used in BHttpConnectionBase (See apache/httpcore@c166b587fb1861b46634b2697450fc6bda979bd8 ). Going from 4.3.5 to 4.5.0 would probably cause users to hit this as it happened with Issue #146.

davidxia:
@rgrunber Ah thanks for explaining. I'll wait for another :+1: from either @rohansingh or @gimaker

maxandersen:
any update on this ?

PR #223, opened 2015-08-05
attachContainer, logs and build should use the 'noTimeoutClient'

rgrunber:
Like waitContainer, and stopContainer, attachContainer may take an
arbitrary amount of time to finish, and if this exceeds the socket read
timeout (default 30s), data will have been overlooked.

I've pushed a commit at rgrunber@58e1cd9 that simply reproduces the issue by modifying the testAttachLog slightly. The issue seems to occur in both TCP/UNIX cases, but only the TCP connection throws the SocketTimeoutException.

I set the socket read timeout to 5s for convenience. It's 30s by default (and 120s in the tests) but it's just as possible to have a program with a larger gap in between data being read. The testAttachLog simply pulls/create a Fedora 22 image that prints 'Seen output after X seconds.', after waiting X seconds, where X is incremement from 1 to 7, by 1. (beyond the read timeout).

I reproduce as follows (assuming the daemon is listening on the default TCP connection) :

$ DOCKER_HOST=tcp://localhost:2375 mvn clean test -Dtest=DefaultDockerClientTest#testAttachLog

...
...

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running com.spotify.docker.client.DefaultDockerClientTest
- testAttachLog
Seen output after 1 seconds.

Seen output after 2 seconds.

Seen output after 3 seconds.

Seen output after 4 seconds.

Seen output after 5 seconds.

java.net.SocketTimeoutException: Read timed out

Reading has finished, waiting for program to end.

Just a few things to  note. The program still runs within the container, but no additional output is collected due to the read timeout. In the case of a UNIX socket the SocketTimeoutException is never thrown likely due to the different implementation. It seems like the stream indicates there's no more content and waitContainer just waits for the program to finish.

Changing the attachContainer call to use 'noTimeoutResource()' resolves the issue.

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running com.spotify.docker.client.DefaultDockerClientTest
- testAttachLog
Seen output after 1 seconds.

Seen output after 2 seconds.

Seen output after 3 seconds.

Seen output after 4 seconds.

Seen output after 5 seconds.

Seen output after 6 seconds.

Seen output after 7 seconds.

Reading has finished, waiting for program to end.

I could write a proper test case for this as part of the commit, but I'd need to look into using some well-known lightweight image that has a shell environment and some basic commands like echo, and sleep . Also, execStart, and logs may also need similar changes but I haven't fully tried those cases out.

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -25,6 +25,7 @@ import com.google.common.util.concurrent.SettableFuture;
 
 import com.fasterxml.jackson.databind.util.StdDateFormat;
 import com.spotify.docker.client.DockerClient.AttachParameter;
+import com.spotify.docker.client.DockerClient.LogsParameter;
 import com.spotify.docker.client.messages.AuthConfig;
 import com.spotify.docker.client.messages.Container;
 import com.spotify.docker.client.messages.ContainerConfig;
@@ -84,6 +85,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
+import static com.google.common.base.Charsets.UTF_8;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static com.spotify.docker.client.DefaultDockerClient.NO_TIMEOUT;
@@ -165,7 +167,14 @@ public class DefaultDockerClientTest {
     authConfig = AuthConfig.builder().email(AUTH_EMAIL).username(AUTH_USERNAME)
         .password(AUTH_PASSWORD).build();
     final DefaultDockerClient.Builder builder = DefaultDockerClient.fromEnv();
-    builder.readTimeoutMillis(120000);
+    // Make it easier to test no read timeout occurs by using a smaller value
+    // Such test methods should end in 'NoTimeout'
+    if (testName.getMethodName().endsWith("NoTimeout")) {
+      builder.readTimeoutMillis(5000);
+    } else {
+      builder.readTimeoutMillis(120000);
+    }
+
     dockerEndpoint = builder.uri();
 
     sut = builder.build();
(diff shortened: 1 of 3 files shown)


rgrunber:
Actually, now that I've looked deeper, it seems entirely possible to use BUSYBOX_LATEST for the test with just some modifications.

rgrunber:
So I've settled on using a simple convention for the test method name. If it ends in 'NoTimeout', then we assume the test should set a small read timeout so that functionality may be tested. I'm also open to getting rid of the test and just having the patch if it seems like too much for just this one case.

maxandersen:
any update on this ? This prevents actually using the client to build even the simplest of docker containers that might need a few minutes to download dependencies.

xcoulon:
To illustrate @maxandersen's comment above: if the Dockerfile needs to run a long command such as 'yum update' and no output sent back to the client for more than 30s, the client looses the connection to the Docker daemon. That does not mean that the image build is not continuing, but from a client perspective, all other steps in the Dockerfile happen in the dark.
On the other hand, @rgrunber's patch would ensure that the client remains connected even if no data is sent to the client.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1543,4 +1592,29 @@ private PoolStats getNoTimeoutClientConnectionPoolStats(final DefaultDockerClien
     return ((PoolingHttpClientConnectionManager) client.getNoTimeoutClient().getConfiguration()
         .getProperty(ApacheClientProperties.CONNECTION_MANAGER)).getTotalStats();
   }
+
+  private String createNoTimeoutContainer() throws Exception {
```
davidxia: Maybe we can name this `createSleepingContainer` to indicate what it does instead of what test it's for?

davidxia:
Thanks for this @rgrunber. Thanks @xcoulon for poking us. It'd be great to also have a no timeout test for building containers. Other than that it looks good to me. It doesn't seem like anything should be adversely affected by this change. @rohansingh thoughts?

rgrunber:
I've made the requested changes as well as added a test case for no-timeout on build().

In case anyone is wondering why I went with a 10 second sleep in the image build test, when the timeout is set for 5s, I noticed that an additional message is printed on the socket even after the sleep starts running, and 6/7 seconds seemed like it might not be enough to truly trigger a timeout, hence a slightly longer sleep to be sure.

xcoulon:
@davidxia @rohansingh is there anything else that we should do on this pull request ? Otherwise, could you merge it ?

davidxia:
@xcoulon @rgrunber Thanks for making the changes. :+1:

xcoulon:
@davidxia great, thanks for reviewing and merging the patch too ;-)

xcoulon:
@davidxia if the process is not too complex for you, would it be possible to trigger a build with this patch included ?

davidxia:
@xcoulon @rgrunber Done. It should be available soon on maven central as 3.1.10

xcoulon:
@davidxia awesome, thanks a lot !

PR #225, opened 2015-08-09
Sample doesn't compile on 2.7.7

nurkiewicz:
`HostConfig` was moved to `startContainer()`.


davidxia:
Thanks. Sorry for the out of date docs.

PR #235, opened 2015-09-02
Handle empty patterns in .dockerignore

ravwojdyla:
Two issues:
 * handle empty line - pattern that matches everything
 * handle white character patterns

Empty lines should be ignored, and patterns should be trimmed.


ravwojdyla:
[Docker code](https://github.com/docker/docker/blob/9d3ad6d892312b544950dc1c2d7aae224a4b9d06/utils/utils.go#L259) for reference.

PR #236, opened 2015-09-02
Fixed resource path when API version is specified

kwart:
The `javax.ws.rs.client.WebTarget.path(String)` method creates a new `WebTarget` instance. The new instance has to be returned from the `resource()` and `noTimeoutResource()` methods of `DefaultDockerClient` class.


brujoand:
https://github.com/spotify/docker-client/pull/218 Might take a while for this to get merged, hehe..

davidxia:
Don't know why the tests failed here when in #218 they passed. I merged #218 so closing this one. @kwart Thanks though.

PR #237, opened 2015-09-05
Use registryUrl in push command

rigidsh:
In docker-maven-plugin pushing to private registry not work. ( I set serverId and registryUrl, but plugin try push to docker.io)

here example in official documentation about pushing to private registry
https://docs.docker.com/reference/api/docker_remote_api_v1.20/#push-an-image-on-the-registry


davidxia:
@rigidsh Thanks! I've created https://github.com/spotify/docker-client/pull/263 where I amended your commit a tiny bit.

davidxia:
@rigidsh I reverted the commit because to actually push to a private registry, you need to prefix the image tag with the registry, eg myregistry.com/repo/image-name. Then in docker-client's AuthConfig, you set the regsitryUrl, but you don't add the registry's URL to the path for docker daemon. Hope this makes sense.

PR #244, opened 2015-09-21
Add helper method to ProgressMessage to extract image digest

gimaker:


gimaker:
@davidxia @rohansingh

rohansingh:
:+1: though not sure why the build failed

davidxia:
@gimaker I think circleCI docker changed. Master branch fails with same errors.

PR #247, opened 2015-09-23
Fix test

davidxia:
busybox:latest no longer has /usr/bin. Use another dir


gimaker:
:+1:

davidxia:
hm wait, test still seems to fail

PR #248, opened 2015-09-25
Support pulling images by digest & fix broken tests

gimaker:


gimaker:
@rohansingh @davidxia

gimaker:
hrrm this works locally for me... will have to dig into why tests are failing still

davidxia:
CircleCI's docker client, which is 1.6.2-circleci as of this writing, doesn't pull from public registry v2. So we can't use image hashes as tags. https://github.com/spotify/docker-client/pull/249/files

rohansingh:
:+1:

PR #251, opened 2015-10-08
add support for Docker 1.8 image digest statuses

### LINKED ISSUE #250: Image digest output changed in Docker 1.8
#244 added support for capturing the image digest on a push if the output looks like 

```java
return status != null && status.startsWith("Digest: ")
       ? status.substring("Digest: ".length()).trim()
       : null;
```

this works for Docker 1.6 (and maybe 1.7?) because the push output looks like:

```
...
8e5a66f28130: Image already exists
Digest: sha256:dbad7f727f340f4bfa1160689eaf57a60242bf9a48cc2c4c14f383c9b66e8f06
```

However in Docker 1.8 (and maybe 1.7, haven't tested it) the output is changed to look like:

```
...
8e5a66f28130: Image already exists
0.2.1-SNAPSHOT-efd3f71.DIRTY: digest: sha256:8138f3c93c333aa1db6634ae77fce1a50febb9b189ba6fb2121f22810b7841b3 size: 41358
```

so it seems like ProgressMessage.digest() should look to capture both the `"Digest: "` substring as well as a string between `digest: ` and `size:`

mattnworb:
The status message for image pushes that contains the image digest is
different in docker v1.8 compared to v1.6, the format looks like:

```
<image/tag name>: digest: <digest> size: <size in bytes>
```

8dd48c2 added support for parsing the 1.6 style message, this adds the
1.8 format as well.

Fixes #250.

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/ProgressMessage.java
+++ b/src/main/java/com/spotify/docker/client/messages/ProgressMessage.java
@@ -23,9 +23,16 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class ProgressMessage {
 
-  // Prefix that appears before the actual image digest in a status message. E.g.:
+  // Prefix that appears before the actual image digest in a 1.6 status message. E.g.:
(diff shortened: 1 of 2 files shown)


gimaker:
:+1:

PR #256, opened 2015-10-28
Add support for missing parameter for execCreate 

zzzsfk:
execCreate parameter "user" is missing so default docker exec's commands files owns root. Added user query param to exeCreate.

### DIFF
--- a/circle.sh
+++ b/circle.sh
@@ -27,6 +27,9 @@ case "$1" in
 
   test)
     set +x
+    # print version info on the CI machine
+    docker version
+    # run a registry locally
     docker run -d -p 5000:5000 \
       -e STANDALONE=false \
       -e MIRROR_SOURCE=https://registry-1.docker.io \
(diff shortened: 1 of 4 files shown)


davidxia:
@zzzsfk Thanks for this. How come the user field isn't documented here? http://docs.docker.com/engine/reference/api/docker_remote_api_v1.18/#exec-create

Also, you should convert ExecParameter to an inner class named ExecParam and add the user field to that similar to the design of LogsParam.

mattnworb:
If I understand the docker source correctly, [the JSON in the body is decoded](https://github.com/docker/docker/blob/v1.8.0/api/server/server.go#L1450) into an [ExecConfig struct](https://github.com/docker/docker/blob/v1.8.0/runconfig/exec.go#L8) so `User` definitely looks supported.

davidxia:
Ah I see, thanks. Docker docs :(

davidxia:
To be replaced by https://github.com/spotify/docker-client/pull/316

PR #264, opened 2015-11-05
Use MoreObjects.toStringHelper()

davidxia:
instead of the deprecated Objects.toStringHelper()


davidxia:
@mattnworb @gimaker

mattnworb:
:+1:

PR #266, opened 2015-11-06
Support for using docker login credentials from ~/.dockercfg and ~/.docker/config.json

brujoand:
I had a PR in #219 but, it had gone stale so I created this in stead.
I'm not a huge fan of parsing json this way, since server name is used as a key, and there are two almost identical format, this seemed like an acceptable tradeoff.

### DIFF
--- a/src/main/java/com/spotify/docker/client/AuthConfigParser.java
+++ b/src/main/java/com/spotify/docker/client/AuthConfigParser.java
@@ -52,10 +52,10 @@ public class AuthConfigParser {
         File dockerCfg = new File(home, ".dockercfg");
 
         if (dockerConfig.isFile()) {
-            log.debug("Using configfile: " + dockerConfig);
+            log.debug("Using configfile: {}", dockerConfig);
             parseAuthBuilders(dockerConfig);
         } else if (dockerCfg.isFile()) {
-            log.debug("Using configfile: " + dockerCfg);
+            log.debug("Using configfile: {} ", dockerCfg);
             parseAuthBuilders(dockerCfg);
         } else {
             log.error("Could not find a docker config. Please run 'docker login' to create one");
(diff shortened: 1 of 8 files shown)


src/main/java/com/spotify/docker/client/AuthConfigParser.java
```diff
@@ -0,0 +1,146 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * This helper class is used to build a {@Link AuthConf.Builder} from the docker config json
```
davidxia: s/@Link/@link/

src/main/java/com/spotify/docker/client/AuthConfigParser.java
```diff
@@ -0,0 +1,146 @@
...
+        } else if (dockerCfg.isFile()) {
+            log.debug("Using configfile: {} ", dockerCfg);
+            parseAuthBuilders(dockerCfg);
+        } else {
+            log.error("Could not find a docker config. Please run 'docker login' to create one");
+        }
+    }
+
+    /**
+     * @return a {@Link AuthConf.Builder} based on docker config json
```
davidxia: s/@Link/@link/

brujoand:
A little premature PR. Sorry about that. Should be up to scratch now.

brujoand:
Checks are failing because of "429 Too Many Requests". So if they autorun or if someone with super powers can rerun them, we should be good :)

mattnworb:
@brujoand there are some tests that do actual pushes to the docker hub and it seems like the registry will fail them if there are too many with the same credentials (and I guess same image name too) at the same time. Looks like a few builds were submitted within minutes of each other.

Ultimately it'd be best if the tests didn't have that weakness but in the meantime I'll try to rerun the tests a few times when it seems quiet. Since you triggered the build are you unable to hit the Rebuild button yourself in Circle?

brujoand:
@mattnworb Yeah I've come across the same issue myself. The only solution I could think of at the time was setting up a local repository. As for the Circle build, it seems I'm not allowed to rerun it.

davidxia:
@brujoand @mattnworb I tried simplifying this by moving logic into static methods in AuthConfig that return AuthConfig.Builder. https://github.com/spotify/docker-client/pull/268

davidxia:
Let me know what you think of the changes.

davidxia:
Replaced by https://github.com/spotify/docker-client/pull/268

PR #267, opened 2015-11-06
don't use trim on progress streams in AnsiProgressHandler

cap10morgan:
Trim deletes ANSI escape codes in the stream messages, so you end up
with everything after the escape code and then the message (sans color
or whatever else the escape code was trying to do).

Now the code explicitly just deletes any '\n' chars at the end of the
string as the comment claims it should be doing.

### DIFF
--- a/src/main/java/com/spotify/docker/client/AnsiProgressHandler.java
+++ b/src/main/java/com/spotify/docker/client/AnsiProgressHandler.java
@@ -59,7 +59,7 @@ public class AnsiProgressHandler implements ProgressHandler {
     String value = message.stream();
     if (value != null) {
       // trim trailing new lines which are present in streams
-      value = value.trim();
+      value = value.replaceFirst("\n$", "");
     } else {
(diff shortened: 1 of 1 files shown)


cap10morgan:
Those build failures seem transient (too many requests), but it wasn't clear how to re-run it.

davidxia:
Are you able to login to CircleCI to rebuild it?

On Fri, Nov 6, 2015 at 3:36 PM, Wes Morgan <notifications@github.com> wrote:

> Those build failures seem transient (too many requests), but it wasn't
> clear how to re-run it.
>
> —
> <https://github.com/spotify/docker-client/pull/267#issuecomment-154525174>
> .
>



-- 
David Xia
Software Engineer
dxia@spotify.com

mattnworb:
@cap10morgan agreed those failures are transient, and this change wouldn't cause any problems with that functionality anyway. 

Thanks for the contribution!

PR #271, opened 2015-11-07
Only run tests that push when explicitly enabled

### LINKED ISSUE #270: Transient failures from pushing images in DefaultDockerClientTest
These two tests in DefaultDockerClientTest often fail transiently when run in CircleCI

> testPushPrivateImageWithAuth - in com.spotify.docker.client.DefaultDockerClientTest
> testPushPublicImageWithAuth - in com.spotify.docker.client.DefaultDockerClientTest

error is `429 Too Many Requests` from the registry

> com.spotify.docker.client.ImagePushFailedException: Image push failed: dxia4/docker-client-test-push-private-image-with-auth: ProgressMessage{id=null, status=null, stream=null, error=Post https://registry-1.docker.io/v2/dxia4/docker-client-test-push-private-image-with-auth/blobs/uploads/: token auth attempt for registry: https://auth.docker.io/token?account=dxia4&scope=repository%3Adxia4%2Fdocker-client-test-push-private-image-with-auth%3Apush%2Cpull&service=registry.docker.io request failed with status: 429 Too Many Requests, progress=null, progressDetail=null}

It doesn't help that we run each test twice, in parallel, with DOCKER_HOST set to `tcp://...` and `unix://..`, and the issue is compounded if there happen to be more than branch being tested at the same time.

It would be nice to fix these issues to avoid transient errors on pull requests from contributors. A few ideas:

- do we really need to run the same tests with both `DOCKER_HOST=tcp://...` and `DOCKER_HOST=unix://..` still?
- do we even need to be pushing to the docker.io registry to hit whatever logic the test is trying to test?
- if answer to both above is yes, perhaps we could add some logic to the test to cycle through a handful of repository-name / authentication credentials (not sure which of these being used too concurrently is causing the 429) so that two runs of the test at roughly the same time don't use the same pair.

davidxia:
because Docker Hub rate limits pushes.

Fixes #270

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -362,4 +362,15 @@
       </plugin>
     </plugins>
   </reporting>
+
+  <profiles>
+    <profile>
+      <id>test-pushes</id>
+      <activation>
+        <property>
+          <name>env.TEST_PUSHES</name>
+        </property>
+      </activation>
+    </profile>
+  </profiles>
 </project>
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -150,6 +150,7 @@ public class DefaultDockerClientTest {
       "dxia4/docker-client-test-push-private-image-with-auth";
 
   private static final boolean CIRCLECI = !isNullOrEmpty(getenv("CIRCLECI"));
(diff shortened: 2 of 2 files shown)


davidxia:
We can make the build server that releases and uploads to sonatype enable these 2 tests.

davidxia:
closed in favor or #278

PR #276, opened 2015-11-10
cleanup duplicated code in DefaultDockerClient

mattnworb:
There are a handful of places where code is duplicated with just a small change in the HTTP method or the WebTarget resource path or parameters.


davidxia:
:+1:

PR #278, opened 2015-11-10
Add integration tests for pushing images

davidxia:
We start a docker registry as a container and push to that


davidxia:
@mattnworb

src/test/java/com/spotify/docker/it/PushIT.java
```diff
@@ -0,0 +1,199 @@
...
+    System.out.printf("- %s\n", testName.getMethodName());
+  }
+
+  @Test
+  @CreateContainer(
+      image = REGISTRY_IMAGE,
+      name = REGISTRY_NAME,
+      portBindings = @PortBinding(hostPort = 5000, containerPort = 5000, protocol = "tcp"),
+      binds = {
+          "/Users/david/src/docker-client/src/test/resources/dockerRegistry/auth:/auth",
```
davidxia: How can I make this path dynamic?

src/test/java/com/spotify/docker/it/PushIT.java
```diff
@@ -0,0 +1,199 @@
...
+          "/Users/david/src/docker-client/src/test/resources/dockerRegistry/certs:/certs"
+      },
+      env = {
```
davidxia: Doesn't seem like there's a way to have a private static final String[] here? http://stackoverflow.com/a/2065973/553994

src/test/resources/dockerRegistry/certs/domain.crt
```diff
@@ -0,0 +1,38 @@
+-----BEGIN CERTIFICATE-----
```
davidxia: Docker registry doesn't allow authentication unless it's over TLS: https://github.com/docker/distribution/blob/master/docs/deploying.md#native-basic-auth

davidxia:
Hm not sure this will pass on other build agents. I had to create my docker vm with `--engine-insecure-registry localhost:5000`

`docker-machine create -d virtualbox --virtualbox-boot2docker-url https://github.com/boot2docker/boot2docker/releases/download/v1.8.3/boot2docker.iso --virtualbox-cpu-count 2 --engine-insecure-registry localhost:5000 1.8.3`

src/test/java/com/spotify/docker/it/PushIT.java
KostyaSha: `@TempRule` that produced temp folder for binding

KostyaSha:
@davidxia i'm working on additional jenkins plugin and thought to use DinD (docker-in-docker) tests. I.e. point to any docker daemon, then run container with required daemon that will run tests with any envs.

src/test/java/com/spotify/docker/Polling.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
```
KostyaSha: 2015

src/test/java/com/spotify/docker/it/PushIT.java
```diff
@@ -0,0 +1,199 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
```
KostyaSha: 2015

src/test/java/com/spotify/docker/it/PushIT.java
KostyaSha: Maybe migrate to List?

src/test/java/com/spotify/docker/it/PushIT.java
KostyaSha: Ah, it's annotation. Interesting what was the decision for having annotation vs code call to some custom `@Rule`.

mattnworb: Can the path be relative to the build directory? Otherwise if the annotation only supports absolute paths, then I think that is an argument against using annotations for this sort of thing

KostyaSha: Btw, any path in binding is relative to docker daemon configuration...

mattnworb:
Does `PushIT` replace the push tests in `DefaultDockerClientTest`? Can you remove those if so?

Why start the registry in the JUnit code of the IT as opposed to circle.sh (or another shell script)? If it is awkward to start it in Java then I think falling back to the shell is reasonable.

src/test/java/com/spotify/docker/Polling.java
davidxia: The linter might check this against the java.header file which is 2014. I can do this in a later commit if changing to 2015 breaks the build

davidxia:
@mattnworb These ITs can be run every build without fear of rate limiting. The docker hub tests are probably still good to keep as the hub has ACLs for private repos which I can't easily mimic here.  We can just run those tests on release builds as done here https://github.com/spotify/docker-client/pull/271

src/test/java/com/spotify/docker/it/PushIT.java
davidxia: Not sure I can use a list here. Annotations are evaluated at compile-time.

src/test/java/com/spotify/docker/it/PushIT.java
davidxia: @mattnworb Yea I used annotations because it looked nice to be able to do this https://github.com/spotify/docker-client/blob/master/src/test/java/com/spotify/docker/test/DockerContainer.java#L44. But there are limitations with it. I might try using `@Rule` instead.

davidxia:
I think this approach is better https://github.com/spotify/docker-client/pull/280

mattnworb:
`java.header` validates that the license contains `Copyright (c) 20.*$`, so any year will match for the next 85 years.

With that said, the year in the header (and the copyright notice itself) [is largely irrelevant](http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/) so IMHO it is not worth worrying over in either direction.

davidxia:
closing in favor of #280

PR #280, opened 2015-11-11
Add integration tests for pushing images

davidxia:
We start a docker registry as a container and push to that.

We don't use @Rule because each method needs a slightly different
startup and teardown.

We don't use @CreateContainer because annotation parameters are decided
at compile-time, but it's hard to know what these parameters should be
at compile-time.

Just do it the old-fashioned way with some code at the beginning and end
of each PushIT method.


davidxia:
@mattnworb @negz

src/test/java/com/spotify/docker/it/PushIT.java
```diff
@@ -0,0 +1,296 @@
...
+    try {
+      client.push(LOCAL_IMAGE);
+    } catch (ImagePushFailedException e) {
+      if (e.getMessage().contains("no basic auth credentials")) {
+        correctExceptionThrown = true;
+      }
+    }
+    assertTrue(correctExceptionThrown);
+
+    stopAndRemoveContainer(client, containerId);
```
mattnworb: if you moved the call to `stopAndRemoveContainer` to a `@After` method, then this part here with `correctExceptionThrown` be done away with

mattnworb:
:+1: :100:

davidxia:
These ITs don't work on docker 1.6.2. The pull fails when a client that's 1.6.2 tries to pull from the registry:2 container.

PR #281, opened 2015-11-11
Fix indentation and whitespace

davidxia:


davidxia:
@mattnworb

mattnworb:
LGTM. We should add a Checkstyle rule if we want to maintain this "fix".

davidxia:
we didn't include test sources in maven-checkstyle-plugin. I can add that.

mattnworb:
:ok_hand:

PR #282, opened 2015-11-11
Fix integrationTest()

davidxia:
Don't try copying files to container if docker API version
doesn't support it


davidxia:
@mattnworb

walbus:
@davidxia It is also supported in version 1.20
http://docs.docker.com/engine/reference/api/docker_remote_api_v1.20/#extract-an-archive-of-files-or-folders-to-a-directory-in-a-container

Issue #253, opened 2015-10-16
Support labels on container


kmbulebu:
Docker create a container API provides facilities for assigning metadata via labels to containers. Enhancement would add some matching support in docker-client:

1) Enhance DockerClient.createContainer and ContainerConfig to support adding labels.
2) Enhance DockerClient.inspectContainer and ContainerInfo to include labels.
3) Enhancer DockerClient.listContainers and ListContainersParam to support querying by label.
4) Enhance Container returned by DockerClient.listContainers to include labels.

davidxia:
@kmbulebu sounds like something that'd be good to have. We're open to PRs :)

kmbulebu:
I'm open to submitting. It's in my backlog :)

kmbulebu:
Step 1, creating a container with labels was addressed with commit 6109390.

kmbulebu:
PR #285 addresses Steps 3 and 4. That should close it out.

PR #285, opened 2015-11-17
Added list containers filter by label.

kmbulebu:
Finishing out the tasks to support labels on container, #253.


davidxia:
@kmbulebu Thanks! Can you write a test for this?

davidxia:
replaced by #313

PR #287, opened 2015-11-19
testAttachContainer: add sleep to avoid hanging

### LINKED ISSUE #273: DefaultDockerClientTest testAttachContainer hangs forever with Docker API version 1.21
Just leaving this here for future reference, but this test seems to block forever when running against Docker API version 1.21 (Docker 1.9.0).

I took a look in the debugger and it seems like the blocking occurs after the actual logs from the (no longer running) container are read and returned. The log output from the container's command of `ls -la` is consumed and returned from LogReader, but then [the next call to](https://github.com/spotify/docker-client/blob/49130141d4070013881f16a46c1baf9de473bb47/src/main/java/com/spotify/docker/client/LogReader.java#L51)

```
final int n = ByteStreams.read(stream, headerBytes, 0, HEADER_SIZE);
```

blocks forever. Seems like the daemon is holding the connection open but not returning any more bytes.

Haven't had a chance yet to test this on other versions.

mattnworb:
Seems like if the container is stopped, then attaching to it will hang
forever as there will be no more output.

Future TODO: docker-client should be smarter about attaching to a
stopped container.

Resolves #273

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -23,6 +23,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import com.google.common.io.Resources;
 import com.google.common.util.concurrent.SettableFuture;
+
 import com.fasterxml.jackson.databind.util.StdDateFormat;
 import com.spotify.docker.client.DockerClient.AttachParameter;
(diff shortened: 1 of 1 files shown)


mattnworb:
@davidxia

davidxia:
:+1: I'll open an issue to track the TODO

davidxia:
https://github.com/spotify/docker-client/issues/288

PR #290, opened 2015-11-22
Add support for extra host

tdomzal:
Allow adding of entries to containers /etc/hosts with
host config 'extra hosts' option. It is equivalent
of 'docker run --add-host .. '.

Signed-off-by: Tomasz Domzal <tdomzal@gmail.com>

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -1332,43 +1332,79 @@ public class DefaultDockerClientTest {
   }
 
   @Test
-  public void testVolumesFrom() throws Exception {
+  public void testExtraHosts() throws Exception {
+
+    assumeTrue("Docker API should be at least v1.15 to support Container Creation with " +
+        "HostConfig ExtraHosts, got " + sut.version().apiVersion(),
+         compareVersion(sut.version().apiVersion(), "1.15") >= 0);
+
     sut.pull(BUSYBOX_LATEST);
 
-    final String volumeContainer = randomName();
-    final String mountContainer = randomName();
+    final HostConfig expected = HostConfig.builder()
+     .extraHosts("extrahost:1.2.3.4")
+     .build();
 
-    final ContainerConfig volumeConfig = ContainerConfig.builder()
-        .image(BUSYBOX_LATEST)
-        .volumes("/foo")
-        .cmd("touch", "/foo/bar")
-        .build();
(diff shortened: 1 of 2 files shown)


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1332,43 +1332,79 @@ public void testPauseContainer() throws Exception {
...
+    assertThat(actual.extraHosts(), equalTo(expected.extraHosts()));
 
     final String logs;
-    try (LogStream stream = sut.logs(info.id(), stdout(), stderr())) {
-      logs = stream.readFully();
+    try (LogStream stream = sut.logs(id, stdout(), stderr())) {
+        logs = stream.readFully();
     }
-    assertThat(logs, containsString("bar"));
+    assertThat(logs, containsString("1.2.3.4"));
```
mattnworb: is this just testing that the logs contain the IP address `1.2.3.4` from adding it to `/etc/hosts`? I would worry that something might change in the future with regard to logging that string where the actual functionality still works but the test fails.

mattnworb:
Besides comment above looks good to me :+1:

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
tdomzal: Not logs in general, but stdout from cmd to be exact.
My intention was checking contents of /etc/hosts in container so figured simple method would be just 'cat' it's contents and grep to leave extrahost line only.
So test executes single piped command via shell: 

    cat /etc/hosts | grep extrahost

and expects single line with `1.2.3.4` in stdout.

Maybe those small corrections would make it more explicit and rebust?
* change 'sut.logs(...)' to read stdout only (to exclude possibility that stderr  contains the same IP pattern)
* rename variable 'logs' to 'stdoutLogs' to express it's purpose - checking stdout only
* make sure stdoutLogs will contain single line matching string with `1.2.3.4` and ending with `extrahost`

What do you think ?

mattnworb: Looks like I misinterpreted this part of the test - `cat /etc/hosts | grep extrahost` and checking the output sounds robust to me.

I misinterpreted this as streaming the logs of the original container starting up

tdomzal: Great. 
Thanks !

PR #292, opened 2015-11-24
Fix compilation error

davidxia:
Also update deprecated code and fix some checkstyle errors

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -259,7 +259,7 @@ public class DefaultDockerClient implements DockerClient, Closeable {
                                              builder.dockerCertificates.hostnameVerifier());
     }
 
-    final RegistryBuilder registryBuilder = RegistryBuilder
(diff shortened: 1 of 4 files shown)


davidxia:
@mattnworb

mattnworb:
jinx https://github.com/spotify/docker-client/commit/3d7ba161ec637e365f8bad4aa2dea76ea164dbac

mattnworb:
:+1: 

I think these checkstyle rules that are currently set to warnings, e.g.

```xml
<module name="LeftCurly">
      <!-- Checks for placement of the left curly brace ('{'). -->
      <property name="severity" value="warning"/>
</module>
```

should have the warning part removed, otherwise contributors will probably keep committing violations

davidxia:
@mattnworb Ha I didn't realize you also had one. Want to take this PR one since it has extra stuff?

Do you mean we should set the checkstyle severity to error?

mattnworb:
> Do you mean we should set the checkstyle severity to error? If so, just that one or more?

Yes, sorry for not being clear. I think if those whitespace-type rules are only severity warning then they would be constantly ignored. I'll add a PR for changing that.

davidxia:
cool, I'll merge this then

PR #293, opened 2015-11-24
change many checkstyle rules from warning to error

mattnworb:
Some checkstyle rules were previously set to severity of warning instead
of error, meaning that any violations of those rules did not fail the
Maven or CircleCI build.

Given that a lot of these are not controversial, and that we would
prefer if contributions to the project do not introduce new Checkstyle
warnings (where the warning output from the plugin may get lost in
between all the other output from testing in CircleCI), change these
rules to error severity.


mattnworb:
@davidxia

rohansingh:
:+1:

davidxia:
:+1:

PR #296, opened 2015-11-29
Added load/save support

yduchesne:
Hi, I've added support for load/save, with the accompanying tests. A few notes:

- DockerClient.load allows loading the tarball of an image as a stream. An image name must be passed in, and the load implementation internally tags the just loaded image with the provided name. The load implementation is based on the pull resource of the  Docker API, which takes a fromSrc parameter, indicating it to treat the request's body as the .tar payload of the image.

The above mimick the load and save commands of Docker's command-line (hence the naming).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
   <groupId>com.spotify</groupId>
   <artifactId>docker-client</artifactId>
-  <version>3.1.9</version>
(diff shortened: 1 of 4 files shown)


yduchesne:
Hmm. Got CI environment-related failures methinks: https://circleci.com/gh/spotify/docker-client/1019

I've hack the tests and will submit another pull request.

PR #297, opened 2015-11-29
Added load/save support

yduchesne:
Hi, I've added support for load/save, with the accompanying tests. A few notes:

- DockerClient.load allows loading the tarball of an image as a stream. An image name must be passed in, and the load implementation internally tags the just loaded image with the provided name. The load implementation is passed on Docker API's pull, which takes a fromSrc parameter, indicating it to treat the request's body as the .tar payload of the image.

The above mimick the load and save commands of Docker's command-line (hence the naming).

Also, I've modified the tests to work around some failures I observed in the CI environment (failures not related to my changes). See: https://circleci.com/gh/spotify/docker-client/1019


yduchesne:
Closing this, will fix conflicts pertaining to latest master state.

PR #298, opened 2015-11-29
Added load/save support

yduchesne:
Hi, I've added support for load/save, with the accompanying tests. A few notes:

DockerClient.load allows loading the tarball of an image as a stream. An image name must be passed in, and the load implementation internally tags the just loaded image with the provided name. The load implementation is passed on Docker API's pull, which takes a fromSrc parameter, indicating it to treat the request's body as the .tar payload of the image.

DockerClient.save allows pulling the .tar stream of an image (and saving it locally or whatever).

PR #300, opened 2015-12-03
Added load/save support (second attempt)

yduchesne:
Hi guys, me again. I think I've reach the limits of my Git skills (still need honing those).  Thought I'd do better than my previous attempt (see: https://github.com/spotify/docker-client/pull/298) but could not get  the squashing right.

Aside from the squashing issue, I've factored in all comments. Here's a reminder of what the contribution is about:

- DockerClient.load allows loading the tarball of an image as a stream. An image name must be passed in, and the load implementation internally tags the just loaded image with the provided name. The load implementation is passed on Docker API's pull, which takes a fromSrc parameter, indicating it to treat the request's body as the .tar payload of the image.

You'd be very kind to accept this contribution and rest assured I'll try to polish my Git skills for next time (I look forward to using your client in a current project, and would not hesitate contributing to it again).


yduchesne:
BTW, one thing I don't get is that I dont see the "Fixed CircleCI" in my commits in my branch after my rebasing/squashing attempt. Yet it sill appears above. Not sure where that's taken from.

mattnworb:
@yduchesne sounds like maybe you rebased/squashed locally but didn't push the updated branch to your fork? You would need to do `git push -f` to update it

yduchesne:
That's the point, I did... So I'm not sure how the history above is "composed". It does not seem like an actual representation of the git repo's history. Strange.

davidxia:
@yduchesne It was a bit tricky, but I've opened a new PR here https://github.com/spotify/docker-client/pull/301 with all the commits squashed.

PR #301, opened 2015-12-03
Add load/save support

davidxia:
* DockerClient.load allows loading the tarball of an image as a stream. An
image name must be passed in, and the load implementation internally
tags the just loaded image with the provided name. The load
implementation is passed on Docker API's pull, which takes a fromSrc
parameter, indicating it to treat the request's body as the .tar payload
of the image.

* DockerClient.save allows pulling the .tar stream of an image (and saving
it locally or whatever).

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -20,9 +20,9 @@
 package com.spotify.docker.client;
 
 import com.google.common.base.Optional;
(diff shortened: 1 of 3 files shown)


davidxia:
@mattnworb @yduchesne

mattnworb:
:+1:

yduchesne:
That's great guys. Thanks a lot for the help and accepting my contribution.

mattnworb:
Thank you @yduchesne for the contribution!

PR #304, opened 2015-12-08
Add RANDOM_AVAILABLE_PORT for convenience/clarity.

negz:
The API to bind a random available port (as opposed to an explict one) is currently like `PortBinding.of("0.0.0.0", "")`, where the empty string (or int 0) will result in Docker picking a port.

This allows API callers to get the same effect in a more readable fashion using `PortBinding.of("0.0.0.0", PortBinding.RANDOM_AVAILABLE_PORT)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,12 @@ for (String port : ports) {
     hostPorts.add(PortBinding.of("0.0.0.0", port));
     portBindings.put(port, hostPorts);
 }
+
+// Bind container port 443 to an automatically allocated available host port.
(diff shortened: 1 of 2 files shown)


davidxia:
:+1: Perhaps also mention it in the README if that helps?

negz:
Will do!

mattnworb:
LGTM.

From an API perspective could also have something like `PortMapping.randomPort(String ip)`

gimaker:
+1 on matts comment on a `randomPort()` type function. That's easier to discover than a final static

davidxia:
replaced by #311

PR #306, opened 2015-12-10
Basic support for the network API.

teetrinkers:
Hi, I implemented basic support for the network API introduced in 1.21 / docker 1.9. It doesn't do everything yet, but allows to list, create and remove networks.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -19,48 +19,23 @@
 
 package com.spotify.docker.client;
 
-import com.google.common.base.Optional;
-import com.google.common.base.Preconditions;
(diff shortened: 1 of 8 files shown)


teetrinkers:
The tests are failing because CircleCI is still on docker 1.8. Can you update it?

davidxia:
@teetrinkers Thanks for the PR. Can you use an `assumeTrue` for the version issue? https://github.com/spotify/docker-client/blob/master/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java#L1198

We're probably not going upgrade our CircleCI setup to docker 1.9 because we're still using the previous docker version for production and we'd like to not deviate too much.

davidxia:
#307 Looks more full featured. Let's replace this with that one?

teetrinkers:
Hi, I just came back from christmas. #307 looks great!

PR #313, opened 2015-12-22
Support listing containers with a label filter

davidxia:


davidxia:
replaces #285

davidxia:
@mattnworb

src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -989,6 +989,35 @@ public static ListContainersParam containersCreatedBefore(final String id) {
...
+    }
+    
+    /**
+     * Show containers with a label
+     *
+     * @return ListContainersParam
+     */
+    public static ListContainersParam withLabel(String key) {
+      try {
+        return create("filters", URLEncoder.encode("{\"label\":[\"" + key + "\"]}", "UTF-8"));
```
mattnworb: seems like this could just call `withLabel(key, null)` and then the other version could check if value is null or not

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1884,8 +1887,55 @@ public void testLabels() throws DockerException, InterruptedException {
     sut.startContainer(id);
 
     ContainerInfo containerInfo = sut.inspectContainer(id);
-    assertThat(containerInfo.config().labels().keySet(), contains("name", "version"));
-    assertThat(containerInfo.config().labels().values(), contains("starship", "1.6.2"));
+    assertThat(containerInfo.config().labels().keySet(), containsInAnyOrder("name", "foo"));
+    assertThat(containerInfo.config().labels().values(), containsInAnyOrder("starship", "bar"));
```
mattnworb: why not assert the mapping between the key and value instead?

    assertThat(containerInfo.config().labels(), hasEntry("name", "starship"));
    assertThat(containerInfo.config().labels(), hasEntry("foo", "bar"));

or a simple equality test

    assertThat(containerInfo.config().labels(), is(labels));

mattnworb:
looks good overall

davidxia:
Thanks. I included your suggested changes.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1883,9 +1886,55 @@ public void testLabels() throws DockerException, InterruptedException {
...
+    final String name2 = randomName();
+    final ContainerCreation creation2 = sut.createContainer(config2, name2);
+    final String id2 = creation2.id();
+
+    // Start the second container
+    sut.startContainer(id2);
+
+    ContainerInfo containerInfo2 = sut.inspectContainer(id2);
+    assertThat(containerInfo2.config().labels().keySet(), containsInAnyOrder("name", "foo"));
+    assertThat(containerInfo2.config().labels().values(), containsInAnyOrder("starship", "baz"));
```
mattnworb: same thing here too

mattnworb:
this closes https://github.com/spotify/docker-client/issues/253 too i think

PR #310, opened 2015-12-22
Provides basic filtering of container list by labels

itzg:
...and also populates the `Node` field of the `ContainerInfo` when accessing a Swarm cluster.

PR #317, opened 2015-12-22
Add support for Swarm node content in ContainerInfo response

itzg:
This is a rebase of #310 since #313 covers the label filtering part.

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/ContainerInfo.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerInfo.java
@@ -53,6 +53,10 @@ public class ContainerInfo {
   @JsonProperty("MountLabel") private String mountLabel;
   @JsonProperty("Volumes") private ImmutableMap<String, String> volumes;
   @JsonProperty("VolumesRW") private ImmutableMap<String, Boolean> volumesRW;
+  /**
+   * This field is an extension defined by the Docker Swarm API, therefore it will only be populated when
+   * communicating with a Swarm cluster.
+   */
   @JsonProperty("Node") private Node node;
 
   public String id() {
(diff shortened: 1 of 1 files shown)


src/main/java/com/spotify/docker/client/messages/ContainerInfo.java
```diff
@@ -53,6 +53,7 @@
   @JsonProperty("MountLabel") private String mountLabel;
   @JsonProperty("Volumes") private ImmutableMap<String, String> volumes;
   @JsonProperty("VolumesRW") private ImmutableMap<String, Boolean> volumesRW;
+  @JsonProperty("Node") private Node node;
```
davidxia: Maybe add a comment here that this field is only used when using docker swarm which extends existing docker API behavior?

itzg: I was considering that, but there wasn't a precedent for comments elsewhere in this file ;) Will do.

src/main/java/com/spotify/docker/client/messages/ContainerInfo.java
```diff
@@ -252,6 +265,64 @@ public String toString() {
...
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+      Node node = (Node) o;
+      return com.google.common.base.Objects.equal(id, node.id) &&
```
davidxia: Can we use `import com.google.common.base.Objects;` for these?

davidxia:
LGTM other than my comment and the checkstyle error.

src/main/java/com/spotify/docker/client/messages/ContainerInfo.java
itzg: I'll blame IntelliJ for that one. New commit in just a sec.

PR #318, opened 2015-12-23
Update README about testing on Docker API 1.20

davidxia:
Our jenkins server gets test errors on 1.18 so we now
test on 1.20

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 This is a simple [Docker](https://github.com/docker/docker) client written in Java.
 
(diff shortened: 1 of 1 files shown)


davidxia:
@mattnworb @rohansingh

mattnworb:
If we have errors on 1.18, but we use docker 1.6.x internally, isn't that a problem? (1.6.x has the 1.18 api version)

It would be nice if we could set up some sort of matrix CI job in Circle or Travis to test this against each version of the docker remote API.

davidxia:
obsolete

PR #321, opened 2016-01-11
Add DockerHost class

davidxia:


davidxia:
This moves this class https://github.com/spotify/helios/blob/master/helios-testing/src/main/java/com/spotify/helios/testing/DockerHost.java#L18 and this class https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/servicescommon/DockerHost.java#L29 from Helios to docker-client.

davidxia:
@mattnworb @gimaker @rohansingh

mattnworb:
lgtm

davidxia:
Tests passed but I canceled because I used rebuild with ssh.

PR #323, opened 2016-01-12
added try/catch on ObjectMapperProvider

MarcoLotz:
Should a dependency conflict happens, no information can be extracted from ObjectMapperProvider.java due to the static block. Suggest adding a try/catch and log block to be able to find out what is causing exceptions.

Following the information on:
http://stackoverflow.com/questions/7325579/java-lang-noclassdeffounderror-could-not-initialize-class-xxx.

### DIFF
--- a/src/main/java/com/spotify/docker/client/ObjectMapperProvider.java
+++ b/src/main/java/com/spotify/docker/client/ObjectMapperProvider.java
@@ -42,9 +42,14 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.ext.ContextResolver;
 import javax.ws.rs.ext.Provider;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 @Provider
 @Produces(MediaType.APPLICATION_JSON)
 public class ObjectMapperProvider implements ContextResolver<ObjectMapper> {
+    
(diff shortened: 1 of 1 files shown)


davidxia:
@MarcoLotz, thanks. Can you fix up the indentation in the try block?

MarcoLotz:
The exception that I was getting was a MultiException cause by having an older version of Jackson as a dependency of a parent project.
It looked like 

com.spotify.docker.client.DockerException: java.util.concurrent.ExecutionException: javax.ws.rs.ProcessingException: Unexpected error during response processing.

And

mattnworb:
Is MultiException something from Jetty? It sounds to me like the exception is being thrown correctly (the `VerifyError`) but your app server (or wherever this MultiException is coming from) is either swallowing it or not giving you the stacktrace that would have pointed to the static initializer in `ObjectMapperProvider`.

MarcoLotz:
@mattnworb you are right, it is something from jetty. Just the lack of that block was making it harder to detect. This may happen to other people in the future, due to dependency conflicts (updated from docker-client 2.5 to 3.5, and there's a lot of differences in the dependencies).

mattnworb:
I suppose logging the throwable won't help in every case, for instance if someone doesn't have slf4j set up, but it seems harmless so :+1:

PR #326, opened 2016-01-15
Add public BuildParam methods

davidxia:
Replace package-private BuildParam instances with
public static factory methods. We do this with the other
Param classes.


davidxia:
@mattnworb @rohansingh

rohansingh:
:+1:

PR #329, opened 2016-01-28
Updated ContainerConfig for API 1.21 and added imageid to Container

MarcoLotz:
Added fields to container inspection in order to match Docker API 1.21
Added imageid to Container, in order to match the docker API 1.21

No removals. Should work properly with previous docker versions.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/Container.java
+++ b/src/main/java/com/spotify/docker/client/messages/Container.java
@@ -35,13 +35,14 @@ public class Container {
   @JsonProperty("Id") private String id;
   @JsonProperty("Names") private ImmutableList<String> names;
   @JsonProperty("Image") private String image;
+  @JsonProperty("ImageID") private String imageId;
   @JsonProperty("Command") private String command;
   @JsonProperty("Created") private Long created;
   @JsonProperty("Status") private String status;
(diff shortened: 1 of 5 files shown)


MarcoLotz:
testInspectContainerWithSecurityOpt test passes with no errors here. Can the timeout be something related with the testing env there?

davidxia:
@MarcoLotz Tests are sometimes flakey. It passed now though :)

davidxia:
@MarcoLotz Looks great. Do you mind adding some tests for these new properties? See here for an example https://github.com/spotify/docker-client/blob/master/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java#L1218

MarcoLotz:
Added tests and modified getter syntax to be inspection.mounts() instead of inspection.getMounts() from the autogenerated getter.

davidxia:
@MarcoLotz Thanks for adding the tests. Can you undo the changes in DefaultDockerClientTest that are just formatting changes though? It makes the diff look bigger than it actually is.

davidxia:
And squashing into one commit would be great.

MarcoLotz:
@davidxia  Done :)

davidxia:
Thanks

PR #330, opened 2016-01-29
Networks information in inspectContainer

damianopezzotti:
Returns information about networks associated to the container in _inspectContainer_ flow, as explained [here](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#inspect-a-container)

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/NetworkSettings.java
+++ b/src/main/java/com/spotify/docker/client/messages/NetworkSettings.java
@@ -41,6 +41,7 @@ public class NetworkSettings {
   @JsonProperty("PortMapping") private ImmutableMap<String, Map<String, String>> portMapping;
   @JsonProperty("Ports") private Map<String, List<PortBinding>> ports;
   @JsonProperty("MacAddress") private String macAddress;
+  @JsonProperty("Networks") private ImmutableMap<String, AttachedNetwork> networks;
 
   private NetworkSettings(final Builder builder) {
     this.ipAddress = builder.ipAddress;
@@ -50,6 +51,7 @@ public class NetworkSettings {
     this.portMapping = builder.portMapping;
     this.ports = builder.ports;
     this.macAddress = builder.macAddress;
+    this.networks = builder.networks;
   }
 
   @SuppressWarnings("unused")
@@ -79,9 +81,13 @@ public class NetworkSettings {
   public Map<String, List<PortBinding>> ports() {
     return (ports == null) ? null : Collections.unmodifiableMap(ports);
   }
-  
+
   public String macAddress() {
     return macAddress;
+  }
+
+  public Map<String, AttachedNetwork> networks() {
+    return (networks == null) ? null : Collections.unmodifiableMap(networks);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


davidxia:
@damianopezzotti Thanks for the PR. I'll take a look soon. Could you squash into one commit?

damianopezzotti:
@davidxia squash done

src/main/java/com/spotify/docker/client/messages/NetworkSettings.java
```diff
@@ -215,6 +228,17 @@ public Builder macAddress(final String macAddress) {
       return this;
     }
 
+    public Builder networks(final Map<String, AttachedNetwork> networks) {
+      if (networks != null) {
+        final ImmutableMap.Builder<String, AttachedNetwork> builder = ImmutableMap.builder();
```
davidxia: I think you can replace everything in this `if` with `this.networks = ImmutableMap.copyOf(networks);`

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -2108,6 +2109,17 @@ public void testNetworksConnectContainer() throws Exception {
...
+    final ContainerInfo containerInfo = sut.inspectContainer(containerCreation.id());
+    assertThat(containerInfo.networkSettings().networks().size(), is(2));
+    final AttachedNetwork attachedNetwork =
+        containerInfo.networkSettings().networks().get(networkName);
+    assertThat(attachedNetwork, is(notNullValue()));
+    assertThat(attachedNetwork.endpointId(), is(notNullValue()));
+    assertThat(attachedNetwork.gateway(), is(notNullValue()));
+    assertThat(attachedNetwork.ipAddress(), is(notNullValue()));
+    assertThat(attachedNetwork.ipPrefixLen(), is(notNullValue()));
+    assertThat(attachedNetwork.macAddress(), is(notNullValue()));
```
davidxia: Can we also test `ipv6Gateway()`, `globalIPv6Address()`, and `globalIPv6PrefixLen()` here?

davidxia:
It'd be nice to add tests for attaching the container to a network and inspecting the container to see it works, but we can add this later.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
damianopezzotti: I think that result depends from machine setup (if ipv6 is enabled or not). On my machine `ipv6Gateway()` and `globalIPv6Address()` return void String and `globalIPv6PrefixLen()` returns 0.
I added a check that verifies that `ipv6Gateway()` and `globalIPv6Address()` return not null values and `globalIPv6PrefixLen()` >=0. Is it ok?

damianopezzotti:
@davidxia Test `testNetworksConnectContainer` creates a network, attach a container to the network and then inspect it.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
davidxia: Yea that seems good.

damianopezzotti: pushed

src/main/java/com/spotify/docker/client/messages/NetworkSettings.java
damianopezzotti: pushed

davidxia:
@damianopezzotti Thanks!

Issue #327, opened 2016-01-19
Unit tests failing on windows


sayarg:
pulling the code and running test target on windows as is, fails on a few unit tests. this is in addition to the integration test, which i am out-scoping for the sake of this issue.

specifically:
AuthConfigTest fails on path issues e.g. java.nio.file.InvalidPathException: Illegal char <:> at index 2: /H:/workspace/spotify/docker-client-master/target/test-classes/dockerConfig/fullDockerCfg

DockerHostTest need to be using different defaults for windows
and CompressedDirectoryTest. CompressedDirectory need to use different values when creating the regex matches on windows for the ignore file to work

sayarg:
hello, any thoughts on this? ok to merge the PR?

mattnworb:
@sayarg thanks for the fix and contribution and sorry for the delay in reviewing. I think with #331 merged this closes this issue, right?

PR #331, opened 2016-01-29
Unit tests windows

sayarg:
Fixing https://github.com/spotify/docker-client/issues/327

Includes:
1. CompressedDirectory ignore pattern needs to be os specific
2. DockerHostTest needs default env needs to be os specific
3. AuthConfigTest file paths needs to be os specific

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
   <groupId>com.spotify</groupId>
   <artifactId>docker-client</artifactId>
-  <version>3.4.1-SNAPSHOT</version>
+  <version>3.5.0-SNAPSHOT</version>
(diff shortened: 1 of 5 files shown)


mattnworb:
Looks fine to me. Seems like it would have been wise to just always refer to `File.pathSeparator` from the beginning. Thanks @sayarg !

sayarg:
only point for os utils that you can look to mock it if you wanted in the future. as soon as there's more than one place where you'd look at file.seperator it made sense to give something that all these places use and as above if you want to test how you behave on win for example you can do that more easily.

ok thanks vm! moving on now, will look at restart policy.

PR #334, opened 2016-02-03
Allow specifying LogConfig when creating a container

stephan-huttenhuis:
See:
https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#create-a-container
https://docs.docker.com/engine/reference/logging/overview/

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/LogConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/LogConfig.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.docker.client.messages;
+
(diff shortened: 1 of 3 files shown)


davidxia:
Were the permissions on `HostConfig.java` accidentally changed?

src/main/java/com/spotify/docker/client/messages/LogConfig.java
```diff
@@ -0,0 +1,96 @@
...
+  public Map<String, String> logOptions() {
+    return (logOptions == null) ? null : Collections.unmodifiableMap(logOptions);
+  }
+
+  public void logOptions(final Map<String, String> logOptions) {
+    this.logOptions = logOptions;
+  }
+
+  public static LogConfig create(final String logType, final Map<String, String> logOptions) {
+    final LogConfig logConfig = new LogConfig();
```
davidxia: Are `logType` and `logOptions` required? If so, specify a private constructor `LogConfig(final String logType, final Map<String, String> logOptions)` and use that here. Then remove the setters for these two attributes. This will make sure this class is never in a half-way configured state.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1539,6 +1541,43 @@ public void testExtraHosts() throws Exception {
...
+      logOptions.put("max-size", "10k");
+      logOptions.put("max-file", "2");
+      logOptions.put("labels", name);
+      
+      final LogConfig logConfig = new LogConfig();
+      logConfig.logType("json-file");
+      logConfig.logOptions(logOptions);
+        
+      final HostConfig expected = HostConfig.builder()
+      .logConfig(logConfig)
```
davidxia: nitpick: these lines should be indented 4 spaces.

davidxia:
@stephan-huttenhuis Thanks for your PR. Looks pretty good other than my comments above.

davidxia:
replaced by https://github.com/spotify/docker-client/pull/346

PR #335, opened 2016-02-03
Updated ContainerInfo hash, equals and to String to new 1.21API items

MarcoLotz:
Realized that was missing new attributes on hash, equals and toString at containerInfo methods


davidxia:
Thanks for fixing.

Issue #338, opened 2016-02-04
Support exclusion exceptions in .dockerignore


xcoulon:
Docker 1.7 introduce support for exclusion exception (similar to what exists in `.gitignore`): https://github.com/docker/docker/issues/9171, but this is support in the Docker client, which means that the context sent to the Docker daemon during a build operation is not accurate compared to the expectations.

Reference to docs: https://docs.docker.com/engine/reference/builder/#dockerignore-file

PR #339, opened 2016-02-04
Support for exclusion exception pattern in .dockerignore

xcoulon:
Using an innerclass around the PathMatcher to support the
 exclusion exception.
Removing the 'pre-visit' of a directory to make sure that
 a specific file is included.
Storing the PathMatchers in a list instead of a set because
 the order in which they are declared matters (the list should
 be read from bottom to top)
Added more files in the test case.
Also added .checkstyle in the .gitignore (this file is
  generated by the Eclipse checkstyle plugin but does not
  need to be shared)
Fixes https://github.com/spotify/docker-client/issues/338

Signed-off-by: Xavier Coulon <xcoulon@redhat.com>

### DIFF
--- a/src/main/java/com/spotify/docker/client/CompressedDirectory.java
+++ b/src/main/java/com/spotify/docker/client/CompressedDirectory.java
@@ -17,14 +17,8 @@
 
 package com.spotify.docker.client;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableSet;
-
-import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
-import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
-import org.apache.commons.compress.compressors.gzip.GzipCompressorOutputStream;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.BIGNUMBER_POSIX;
+import static org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.LONGFILE_POSIX;
 
 import java.io.Closeable;
 import java.io.IOException;
@@ -46,8 +40,14 @@ import java.util.EnumSet;
 import java.util.Set;
 import java.util.regex.Pattern;
 
-import static org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.BIGNUMBER_POSIX;
-import static org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.LONGFILE_POSIX;
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
+import org.apache.commons.compress.compressors.gzip.GzipCompressorOutputStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 
 /**
  * This helper class is used during the docker build command to create a gzip tarball of a directory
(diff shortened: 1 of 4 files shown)


xcoulon:
Hello @davidxia and @rohansingh,
I just updated this PR to resolve the merge conflicts. 
Would you have time to review it, please ?

src/main/java/com/spotify/docker/client/CompressedDirectory.java
```diff
@@ -341,4 +347,62 @@ private static int getModeFromPermissions(boolean read, boolean write, boolean e
...
+    public boolean isExclude() {
+      return this.exclude;
+    }
+
+    /**
+     * @param path the path to match
+     * @return <code>true</code> if the given {@code path} starts with the pattern or matches the
+     *         pattern
+     * 
+     * @See {@link Path#startsWith(String)}
```
davidxia: nitpick: lowercase `See` here.

src/main/java/com/spotify/docker/client/CompressedDirectory.java
```diff
@@ -341,4 +347,62 @@ private static int getModeFromPermissions(boolean read, boolean write, boolean e
     }
 
   }
+  
+  /**
+   * A decorator for the {@link PathMatcher} with a type to determine if it is an exclusion pattern
+   * or an exclude to an aforementioned exclusion. See
+   * https://docs.docker.com/engine/reference/builder/#dockerignore-file
+   */
+  private static class DockerIgnorePathMatcher {
```
davidxia: To be a decorator, should we make this `implements PathMatcher`?

src/main/java/com/spotify/docker/client/CompressedDirectory.java
```diff
@@ -341,4 +347,62 @@ private static int getModeFromPermissions(boolean read, boolean write, boolean e
...
+    /**
+     * Constructor.
+     * 
+     * @param fileSystem the current {@link FileSystem}
+     * @param pattern the exclusion or inclusion pattern
+     * @param exclude flag to indicate if the given {@code pattern } is an exclusion (
+     *        <code>true</code>) or if it is an inclusion (<code>false</code>).
+     */
+    public DockerIgnorePathMatcher(final FileSystem fileSystem, final String pattern,
+        final boolean exclude) {
```
davidxia: nit: indent up to beginning of `(` above.

xcoulon:
@davidxia thanks for the review ! I just updated the pull request with the changes that your requested ;-)

davidxia:
Thanks I haven't reviewed all of it yet, but will soon.

On Thursday, March 3, 2016, Xavier Coulon <notifications@github.com> wrote:

> @davidxia <https://github.com/davidxia> thanks for the review ! I just
> updated the pull request with the changes that your requested ;-)
>
> —
> <https://github.com/spotify/docker-client/pull/339#issuecomment-191843536>
> .
>


-- 
David Xia
Software Engineer
dxia@spotify.com

xcoulon:
@davidxia ok. I'll follow-up with your comments, then.

src/test/java/com/spotify/docker/client/CompressedDirectoryTest.java
```diff
@@ -71,8 +71,8 @@ public void testFileWithIgnore() throws Exception {
         final String name = entry.getName();
         names.add(name);
       }
-      assertThat(names, containsInAnyOrder("Dockerfile", "bin/date.sh", "subdir2/keep-me",
-                                           ".dockerignore"));
+      assertThat(names, containsInAnyOrder("Dockerfile", "bin/date.sh", "subdir2/keep.me",
```
davidxia: We should also check the size of `names` to ensure we're not including files that should be ignored.

src/main/resources/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,40 @@
+Manifest-Version: 1.0
```
davidxia: Do we need this file for the PR? Just asking because I'm not sure what the team would like for this file if we include one.

davidxia: If it's not related, we can add one in another PR.

davidxia:
Other than my comments, things look good. Thanks for updating the test.

src/main/resources/META-INF/MANIFEST.MF
xcoulon: This was not intended, I'll remove that file. Sorry about that.

src/test/java/com/spotify/docker/client/CompressedDirectoryTest.java
xcoulon: good point, I'll add that check.

xcoulon:
@davidxia I updated the patch by removing the unwanted `META-INF/MANIFEST.MF` file in `src/main/resource`.

src/test/java/com/spotify/docker/client/CompressedDirectoryTest.java
davidxia: Ah, I didn't realize this. It's good as is then.

xcoulon:
Thanks a lot, @davidxia !

PR #340, opened 2016-02-04
Adding ClientFactory abstraction to allow alternate JAXRS implementations to be used

jskjons:
We use RESTEasy in most of our applications/tooling and because the docker client currently uses the standard JAXRS client api it's hard to control which implementation gets used and it's impossible to customize the client if Resteasy gets picked. This abstractions allows us to just create RESTEasy clients and then only have 1 JAXRS implementation on our classpath. I've tested it with both Jersey and RESTEasy (not included) implementations and everything seems to work. 

I also switched the JerseyClientFactory to directly use the JerseyClientBuilder so that it'll still use Jersey even if there's an alternate JAXRS implementation on the classpath.


src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -57,11 +57,7 @@
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
 import org.glassfish.hk2.api.MultiException;
```
JWGmeligMeyling: It seems this code is still dependent on `org.glassfish.hk2.api.MultiException` and `org.glassfish.jersey.internal.util.Base64`. Are you sure that this client still works when glassfish has completely vanished from the classpath? And that the exception handling works well if `glassfish` is replaced with `resteasy`?

JWGmeligMeyling:
Great work! The PR looks quite good. There are a few remarks from my side (as a contributor, not maintainer).

It seems the code is still dependent on `org.glassfish.hk2.api.MultiException` and `org.glassfish.jersey.internal.util.Base64`. Besides that this implied that (at least some) `glassfish` dependencies are still required on the classpath, I think that exceptions may not be handled properly when another `HttpClient` implementation is used.

I would love to see a few additions:
* Add `resteasy-client` (and its Jackson companion) as an `optional` and `provided` dependency:
```xml
<dependency>
	<groupId>org.jboss.resteasy</groupId>
	<artifactId>resteasy-client</artifactId>
	<version>3.0.16.Final</version>
	<scope>provided</scope>
	<optional>true</optional>
</dependency>
```
*(In fact, this should also be done with the `glassfish` dependencies in my opinion, but that would be a breaking change for existing users, so lets stick to adding that as `<exclusion>` for now)*
* Alter the intergration tests in such as way (either through test class inheritance or for example JUnit theories or parameterized) that the same tests are ran for both the `glassfish` and `resteasy-client`. This would address my concern that some exception handling (or for example any timeout vs. non-timeout issues) are not working properly.
* Some example usage of how to use another `HttpClient` can be used and specifically how the `resteasy-client` should be configured in practice would be nice!

JWGmeligMeyling:
ping @grexe

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
jskjons: MultiException is in hk2-api which doesn't seem to conflict with RESTEasy anyways so I don't think I need to worry about getting rid of the reference, although explicitly adding the maven dependency rather than using it transitively would be good. I'll make that change.

Base64 I didn't notice and I assume would've failed if I tried to use it with registry authentication. Any objections to switching that to commons-codec?

grexe: We could also switch to Java8 and use Base64.getEncoder()

jskjons: yeah that was the first thing I thought of but thought it might be a bit presumptuous to update the jdk version in this PR :)

jskjons:
I can add the optional dependencies but I'm not sure if it really adds much value since the user still has to add those dependencies themselves anyway. Unless you're going to fully support multiple JAXRS implementations I'm not sure how much value there is in kinda-sorta supporting it. I mainly wanted to clear the roadblock preventing usage of anything else. 

If you want to full on support multiple implementations I think ideally there should be a few more breaking changes like extracting the dependencies and factories into separate modules. 

I'm happy to add an example usage of using RESTEasy with this, should that go in user_manual.md?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
JWGmeligMeyling: But resteasy does not throw hk2-exceptions, but jax-rs exceptions, so that is different right?

JWGmeligMeyling:
The dependencies should be included anyway in order to be able to test the Resteasy client. My opinion is still that this should be tested properly, including timeout and exception behaviour. So I would like to see the integration tests to be ran on both client implementations. 

The optional inclusion would mostly be a cosmetic feature indicating it can be used with other clients.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
jskjons: The only place the MultiException is used is in an multi-catch with ExecutionException which is what gets thrown by either implementation from what I can tell in the `.get()` call because it's async. In my tests it seems to propagate exceptions just fine. They get dealt with by the line `} else if (cause instanceof WebApplicationException) {...`

jskjons:
fair enough, I'll make that change today.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
JWGmeligMeyling: Ah, missed that statement. Exception handlig should be fine then! :+1:

JWGmeligMeyling:
Allright. I can have a look on Thursday to maybe add some tests if you'd like :)

jskjons:
I added the ResteasyClientFactory implementation as well as deps + docs. It turned out there was a few of the features that we were using (streaming stuff in particular) that was a bit of a pain to get working in RESTEasy but the integration tests are passing now. The Events stuff currently directly relies on jersey so it would be more difficult to get working so I just left it for now and documented that fact.

JWGmeligMeyling:
Whoa, great work! Will defenitely start using our fork 😄  Any idea on the last failing test case?

> Tests in error: 
>  testEventStreamWithUntilTime[resteasy](com.spotify.docker.client.DefaultDockerClientTest): The supplied component "org.jboss.resteasy.client.jaxrs.ResteasyClient" is not assignable from JerseyClient or JerseyWebTarget.

docs/user_manual.md
```diff
@@ -54,6 +54,7 @@ This user manual is made to correspond to Docker's [API docs][1] (e.g. [API 1.18
   * [Exec Resize](#exec-resize)
   * [Exec Inspect](#exec-inspect)
   * [Mounting volumes in a container](#mounting-volumes-in-a-container)
+  * [Using RESTEasy instead of Jersey]#(resteasy)
```
JWGmeligMeyling: Seems that the `#(` should be flipped to `(#`

jskjons:
that last test was failing because the events stuff doesn't work with resteasy so I needed to add 

`assumeThat(sut.getClient(), not(instanceOf(ResteasyClient.class)));` 

to all of those tests and missed one. Should be good now.

JWGmeligMeyling:
Great! Waiting on response from the maintainers then.

JWGmeligMeyling:
On 04 May 2016, at 17:32, Jeff Skjonsby <notifications@github.com<mailto:notifications@github.com>> wrote:


that last test was failing because the events stuff doesn't work with resteasy so I needed to add

assumeThat(sut.getClient(), not(instanceOf(ResteasyClient.class)));

to all of those tests and missed one. Should be good now.

johnflavin:
That [test failure on 1.10.3](https://travis-ci.org/spotify/docker-client/jobs/127826597) is caused by the same problem as #412. That issue (or at least the test that failed because of the issue) was fixed in #415. Rebase this PR on master and that issue should go away.

bitsofinfo:
+1

mydeadlyvenoms:
Any updates on this?

jskjons:
I haven't had time to keep the branch up to date with master and I haven't heard anything from the maintainers so it appears this isn't a priority or not a direction the maintainers want to go it. We're still internally using this branch so it still works.

mydeadlyvenoms:
I am having trouble to use the library under Glassfish / Payara due to Jersey conflicts - this might be the solution.

rmannibucau:
+1, i'd like to use CXF and JSONB instead of jackson, using portable API is really important

PR #341, opened 2016-02-05
Support for image labels

johnflavin:
Added methods for
* Reading Labels from Image objects
* Filtering image searches by Label:
    * Key only
    * Key=Value
    * Map of Key, Value pairs

Includes one test which creates images with labels and searches for them.

(This code is based almost entirely on the code for container labels. See PRs #285 and #313.)

### DIFF
--- a/src/main/java/com/spotify/docker/client/DockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DockerClient.java
@@ -19,6 +19,7 @@ package com.spotify.docker.client;
 
 import com.google.common.base.Throwables;
 
+import com.google.common.collect.ImmutableMap;
 import com.spotify.docker.client.messages.AuthConfig;
 import com.spotify.docker.client.messages.Container;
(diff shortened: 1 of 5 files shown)


src/main/java/com/spotify/docker/client/messages/ImageInfo.java
```diff
@@ -21,6 +21,7 @@
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.ImmutableMap;
```
davidxia: Unused import

davidxia:
Thanks for your PR. Things look great other than the one unused import.

johnflavin:
Fixed. Happy to contribute.

davidxia:
@johnflavin Great. Would you mind squashing into 1 commit? Then I'll merge.

johnflavin:
I've never squashed commits before, but I think I made it work.

Issue #342, opened 2016-02-06
Cookies support


sayarg:
i have a use case for providing cookies as part of the http flow. if there are no objections, i'll submit a PR. i know it's probably not a common one.

sayarg:
hi @mattnworb  @davidxia 

i've made the changes to support headers generally, could you take a look please?

https://github.com/spotify/docker-client/pull/362

davidxia:
fixed by #362

PR #343, opened 2016-02-06
add support for http cookies

sayarg:
i had to do a bit of a refactor to allow me to test for backwards compatibility for when no cookies are provided as well as with the new functionality.

addressing issue https://github.com/spotify/docker-client/issues/342

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,6 +154,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>1.10.19</version>
+      <scope>test</scope>
+    </dependency>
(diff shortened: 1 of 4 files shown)


sayarg:
@mattnworb @davidxia 

hey guys, any objections to add this cookie support?

davidxia:
What kind of cookies do you need to send to the docker daemon? Just curious.

sayarg:
@davidxia yes that sounds awkward. The cookies are for an nginx proxy I front the daemon with. The cookie usage ends there. I don't even know if the daemon takes any cookies.

davidxia:
Arbitrary header support might be the way to go in this case. Users can then set cookies in the headers.

sayarg:
@davidxia yes agreed. i'll add headers then and post.

Issue #348, opened 2016-02-16
Implement `filter` query param on `listImages`


johnflavin:
The `listImages` method is missing a feature in the Docker Remote API: filtering by image name, a.k.a. repo tag.

From the [Remote API docs](https://docs.docker.com/reference/api/docker_remote_api_v1.22/#list-images) (latest API version, though this quoted section has not changed in any recent API version):

> * filters – a JSON encoded value of the filters (a map[string][]string) to process on the images list. Available filters:
 `dangling=true`
 `label=key` or `label="key=value"` of an image label

> * filter - only return images with the specified name

Currently, `listImages` only supports the former. All `ListImagesParam`s passed into the method get appended to the request url as values for the `filters` query param. There is no way to give any value for the `filter` query param, thus no way to filter the full list of images by image name.

(See #180 for a question that appears to have this issue at its root. It looks like @bwahn is trying to query for images by name, but the `listImages` method does not allow it.) 

I think that supporting the `filter` query param and allowing `listImages` to be filtered by image name would be a useful feature. But is there a reason that it should not be implemented? On its face this feature seems to duplicate functionality: one can already use `inspectImage(name)` to query about a particular tagged image, so why support `listImages` with `filter=name`? The reason is that those two methods return different information. The former returns an `ImageInfo` and the latter a `List` of `Image`s. The information contained in either an `Image` or `ImageInfo` object doesn't completely overlap the other. (Specifically, `ImageInfo` almost contains a superset of the fields of `Image`, except that `ImageInfo` does not contain the image's `Labels`. For some reason Docker doesn't report those on `GET /images/(name)/json`, which is the request made by `inspectImage`.)

mattnworb:
> I think that supporting the filter query param and allowing listImages to be filtered by image name would be a useful feature. But is there a reason that it should not be implemented? 

I agree it would be useful. I think we'd be happy to merge a PR implementing this functionality.

bwahn:
@johnflavin Thanks for detailed that issue. Using latest version of Docker, docker-client of spotify, I'm going to test it.

johnflavin:
I did some digging, and I was totally wrong. This is already supported.

Here's why...
`listImages` takes an array of `ListImagesParam`s. Those objects are pretty much just a key and a value. Some of those params may be instances of the subclass `ListImagesFilterParam`. Only those instances of the subclass will be sent through to the request as `filters=["key1":"value1","key2":"value2",...]`. Everything else will be sent through as `key=value`.

[...]

PR #349, opened 2016-02-17
Refactored ListImagesParam and ListContainersParam

johnflavin:
After my confusion in #348, I wanted to refactor the query params for listing images and containers. There are not very many query params one can pass, and they are all explicitly listed in the Docker Remote API docs, so it should not be necessary to create `new List*Param` objects directly from the constructor for common query params, or preferably not at all. There should be explicitly named methods to create the query params that are known to exist. So I did that. (I almost made the constructors for the `List*Params` private so the explicit methods would be the *only* way to use the `List*Param`s, but I thought that might break a lot of other code for not much gain. I do think it's a good idea, but maybe too brazen for me to do unilaterally.)

While I was at it, I found that the code to serialize the `filters` values out to json was inconsistent between `listImages` and `listContainers`. So I fixed that. And several methods that should be returning `List*FilterParams` were serializing their `filters` directly out to json as if they would be the only `filters` params on a request. (This includes some methods that I wrote. *Mea culpa*.)

I added a big test for different combinations of `ListContainersParam`s.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -21,6 +21,7 @@ package com.spotify.docker.client;
 
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
 import com.google.common.io.CharStreams;
 import com.google.common.net.HostAndPort;
 
@@ -75,6 +76,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.InterruptedIOException;
 import java.io.StringWriter;
+import java.io.UnsupportedEncodingException;
 import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.net.URLEncoder;
@@ -362,25 +364,25 @@ public class DefaultDockerClient implements DockerClient, Closeable {
     WebTarget resource = resource()
         .path("containers").path("json");
 
+    final Map<String, List<String>> filters = newHashMap();
     for (ListContainersParam param : params) {
-      resource = resource.queryParam(param.name(), param.value());
-    }
-
-    return request(GET, CONTAINER_LIST, resource, resource.request(APPLICATION_JSON_TYPE));
-  }
-
-  @Override
-  public List<Image> listImages(ListImagesParam... params)
-      throws DockerException, InterruptedException {
-    WebTarget resource = resource()
-        .path("images").path("json");
-
-    final Map<String, String> filters = newHashMap();
-    for (ListImagesParam param : params) {
-      if (param instanceof ListImagesFilterParam) {
-        filters.put(param.name(), param.value());
+      if (param instanceof ListContainersFilterParam) {
+        List<String> filterValueList;
+        if (filters.containsKey(param.name())) {
+          filterValueList = filters.get(param.name());
+        } else {
+          filterValueList = Lists.newArrayList();
+        }
+        filterValueList.add(param.value());
+        filters.put(param.name(), filterValueList);
       } else {
-        resource = resource.queryParam(param.name(), param.value());
+        try {
+          final String encodedName = URLEncoder.encode(param.name(), UTF_8.name());
+          final String encodedValue = URLEncoder.encode(param.value(), UTF_8.name());
+          resource = resource.queryParam(encodedName, encodedValue);
+        } catch (UnsupportedEncodingException e) {
+          throw new DockerException(e);
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


johnflavin:
Any thoughts on this PR? I can answer questions if you have them.

johnflavin:
@mattnworb @davidxia
Any opinions on this PR?

davidxia:
Taking a look now. Sorry for the delay.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -362,25 +364,72 @@ public Info info() throws DockerException, InterruptedException {
...
+          final String encodedValue = URLEncoder.encode(param.value(), UTF_8.name());
+          resource = resource.queryParam(encodedName, encodedValue);
+        } catch (UnsupportedEncodingException e) {
+          throw new DockerException(e);
+        }
+      }
+    }
+
+    // If filters were specified, we must put them in a JSON object and pass them using the
+    // 'filters' query param like this: filters={"dangling":["true"]}
```
davidxia: Will the parameter value always be a JSON object with a list of strings as the values? The [docs here](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.18/#list-containers) imply that the exited and status filters should look like `?filters={"exited":0, "status": "running"}`, but perhaps the docs are misleading?

src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -1275,49 +1277,102 @@ public static ListContainersParam containersCreatedBefore(final String id) {
...
+    public static ListContainersParam withStatusExited() {
+      return filter("status", "exited");
+    }
+
+    /**
+     * Show exited containers.
+     *
+     * @return ListContainersParam
+     */
+    @Deprecated
```
davidxia: Say what users should switch to in javadoc.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
johnflavin: Yes, the `filters` value will always be of type `Map<String, List<String>>` (in Java parlance). I don't speak go, but apparently that is what `map[string][]string` means. See [this SO answer](http://stackoverflow.com/a/28055631/1288474).

src/main/java/com/spotify/docker/client/DockerClient.java
johnflavin: Done. Pushing now.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
johnflavin: You can see that the code I replaced was also writing out a JSON list for each key in filters. Lines 395–397:
```
generator.writeArrayFieldStart(entry.getKey());		
generator.writeString(entry.getValue());		
generator.writeEndArray();
```
But it assumed there would only ever be one value for each key, which isn't always the case.

davidxia: Gotcha, thanks.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -2063,27 +2073,63 @@ public void testExecInspect() throws DockerException, InterruptedException, IOEx
   }
 
   @Test
-  public void testExitedListContainersParam()
-      throws DockerException, InterruptedException, UnsupportedEncodingException {
+  public void testListContainers() throws DockerException, InterruptedException {
```
davidxia: yay for tests!

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -362,25 +364,72 @@ public Info info() throws DockerException, InterruptedException {
     WebTarget resource = resource()
         .path("containers").path("json");
 
+    final Map<String, List<String>> filters = newHashMap();
     for (ListContainersParam param : params) {
-      resource = resource.queryParam(param.name(), param.value());
+      if (param instanceof ListContainersFilterParam) {
```
davidxia: Perhaps move this into a private method and use it in `public List<Image> listImages(final ListImagesParam... params)` as well?

johnflavin: I considered something like that. Making one method write out filter params for images and containers would avoid a lot of duplicate code between `listContainers` and `listImages`. But the inheritance of the classes you would pass in to it seems... complicated. 

To have one method handle both types, it would have to accept a generic `ListParam`, and `ListImagesParam` and `ListContainersParam` would have to be modified to inherit from `ListParam`. That wouldn't be hard.

The part that confuses me is how to handle the `List*FilterParam`s. Right now they are subclasses of the respective non-`Filter` classes. But if we made a generic `ListParam`, it seems we would need to make a generic `ListFilterParam` subclass. But `ListImagesFilterParam` can't be both a `ListImagesParam` and a `ListFilterParam`. So which would it be? How would the methods in `ListImagesParam` and `ListContainersParam` need to be changed to support the new `List*FilterParam` classes?

I'm not saying there aren't good answers for these questions, just saying that it may be a little tricky to get right. (Or maybe it's super simple and I'm missing something. Could be that too.)

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -362,25 +364,72 @@ public Info info() throws DockerException, InterruptedException {
...
+          resource = resource.queryParam(encodedName, encodedValue);
+        } catch (UnsupportedEncodingException e) {
+          throw new DockerException(e);
+        }
+      }
+    }
+
+    // If filters were specified, we must put them in a JSON object and pass them using the
+    // 'filters' query param like this: filters={"dangling":["true"]}
+    try {
```
davidxia: We can pull the url-encoding part out like so: https://github.com/NrgXnat/docker-client/pull/1/files.

davidxia:
@johnflavin Other than my minor comments, :+1: Thanks for the PR! I can merge whenever you're ready.

johnflavin:
I just merged your patch, so once the build passes, feel free to merge away.

PR #350, opened 2016-02-17
Also close ApacheUnixSocket resources in case of exceptions

stephan-huttenhuis:
When calling ApacheUnixSocket.close three resources are closed:
1. shutdownInput();
2. shutdownOutput();
3. inner.close();

But if e.g. the first throws an exception then the other two resources are not closed. Inner.close() will not close its socket file descriptors. This can be tested by calling DockerClient.ping without an active docker daemon

### DIFF
--- a/src/main/java/com/spotify/docker/client/ApacheUnixSocket.java
+++ b/src/main/java/com/spotify/docker/client/ApacheUnixSocket.java
@@ -270,9 +270,16 @@ public class ApacheUnixSocket extends Socket {
          sleeping = false;
       }
     }
-    shutdownInput();
-    shutdownOutput();
-    inner.close();
+
+    try {
+      shutdownInput();
+    } finally {
+      try {
+        shutdownOutput();
+      } finally {
+        inner.close();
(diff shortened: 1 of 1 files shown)


davidxia:
@stephan-huttenhuis Thanks for the PR. @rohansingh Is this a good approach or should we just handle the IOExceptions in their respective methods, ie is it useful to let the caller handle `IOException` from `ApacheUnixSocket.shutdownInput()` or `shutdownOutput()` or should we just log it and not throw `IOException`?

davidxia:
@rohansingh Any thoughts on this?

rohansingh:
@davidxia I think this is the right way to go. It bubbles up the exception as before, but ensures the socket does actually get closed. I don't think `ApacheUnixSocket` should be the in the business of handling its own exceptions

mattnworb:
:+1: from me too.

I am going to close/re-open this to trigger a Travis build.

mattnworb:
this is in `docker-client:3.6.6` now

PR #351, opened 2016-02-18
Add 'Tty' flag for the execStart call.

rgrunber:
The Docker reference implementation for the remote api passes 'Tty' to
both 'Exec Create' and 'Exec Start' when the '-t' option is supplied to
'docker exec' so the option should exist in both cases.

If you look in https://github.com/docker/docker/blob/master/api/client/exec.go#L51 , you'll see this to be the case. Without the option it's still possible to interact with some shell, but some characters that have special meaning in the shell (carriage return, tab, ctrl+c) don't seem to be interpreted correctly.


davidxia:
:+1:

PR #358, opened 2016-02-28
Support event callbacks

davidxia:


mattnworb:
can't really comment on the implementation but the tests look good so :+1:

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1341,6 +1342,76 @@ public void testContainerWithCpuQuota() throws Exception {
...
+    assertThat(startEvent.from(), startsWith("busybox:"));
+    assertThat(startEvent.time(), notNullValue());
+
+    eventStream.close();
+  }
+
+  @Test
+  public void testEventStreamWithSinceTime() throws Exception {
+    Thread.sleep(1000); // ensure we push to the next second
+                        // so we don't get events from the last test
```
mattnworb: how could you get data from a previous test if `DefaultDockerClient sut` is re-instantiated for each test?

davidxia:
And the tests fail...

@colinmorelli Any idea why off the top of your head? I copied them directly from https://github.com/spotify/docker-client/pull/127, but I haven't dug into it yet.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
colinmorelli: @mattnworb the docker daemon doesn't change between tests. The problem here is that if the first test runs at time 1, and the second test starts running at time 1.9, they share the same unix timestamp (in seconds). Then, when the second test runs, it requests events starting from time t=1, and receives events that the daemon generated during the first test's run.

In an ideal world each test would run with an isolated daemon to avoid this problem, but I see no way to make that feasible.

mattnworb: ah, I missed that the timestamp is being passed to the daemon. thanks for the explanation :+1:

colinmorelli:
@davidxia Looks like there was a change to the way the docker API reports events. Previously, when starting a container from an image repository (using `:latest` implicitly), the events from the daemon included `:latest` appended to the image name. That appears to no longer be the case.

Local testing using client/server 1.10.2 and API version 1.22 confirms that running `docker run -it busybox` results in an event with image = `busybox` rather than `busybox:latest`.

That should be all that needs to change. Can't speak to the log test failure.

davidxia:
@colinmorelli Ah thanks

eduardolundgren:
What would be the recommended way for listening an EventStream to simulate `addListener` behavior?

```java
public void addListener(String event, Consumer consumer) {
  ...
  new Thread() {
	@Override
	public void run() {
		while (eventStream.hasNext()) {
            consumer.accept(eventStream.next().id());
		}
	}
  }.start();
}
```

PR #362, opened 2016-03-05
Add header support

sayarg:
add support for http headers

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,6 +154,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>1.10.19</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.hamcrest</groupId>
       <artifactId>hamcrest-library</artifactId>
       <version>1.3</version>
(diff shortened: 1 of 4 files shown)


mattnworb:
I think this looks fine. I have a few small suggestions but I don't see them as dealbreakers

- to be able to specify a different way of constructing the ClientBuilder, you could also just have the overloaded constructor take a `Supplier<ClientBuilder>`, so that you don't have to define a new interface just for being visible from tests. I think that would read a little easier.
- in the DefaultDockerClientBuilder, looks like a way to add a bunch of headers at once, e.g. `headers(Map)`, would be useful.

Thanks for the contribution and sorry for the slow reply on #343. This would make #343 obsolete, right?

sayarg:
tvm @mattnworb 

just closed https://github.com/spotify/docker-client/pull/343

if ok with you it would be great if you could release this change? happy to do the clean up follow ups too. There's another change i need to put regarding resource leak in the way certificates are loaded, without it you may run out of fd, i'll try to do all these in this week.

cheers.

sayarg:
thanks for merging @mattnworb 

do you know when the next release is going to be?

mattnworb:
@sayarg should be as soon as possible, the Jenkins we use internally is having issues with one of the tests repeatedly failing, so as soon as I fix that we'll have a new version on the way to maven central. Should be a day or two.

davidxia:
I just released 3.6.2

sayarg:
and events are in too! awesome thanks vm.

mattnworb:
thank you for the contribution @sayarg !

Issue #361, opened 2016-03-04
Support renaming container


julian-eggers:
Please support the rename-api
https://docs.docker.com/engine/reference/commandline/rename/

davidxia:
fixed by #363

PR #363, opened 2016-03-07
Support for container rename

johnflavin:
Addresses issue #361.

This is a first pass. I have two main caveats that I would like some feedback on before considering this final.

* The [API](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.18/) says the `rename` request can throw a `409` if the name is already assigned. Should I convert that into a named exception, à la `404 → ContainerNotFoundException`?
* I ran into a docker bug (or, perhaps, non-documented feature) where the `name` field returned by `container inspect <id>` has a leading `/`. See [#18558](https://github.com/docker/docker/issues/18558).  I created a test for my changes, but as I am testing the value of the name I had to change the comparison I'm making based on the API version. I'm running client 1.9.1 (API 1.21), so I am not able to test that the tests pass on lower APIs. (Though I guess CircleCI will make see if it passes on 1.20.) Any help testing on other versions would be appreciated.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -19,16 +19,15 @@
 
 package com.spotify.docker.client;
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.io.CharStreams;
 import com.google.common.net.HostAndPort;
-
-import com.fasterxml.jackson.core.JsonGenerator;
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.spotify.docker.client.messages.AuthConfig;
 import com.spotify.docker.client.messages.AuthRegistryConfig;
 import com.spotify.docker.client.messages.Container;
@@ -48,7 +47,6 @@ import com.spotify.docker.client.messages.NetworkCreation;
 import com.spotify.docker.client.messages.ProgressMessage;
 import com.spotify.docker.client.messages.RemovedImage;
 import com.spotify.docker.client.messages.Version;
-
 import org.apache.commons.compress.utils.IOUtils;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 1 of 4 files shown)


johnflavin:
Update: My initial commit failed the build. Looks like API 1.20 also returns the container name with a leading `/`. I modified the test accordingly.

davidxia:
@johnflavin Thanks for the PR. It'd be good to have tests for the failure conditions as well, i.e. container not found, name already assigned, and invalid name.

It might be nice to have a named exception for the 409 status code, e.g. something liek `RenameConflictException`. I don't have a strong opinion though.

johnflavin:
Added a named exception for the rename conflict and filled out all the different exceptions that could be thrown. (At least, I think I got them all).

Sadly, the `ContainerRenameConflictException` doesn't actually get thrown on a rename conflict due to a docker bug. I opened an issue ([21016](https://github.com/docker/docker/issues/21016)) with them about it. I left the `renameContainer` method as it should be, as if the request will return a 409 on a bad rename like it should, but in the test I'm also catching the real exception from the 500 status.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -731,6 +731,34 @@ public ContainerCreation commitContainer(final String containerId,
   }
 
   @Override
+  public void renameContainer(final String containerId, final String name)
+          throws DockerException, InterruptedException {
+    WebTarget resource = resource()
+            .path("containers").path(containerId).path("rename");
+
+    if (name != null) {
```
mattnworb: can the `name` actually be null? I would have thought that the [rename operation](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/) returned an error if the `name` parameter was not supplied

mattnworb:
I have a question above but otherwise looks good to me. Thanks for yet another contribution @johnflavin! :smile:

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
johnflavin: Good point. I coped that section from `createContainer` and didn't really think about it. The reason it's checked for null in `createContainer` is that trying to pass null to the regex matcher throws a `NullPointerException`, and no one wants to see that.

But what is the behavior we do want here?
* Check for `name=null` at the beginning of `renameContainer` and throw some `Exception`.
* Leave as is: check for `name=null` at the beginning of `renameContainer`, but do nothing. Let the request go through with no `name` query param and let docker return a 500 status. (This may seem crazy, but honestly, all of these methods that accept IDs and names don't check any of them for null and just pass them right on through. So this would kinda be consistent with the existing style.)
* Annotate the `name` param with `@NotNull` in the method signature. (But if we do it here, why not everywhere?)

I'm open to any of these approaches.

mattnworb: Assuming that the remote API requires a name to rename the container to, I think it is best to check for null and throw IllegalArgumentException.

johnflavin: Ok. Fixed, and added a test.

mattnworb:
:+1:

mattnworb:
FWIW I get the same behavior with `docker inspect` showing a leading slash in the name even as far back as docker version 1.6.2. I think it happens any time you create a container with a specified name.

For instance:

```bash
$ docker version
Client version: 1.6.2
Client API version: 1.18
Go version (client): go1.4.2
Git commit (client): 7c8fca2
OS/Arch (client): linux/amd64
Server version: 1.6.2
Server API version: 1.18
Go version (server): go1.4.2
Git commit (server): 7c8fca2
OS/Arch (server): linux/amd64

$ docker create --name=foo busybox:latest
b7610256b0e083e2c2f4f3e6c5a302857ab018d49b317a150917d1ae8e9122c2

$ docker inspect -f '{{.Name}}' foo
/foo
```

PR #364, opened 2016-03-08
Fix testImageLabels

davidxia:
Account for pre-existing images with the same IDs.
This happens when running the test multiple times.
See comment in the code.

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -146,8 +146,8 @@ import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasItems;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.isIn;
 import static org.hamcrest.Matchers.isEmptyOrNullString;
+import static org.hamcrest.Matchers.isIn;
 import static org.hamcrest.Matchers.lessThan;
 import static org.hamcrest.Matchers.not;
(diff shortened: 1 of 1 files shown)


davidxia:
@mattnworb @johnflavin

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -2220,6 +2220,14 @@ public void testImageLabels() throws DockerException,
     final List<Image> preexistingTestImages = sut.listImages(
             ListImagesParam.withLabel("name", "testtesttest"));
     final int numPreexistingTestImages = preexistingTestImages.size();
+    // Make a list of just the image IDs
+    final List<String> preexistingTestImageIds =
+        Lists.transform(preexistingTestImages, new Function<Image, String>() {
```
johnflavin: Could just do

    final List<String> preexistingTestImageIds = imagesToShortIds(preexistingTestImages);

if you’re ok with the short (12-hex long) id.

davidxia: Thanks. I forgot about that function.

johnflavin:
Thanks for fixing this. I wasn't aware it was a problem.

Ideally none of this `preExistingImages` stuff would be necessary. If we could build an image with attributes from a POST body instead of from a Dockerfile, it would be easy to give the test images some random label value at runtime, and then you could be (almost) certain to get no collisions on the labels. But it is what it is. `¯\_(ツ)_/¯`

mattnworb:
:+1:

davidxia:
The tests were still failing so I just removed that part of the test. see https://github.com/spotify/docker-client/commit/fe9083a20249cb010b6f71e8968b3888145a0592

PR #365, opened 2016-03-09
Support force killing in remove container

### LINKED ISSUE #308: support -force when Remove a container
I want to remove a container with "-force"，but the method "removeContainer" doesnt support this parameter.

davidxia:
fixes #308

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -609,18 +609,27 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   @Override
   public void removeContainer(final String containerId)
       throws DockerException, InterruptedException {
-    removeContainer(containerId, false);
+    removeContainer(containerId, new RemoveContainerParam[0]);
   }
 
+  @Deprecated
   @Override
   public void removeContainer(final String containerId, final boolean removeVolumes)
       throws DockerException, InterruptedException {
+    removeContainer(containerId, RemoveContainerParam.removeVolumes(removeVolumes));
+  }
+
+  @Override
+  public void removeContainer(final String containerId, final RemoveContainerParam... params)
+      throws DockerException, InterruptedException {
     try {
-      final WebTarget resource = resource()
-          .path("containers").path(containerId);
-      request(DELETE, resource, resource
-          .queryParam("v", String.valueOf(removeVolumes))
-          .request(APPLICATION_JSON_TYPE));
+      WebTarget resource = resource().path("containers").path(containerId);
+
+      for (final RemoveContainerParam param : params) {
+        resource = resource.queryParam(param.name(), param.value());
+      }
+
+      request(DELETE, resource, resource.request(APPLICATION_JSON_TYPE));
     } catch (DockerRequestException e) {
       switch (e.status()) {
         case 404:
--- a/src/main/java/com/spotify/docker/client/DockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DockerClient.java
@@ -654,8 +654,8 @@ public interface DockerClient extends Closeable {
   /**
    * Remove a docker container.
    *
-   * @param containerId The id of the container to remove.
-   * @throws DockerException      if a server error occurred (500)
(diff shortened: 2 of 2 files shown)


davidxia:
@mattnworb

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -609,18 +609,27 @@ public ContainerExit waitContainer(final String containerId)
...
       throws DockerException, InterruptedException {
-    removeContainer(containerId, false);
+    removeContainer(containerId, new RemoveContainerParam[0]);
   }
 
+  @Deprecated
   @Override
   public void removeContainer(final String containerId, final boolean removeVolumes)
       throws DockerException, InterruptedException {
+    removeContainer(containerId, RemoveContainerParam.removeVolumes());
```
mattnworb: this seems to ignore the `boolean removeVolumes` parameter to the method

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -609,18 +609,27 @@ public ContainerExit waitContainer(final String containerId)
...
     try {
-      final WebTarget resource = resource()
-          .path("containers").path(containerId);
-      request(DELETE, resource, resource
-          .queryParam("v", String.valueOf(removeVolumes))
-          .request(APPLICATION_JSON_TYPE));
+      WebTarget resource = resource().path("containers").path(containerId);
+
+      for (final RemoveContainerParam param : params) {
+        resource = resource.queryParam(param.name().toLowerCase(Locale.ROOT), param.value());
```
mattnworb: I think this pattern probably exists already elsewhere in the class, but instead of `.name().toLowerCase(locale)` why not just make sure that the enum `.name()` method returns the value to use in the query parameter? Or explicitly name the enum method `enum.queryParameterName()`

the locale bit is the part here that throws me off

mattnworb:
would be good to add a test - for example, that a `new RemoveContainerParam("foo", "bar")` has the values added as expected to the query string

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
davidxia: facepalm

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
davidxia: I copied this from an actual enum `AttachParameter`. The `toLowerCase()` makes sense for that one, but not for the `RemoveContainerParam` class.

davidxia:
@mattnworb A test like this

```java
  @Test
  public void testRemoveContainerParam() throws Exception {
    final RemoveContainerParam removeVolumes = RemoveContainerParam.removeVolumes();
    assertEquals(removeVolumes.name(), "v");
    assertEquals(removeVolumes.value(), "1");

    final RemoveContainerParam forceKill = RemoveContainerParam.forceKill();
    assertEquals(forceKill.name(), "force");
    assertEquals(forceKill.value(), "1");
  }
}
```

or something that also tests the `WebResource` is constructed properly? That might involve a larger refactoring of `DefaultDockerClient`.

mattnworb:
Yeah that is what I was thinking, but good point, I didn't realize the test wasn't constructed with a mock http client.

davidxia:
So skip a test for now and refactor later?

mattnworb:
:+1:

davidxia:
BTW, tests fail because on CircleCI there sometimes is a container with an id of 0. Haven't looked into it though.

mattnworb:
@davidxia probably related to Circle's special docker setup

johnflavin:
That’s my fault, sorry. I want to test renaming a container that didn’t exist. In my first pass, I created a new container, got its id, immediately deleted it, then tried to rename it. That test passed for me locally but failed on CircleCI when I pushed; something about the `delete` was bothering Circle. So I removed the delete and hard-coded trying to rename container `0`, thinking. Container wouldn’t exist. That seems to have been a bad idea. The create, delete, rename test would be much better, if we can get it to work.

Issue #284, opened 2015-11-16
Support for cpu-quota option


damianopezzotti:
Actually docker-client doesn't support `cpu-quota` option for container creation

davidxia:
We are open to PRs :)

damianopezzotti:
Ok :-)

lemz90:
Hi bit of a noob willing to look into this as I need support for cpuset and memory. Am I correct in thinking that these are not implemented yet either? After glancing through the code and playing with the api, I could not get it to work.

I was creating ContainerConfig like so: 
`final ContainerConfig containerConfig = ContainerConfig.builder()
            .image(imageName)
            .hostConfig(hostConfig)
            .exposedPorts(ports)
            .cmd(cmd)
            .cpuset("0")
            .memory((long)512)
            .build();`



Anyway. I believe @JsonProperty needs to be added for each of these options we want to implement. Can anyone tell me if I'm looking in the right direction.

Thank you.

davidxia:
@lemz90 I believe it's been done here https://github.com/spotify/docker-client/pull/286

tedyoung:
Because of a Docker Remote API change, issue #286 moved some of these to the HostConfig. It looks like they're still in the ContainerConfig Builder, but -- as @lemz90 noted -- they're no longer `@JsonProperty`s. Seems like they should be completely removed from the ContainerConfig.Builder inner class so as not to mislead folks.

lemz90:
@tedyoung @davidxia Wow thanks for pointing that out! I might work on a pull request to remove these options from ContainerConfig >:)

Otherwise it seems like this issue could be closed.

PR #366, opened 2016-03-09
Remove obsolete builder params from ContainerConfig

davidxia:
Remove cpu and memory options from ContainerConfig. They weren't
being used at all. The options are set using HostConfig.Builder.

related to #284


davidxia:
@mattnworb

mattnworb:
:+1:

PR #371, opened 2016-03-14
ListImages/ListContainers query param fixes

johnflavin:
(This is a small refactor. I plan to use it in a bigger feature PR, but I'd like that feature to stand on its own so I want to submit this first.)

The code in `listImages` and `listContainers` is quite duplicative. Everything to process the params in the two methods is almost identical, except that the params have different types. They all have to be URL-encoded, the `filters` params have to be serialized into JSON, and all the exceptions in that process have to be converted to `DockerException`s. It is nice to have the parts that don't depend on the specific param types extracted into their own convenience functions.

Part of this extraction happened in 7ebb4b8. This continues that job.


davidxia:
Awesome, thanks.

PR #374, opened 2016-03-15
Support cap-add and cap-drop

davidxia:


src/test/java/com/spotify/docker/client/messages/HostConfigTest.java
```diff
@@ -65,6 +66,14 @@ public void testJsonOnFailure() throws Exception {
     assertThat(hostConfig.restartPolicy(), is(HostConfig.RestartPolicy.onFailure(5)));
   }
 
+  @Test
+  public void testCapAddAndDrop() throws Exception {
```
davidxia: A bit of a dumb test, but we can't easily mock WebTarget in DefaultDockerClient right now...

mattnworb:
@davidxia I think a good strategy for testing it would be to use the `DefaultDockerClient` constructor that takes a `RSClientBuilderWrapper` and have that return a mock jersey client.

mattnworb:
I think #376 would help with this

davidxia:
@mattnworb Is this PR dependent on #376? Seems like it's not, but #376 seems good to have nonetheless.

mattnworb:
is this no longer WIP? LGTM as is.

davidxia:
Yup, it's done. I just updated the commit message.

PR #376, opened 2016-03-16
simplify unit testing of DefaultDockerClient

mattnworb:
Supplier from guava (which I believe inspired
`java.util.function.Supplier` in Java 8) already provides a clear way to
communicate that some parameter only exists to "supply" another type.

Simplify the DefaultDockerClient constructor overload that allows for
passing in a custom Jersey ClientBuilder for testing.


davidxia:
:+1:

PR #377, opened 2016-03-17
Add support for providing IpcMode in HostConfig

lresende:
Enables passing ipcMode into host configuration. The default will be empty and you can pass the string as "host" or as a reference to another container.

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/HostConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/HostConfig.java
@@ -57,6 +57,7 @@ public class HostConfig {
   @JsonProperty("CgroupParent") private String cgroupParent;
   @JsonProperty("RestartPolicy") private RestartPolicy restartPolicy;
   @JsonProperty("LogConfig") private LogConfig logConfig;
+  @JsonProperty("IpcMode") private String ipcMode;
 
   private HostConfig() {
   }
@@ -85,6 +86,7 @@ public class HostConfig {
     this.cgroupParent = builder.cgroupParent;
     this.restartPolicy = builder.restartPolicy;
     this.logConfig = builder.logConfig;
+    this.ipcMode = builder.ipcMode;
   }
 
   public List<String> binds() {
@@ -177,6 +179,10 @@ public class HostConfig {
 
   public LogConfig logConfig() {
     return logConfig;
+  }
+
+  public String ipcMode() {
+    return ipcMode;
   }
 
   @Override
@@ -273,6 +279,11 @@ public class HostConfig {
       return false;
     }
 
+    if (ipcMode != null ? !ipcMode.equals(that.ipcMode)
+            : that.ipcMode != null) {
+      return false;
+    }
+
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


lresende:
Could someone please help review this and if possible create a release after this is merged. Thanks

johnflavin:
I'm not familiar with `ipcMode`, but I see in the [Remote API docs (v1.20)](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.20) it is part of the `HostConfig` returned when [inspecting a container](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.20/#inspect-a-container). It doesn't appear in the docs as part of the `HostConfig` when, say, [creating a container](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.20/#create-a-container), but I find it plausible that one could do that.

I'll leave a more specific comment on your test case.

src/test/java/com/spotify/docker/client/messages/HostConfigTest.java
```diff
@@ -65,6 +65,15 @@ public void testJsonOnFailure() throws Exception {
     assertThat(hostConfig.restartPolicy(), is(HostConfig.RestartPolicy.onFailure(5)));
   }
 
+  @Test
+  public void testIpcModeHostConfiguration() throws Exception {
+    HostConfig hostConfig = objectMapper
+            .readValue(fixture("fixtures/hostConfig/ipcModeHost.json"),
+                    HostConfig.class);
```
johnflavin: This isn't how the `HostConfig` is created in real-world use. You should do
```java
HostConfig hostConfig = HostConfig.builder().ipcMode("whatever").build();
```

src/test/java/com/spotify/docker/client/messages/HostConfigTest.java
```diff
@@ -65,6 +65,15 @@ public void testJsonOnFailure() throws Exception {
...
+    assertThat(hostConfig.ipcMode(), is("host"));
```
johnflavin: This assertion only tests that the json deserializes properly into the object. While that is useful to know, there is a much better way to test that.

If you use this `HostConfig` as part of a `ContainerConfig`, then you use that to `createContainer`, you'll test that the `ipcMode` setting is accepted by the Docker Remote API. Then if you `inspectContainer` with the `id` of the container you just made, you can `assertThat` the `ipcMode` on the container you got back is equal to the `ipcMode` you set initially. That tests that the json (received from docker) deserializes properly, and that nothing happened to the value along the way.

lresende:
@johnflavin ipcMode was introduced in API 1.18, and I need that to fix some issues I am having related to a DB2 image I have which needs to be started with ipcMode = Host in a test environment using the docker-client API. I can pass the parameter and fix the problem in the command line, but didn't see the option in the docker-client library.

lresende:
@johnflavin Also, for testing, I went with your suggestion which seems much more aligned with what I was looking for but didn't find in the quick code scan initially. Thanks for the feedback.

lresende:
@johnflavin I would appreciate if you could help with this and push a release to maven if possible as this is blocking me on another project. 
Thank you.

johnflavin:
I have no power to do that. I'm just a user of the library like you are.

Is there a reason you couldn't package your own fork and host that yourself, at least temporarily until a new official release comes out with your changes merged?

lresende:
@davidxia I would appreciate if you could help with this and push a release to maven if possible as this is blocking me on another project.
Thank you.

mattnworb:
LGTM. I can push a release with this next week when back in the office.

lresende:
Thanks @mattnworb, I would also appreciate if you could push a release with this fix.

mattnworb:
Sorry for the delay, but this is now available in `docker-client:3.6.5`

lresende:
Thank you @mattnworb !!!

Issue #228, opened 2015-08-21
Improve/document API for binding a volume to a container


LyndonArmitage:
Given that there are `volumesFrom()` methods on the `HostConfig.Builder`, `volumes()` methods on the `ContainerConfig.Builder` and `binds()` methods on the `HostConfig.Builder`, all of which missing any JavaSoc, I spent some time trying to figure out the correct one to use in order to mount a logs directory for a tomcat container.

An example in the readme and some JavaDoc on these methods would go a long way to alleviate any confusion to others.

Also it might be worth improving the binds API by creating a simple Java object for a mounts settings, similar to a [Pair](http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/Pair.html) object or altering the binds method to take 2 parameters. So instead of doing something like this:

```java
HostConfig hostConfig = HostConfig.builder()
		.portBindings(getPortBindings(usedPort))
		.binds(logsDir.getAbsolutePath() + ":/usr/local/tomcat/logs") // binds docker volumes
		.build();
```
You'd instead do this:

```java
HostConfig hostConfig = HostConfig.builder()
		.portBindings(getPortBindings(usedPort))
		.binds(logsDir.getAbsolutePath(), "/usr/local/tomcat/logs") // binds docker volumes
		.build();
```

or this:

```java
HostConfig hostConfig = HostConfig.builder()
		.portBindings(getPortBindings(usedPort))
		.binds(new MountPair(logsDir.getAbsolutePath(), "/usr/local/tomcat/logs")) // binds docker volumes
		.build();
````

ymruan:
totally agree.

I firstly went wrong with following. 
HostConfig hostConfig = HostConfig.builder()
        .portBindings(getPortBindings(usedPort))
        .binds(logsDir.getAbsolutePath(), "/usr/local/tomcat/logs") // binds docker volumes
        .build();

grexe:
yes the lack of any documentation (besides very sparse Javadoc) also led me here (what about a project wiki?).
I'd propose to stay lightweight and follow the builder pattern like:

    bindVolume(from("/path/on/host/").to("/path/in/container")
    .bindVolune(from("/another/host/path").to("/another/container/path")

grexe:
Just noticed that in v3.2.1, binds() is just a List that allows easy adding of volume mounts (verbatim, the syntax could be hidden by the API, but it's easy enough), so that's fine with me...

johnflavin:
[...]

PR #378, opened 2016-03-20
Document creating a container with volumes. Update support for Mounts.

johnflavin:
Addresses #228.

This PR:
* Adds a section to the README about how to mount a volume in a container,
* Deprecates the `ContainerInfo.volumes()` method in favor of `ContainerInfo.mounts()`. Information about container mounts has been returned in the latter since API 1.20.
* Adds a builder-style `Bind` inner class to `HostConfig`. Instead of passing a string into `HostConfig.builder().binds("/foo:/bar")`, now you can do `HostConfig.builder().binds(Bind.from("/foo").to("/bar").build())`.
* Slightly modifies & cleans up `ContainerMount`
* Tests creating containers with volumes and inspecting them, both in the old way and the new way.

I had some more substantial changes planned for this PR, but that's because I was basing my expectations on the Remote API docs. And when it comes to creating a container with mounts vs volumes, the docs are currently wrong. See [docker/docker#21335](https://github.com/docker/docker/issues/21335) and the comments on [docker/docker#15908](https://github.com/docker/docker/issues/15908).

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/ContainerMount.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerMount.java
@@ -18,6 +18,8 @@
 package com.spotify.docker.client.messages;
 
 import com.google.common.base.MoreObjects;
+import com.google.common.base.Objects;
+
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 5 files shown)


src/main/java/com/spotify/docker/client/messages/ContainerMount.java
```diff
@@ -60,29 +61,15 @@ public boolean equals(final Object o) {
...
   }
 
   @Override
   public int hashCode() {
-    int result = source != null ? source.hashCode() : 0;
-    result = 31 * result + (destination != null ? destination.hashCode() : 0);
-    result = 31 * result + (mode != null ? mode.hashCode() : 0);
-    result = 31 * result + (rwInfo != null ? rwInfo.hashCode() : 0);
-    return result;
+    return Objects.hashCode(source, destination, mode, rw);
```
mattnworb: this can just be `java.util.Objects.hash(..)` (we should replace the usage in the rest of the codebase later too)

mattnworb:
:+1: and thanks for the docs improvement.

Looks like a rebase is necessary, I added a similar method for testing docker versions in DefaultDockerClientTest as 

```java
// in setUp
sut = builder.build();

boolean dockerApiVersionSupported(String expected) throws Exception {
  return compareVersion(dockerApiVersion, expected) >= 0;
}
```

johnflavin:
In fact, I had already rebased, but I changed the name of your function from `dockerApiVersionSupported` to `dockerApiVersionGreaterOrEqualTo`. :grin:

Since it seems to have been confusing, and wasn't really necessary, I changed it back.

mattnworb:
ah sorry about the confusion with the method name! 

Thanks again for the contribution

PR #381, opened 2016-03-21
consolidate API version checks in DefaultDockerClientTest

mattnworb:
Pull all of the API version checks and `assumeTrue`s into a method to prevent accidental errors.

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -236,6 +236,17 @@ public class DefaultDockerClientTest {
     sut.close();
   }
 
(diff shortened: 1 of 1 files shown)


johnflavin:
Request: also include a method for api version `<=` a specific version. I added two tests in #378, one that tests how the API behaves in version > 1.19, and another for <= 1.19.

mattnworb:
@johnflavin something like `requireDockerVersionBefore(String maxVersion, String reason)`? I think that is a good idea. Feel free to add it into your PR, I feel weird adding in this one if there is no code that uses it.

johnflavin:
Ok, will do. I was going to have to rebase on master anyway. You’re pulling in a lot of fixes this morning. :coffee:

PR #386, opened 2016-03-21
fix testListContainers for docker < 1.8

mattnworb:
The container status of "Created" was added in
https://github.com/docker/docker/commit/dc697b1c9f2bf77b5b78aa871e3f365e333fd253,
the test fails when run against Docker 1.6 (and presumably 1.7).

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -119,6 +119,12 @@ import static com.google.common.base.Charsets.UTF_8;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static com.spotify.docker.client.DefaultDockerClient.NO_TIMEOUT;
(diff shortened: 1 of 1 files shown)


mattnworb:
Btw to explain the flurry of PRs today - we test docker-client in CircleCI against Docker 1.8 (IIRC, that is the oldest version that is available) but for unrelated and uninteresting reasons we use Docker 1.6 internally at Spotify. So after merging PRs from contributors into the project that work fine in 1.8, we find subtle test failures when we try to build the project (in order to release to Maven Central) on our internal Jenkins. I've been trying to fix those all today to get the project back into a good shape internally.

I'd like to explore soon if it is possible to have a CI solution that can do a matrix build of docker-client against several versions of docker, to try and catch these issues at submission time, rather than release time. I think it would be a lot easier for us as maintainers but also better for contributors also (to be able to tell you that your change breaks on docker 1.7 and 1.8, for instance).

Alternatively we could run our Jenkins here at Spotify on something > 1.6, but since we are still using that version in production it is important to us to make sure that docker-client is well-tested for 1.6.

Issue #387, opened 2016-03-21
improve CI to test against multiple Docker versions


mattnworb:
We test docker-client in CircleCI against Docker 1.8 (IIRC, that is the oldest version that is available) but for unrelated and uninteresting reasons we use Docker 1.6 internally at Spotify. So after merging PRs from contributors into the project that work fine in 1.8, we find subtle test failures when we try to build the project (in order to release to Maven Central) on our internal Jenkins. 

I'd like to explore soon if it is possible to have a CI solution that can do a matrix build of docker-client against several versions of docker, to try and catch these issues at submission time, rather than release time. I think it would be a lot easier for us as maintainers but also better for contributors also (to be able to tell you that your change breaks on docker 1.7 and 1.8 but works on 1.9, for example).

Alternatively we could run our Jenkins here at Spotify on something > 1.6, but since we are still using that version in production it is important to us to make sure that docker-client is well-tested for 1.6.

mattnworb:
[Travis supports Docker](https://docs.travis-ci.com/user/docker/) now but doesn't appear that you can choose the version.

mattnworb:
One other idea would be to build PRs from contributors on our internal Jenkins, using the Github PR builder plugin (which asks admins to approve testing of each PR to make sure they are not malicious).

mattnworb:
How to install Docker 1.6 or 1.7 on Circle: https://discuss.circleci.com/t/how-use-a-different-docker-version/298/7

1.9: https://discuss.circleci.com/t/docker-1-9-1-is-available/1009

1.10 beta: https://discuss.circleci.com/t/docker-1-10-0-is-available-beta/2100

johnflavin:
Do you want someone (you, me, anyone else) to file issues for all the failures happening over on #389?

mattnworb:
@johnflavin yeah I think that is a fine approach, or a PR that fixes an issue at the same time is fine too. I am going to close this particular issue since we now have a CI solution for Github PRs that tests on > 1 docker version.

rgrunber:
It's good to start addressing this issue. I also wouldn't mind seeing support for, say 2 or 3 previous versions. Generally I don't see checks being contributed when functionality is changed from one version of the docker Remote API to another so that breaks compatibility.

I know that the Remote API has a way to specify what version you expect in response, but it seems versioning isn't really supported for all fields, nor is it clear what fields are versioned (https://github.com/docker/docker/issues/14524) .

mattnworb:
[...]

PR #388, opened 2016-03-22
setup travis-ci for multiple docker versions

mattnworb:
Sets up a matrix build that combines various Docker versions with various options for `DOCKER_HOST`. This can also be used to build against multiple JDK versions as well.

Setting up a specific Docker version is done by simply `apt-get remove` and
`apt-get install` specific versions. Some care is needed with version 1.6.x
to set up an Upstart job as the docker-engine package doesn't seem to do
this until 1.7+.

This same setup works with Docker version 1.7.x, 1.9.x, and 1.10.x, but
there seem to be some real test failures with these versions. I've kept
those versions out of this matrix for now until those failures can be
fixed.

Compared with CircleCI, this allows us to test the docker-client library
across a variety of Docker runtime/API versions, and to be able to catch
possible issues with certain versions at contribution-/CI-time (rather than
when we at Spotify try to build and release a version of the library on our
Jenkins that uses Docker 1.6 days or weeks after a contribution has been
merged).

Unlike CircleCI, there doesn't seem to be an obvious limitation on how many
builds we can launch at once; we would just possibly have to deal with
longer queue times.

I haven't investigated it much yet but by testing against a regular version
of Docker instead of the special/custom version that CircleCI sets up, we
may need to ignore less tests too.

Addresses #387.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,13 +1,8 @@
 #!/bin/bash
 set -e
 
-show_help() {
-  echo "Usage: $0 <command>"
-  echo "Commands: install_docker, dump_docker_config"
-}
-
 if [[ -z $1 ]]; then
-  show_help
+  "I need a command!"
   exit 1
 fi
 
@@ -30,6 +25,9 @@ case "$1" in
     apt-cache policy docker-engine
     sudo apt-get -q -y install docker-engine=$DOCKER_VERSION* "linux-image-extra-$(uname -r)"
 
+    # set DOCKER_OPTS to make sure docker listens on the ports we intend
+    echo 'DOCKER_OPTS="-D=true -H=unix:///var/run/docker.sock -H=tcp://127.0.0.1:2375"' | sudo tee -a /etc/default/docker
+
     if [[ "$DOCKER_VERSION" =~ ^1\.6\..* ]]; then
       # docker-engine 1.6.x packages don't seem to have the upstart job config,
       # so write it ourselves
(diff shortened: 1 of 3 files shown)


mattnworb:
Here is an example build in travis: https://travis-ci.org/mattnworb/docker-client/builds/117786353

It won't show up in this repo's "PR checks" until added as a service

mattnworb:
@spotify/helios-team

.travis.sh
```diff
@@ -0,0 +1,88 @@
...
+case "$1" in
+  install_docker)
+
+    if [[ -z $DOCKER_VERSION ]]; then
+      echo "DOCKER_VERSION needs to be set as an environment variable"
+      exit 1
+    fi
+
+    # TODO detect which docker version is already installed and skip
+    # uninstall/reinstall if it matches $DOCKER_VERSION
```
rohansingh:
```bash

if [ "$DOCKER_VERSION" = $(docker version | grep 'Server version' | cut -d' ' -f2) ]; then
  exit 0
fi
```

rohansingh: Though actually I guess you still want to set `DOCKER_OPTS` so maybe not `exit 0`.

.travis.sh
```diff
@@ -0,0 +1,88 @@
...
+
+    # stop docker service if running
+    sudo stop docker || :
```
negz: Nit, but I find '|| true' to be more readable/obvious than '|| :'.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1510,6 +1511,8 @@ public void testNoDockerCertificatesInDir() throws Exception {
 
   @Test
   public void testSsl() throws Exception {
+    assumeFalse(TRAVIS);
```
rohansingh: Just curious, what about this test doesn't work?

rohansingh:
Love the idea, +:100:

See inline comments.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
mattnworb: didn't investigate yet. it's probably worth trying to get it to work

.travis.sh
```diff
@@ -0,0 +1,88 @@
...
+# See https://github.com/docker/docker/issues/6647
+post-start script
+	DOCKER_OPTS="-D=true -H=unix:///var/run/docker.sock -H=tcp://127.0.0.1:2375"
+	while ! [ -e /var/run/docker.sock ]; do
+		initctl status $UPSTART_JOB | grep -qE "(stop|respawn)/" && exit 1
+		echo "Waiting for /var/run/docker.sock"
+		sleep 0.1
+	done
+	echo "/var/run/docker.sock is up"
+end script
```
negz: Could this go in a separate file rather than a here doc?

.travis.sh
rohansingh: But on some interpreters where `true` is not a builtin, that may require running the whole `/bin/true` process! /s

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
rohansingh: :+1: to merge the way it is and investigate later.

.travis.sh
mattnworb: newer versions have changed the `docker version` output :grinning: 

```
Server:
 Version:      1.10.2
```

mattnworb: but yea, good idea

.travis.sh
mattnworb: yea that probably would have been a lot more sensible than learning how to escape these single quotes!

davidxia:
Great idea, Matt. Long overdue. Thanks for doing this.

On Tuesday, March 22, 2016, Matt Brown <notifications@github.com> wrote:

> Merged #388 <https://github.com/spotify/docker-client/pull/388>.
>
> —
> You are receiving this because you are on a team that was mentioned.
> <https://github.com/spotify/docker-client/pull/388#event-599314424>
>


-- 
David Xia
Software Engineer
dxia@spotify.com

PR #389, opened 2016-03-22
travis: also test against docker 1.7, 1.9, 1.10

mattnworb:

PR #390, opened 2016-03-25
Fix testBuildName for docker 1.10

johnflavin:
There have been a bunch of test failures over on #389, and this fixes one.

See this stack trace ([Line 3337 of build 305.9](https://travis-ci.org/spotify/docker-client/jobs/117967153#L3337)):

    testBuildName(com.spotify.docker.client.DefaultDockerClientTest)  Time elapsed: 0.303 sec  <<< FAILURE!
    java.lang.AssertionError: 
    Expected: a string starting with "bf1f8aff684a"
         but: was "sha256:bf1f8aff684a8497e4f856d15ff0a5bbf0a531db25d5152abb1f9bae4b505a5e"
	    at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
            ...
This is because API 1.22 reports image ids with a "`sha256:`" prefix on `docker inspect`. Nothing in the library is wrong. The test just needs to be updated to assert the correct ID.

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -157,6 +157,7 @@ import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.isEmptyOrNullString;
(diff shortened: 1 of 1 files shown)


johnflavin:
Added a fix for another test that failed for the same reason: `testImageLabels`.

johnflavin:
I do have a philosophical/style question. 🤔

For these two fixes, I just copied the entire test, added one line at the top of each to `assume` the proper version, and changed whatever line(s) had caused it to fail. But I don't think that's very sustainable. Over time the `DefaultDockerClientTest` will be full of slightly different duplicated tests to cover minor API differences.

Is there some better way to fix these tests? I can't think of one, which is why I did things the way I did. But maybe someone else can come up with something better.

johnflavin:
Failures are (I think) caused by #391.

mattnworb:
> Is there some better way to fix these tests? 

Instead of having one `@Test` for when the runtime version is < X and another `@Test` for when it is >= X, I would instead aim for having one test that just branches based on the version. For example

```java
final String expectedId = apiVersion > X ? "sha256:" + imageId : imageId;
assertThat(info.id(),  startsWith(expectedId));
```

mattnworb:
Also thanks for taking a look at these!

johnflavin:
>> Is there some better way to fix these tests?

> Instead of having one @Test for when the runtime version is < X and another @Test for when it is >= X, I would instead aim for having one test that just branches based on the version. 

You’re probably right. For more complicated test variations, it could be appropriate to have two tests. But this issue is so small, I should just keep one test and change the expected ID string based on the API version.

I think I will close this PR and open a new one after I change some things around. I shouldn't be pulling this into master, I should pull it into your `travis-more-docker-versions` branch.

PR #396, opened 2016-03-28
Replace manual calc of hash code with Objects.hash()

davidxia:


mattnworb:
:-1: for use of the Guava method. See the javadoc in the class:

```java
* <p><b>Note for Java 7 and later:</b> This method should be treated as
   * deprecated; use {@link java.util.Objects#hash} instead.
```

davidxia:
derp

davidxia:
fixed now

mattnworb:
:+1:

PR #398, opened 2016-03-29
Do not add filters if there are none

johnflavin:
I noticed a lot of lines in my docker logs like `"GET /containers/json?filters=%7B%7D"`. That is from the `DefaultDockerClientTest#tearDown` method checking if there are running containers to stop. Looks like every request to `listContainers` and `listImages` includes `filters` as a query param whether there are any filters to include or not.

It isn't a bug. It doesn't cause anything to fail. It's just... untidy.

So this PR tidies it up.


davidxia:
:+1:

PR #399, opened 2016-03-29
Moar test fixes for docker 1.10.3

davidxia:
Docker's repository naming became more strict.
See https://docs.docker.com/registry/spec/api/#overview


mattnworb:
:+1:

PR #400, opened 2016-03-29
Yet moar travis fixes

davidxia:


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -284,8 +284,18 @@ private void requireDockerApiVersionLessThan(final String required, final String
...
+    return compareVersion(dockerApiVersion, expected) != 0;
+  }
+
+  private void requireDockerApiVersionNot(final String version, final String msg) {
+    assumeTrue(msg, dockerApiVersionNot(version));
+  }
+
   @Test
   public void testSearchImage() throws Exception {
+    requireDockerApiVersionNot("1.19", "Docker 1.7.x sends the wrong Content-Type header for "
```
davidxia: @mattnworb Is there a way to handle the incorrect `Content-Type text/plain` header? I couldn't find it by googling jersey stuff.

mattnworb: I think you would have to add something to the jersey client config that told it to use Jackson/ObjectMapper when the content-type is text/plain, which feels wrong and likely to cause other bugs. It might be possible to add a filter of some sort that modifies the response header for this one URI before Jersey tries to read it based on the content-type, but I am not sure about that either. It may be best to leave as-is unless there is a strong desire for someone to use this method with 1.7.

davidxia: I'll leave it as is.

On Wednesday, March 30, 2016, Matt Brown <notifications@github.com> wrote:

> In src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
> <https://github.com/spotify/docker-client/pull/400#discussion_r57903985>:
>
> >    @Test
> >    public void testSearchImage() throws Exception {
> > +    requireDockerApiVersionNot("1.19", "Docker 1.7.x sends the wrong Content-Type header for "
>
> I think you would have to add something to the jersey client config that
> told it to use Jackson/ObjectMapper when the content-type is text/plain,
> which feels wrong and likely to cause other bugs. It might be possible to
> add a filter of some sort that modifies the response header for this one
> URI before Jersey tries to read it based on the content-type, but I am not
> sure about that either. It may be best to leave as-is unless there is a
> strong desire for someone to use this method with 1.7.
>
> —
> <https://github.com/spotify/docker-client/pull/400/files/3363249be0565530eb37361e333ca733b15f054f#r57903985>
>


-- 
David Xia
Software Engineer
dxia@spotify.com

davidxia:
@mattnworb @johnflavin Green!

mattnworb:
👍🏻

PR #401, opened 2016-03-29
Refactor exception handling

johnflavin:
The primary objective of this PR is to explicitly throw every `4xx` error status received from docker as a named exception.

Other changes
* Create several new exception types.
* Move exceptions into their own package (because the number was getting unwieldy).
* Add every thrown exception to the `DockerClient` javadoc.

There is one unfortunate bit about the last bullet. To include the exceptions in the javadoc, I either had to import them or use the full class path. If I imported them, I had to also add them to the methods' `throws` lists or the maven checkstyle plugin would throw an error that I had unused imports; but on the `throws` lists they showed up as unused because they all descend from `DockerException` which is also on every `throws` list. On the other hand, if I didn't import them, I had to use the full class path in the javadoc, and that's pretty ugly. I went with the latter. But I am open to doing it the other way.


johnflavin:
Build failure is unrelated to code. Travis VM [rebooted for seemingly no reason](https://travis-ci.org/spotify/docker-client/jobs/119329650#L3155) and the rest of the tests failed.

johnflavin:
Any thoughts on this? It really is mostly moving classes around, adding some 406 and 409 handling, and javadoc. It isn't all _that_ big. :grin:

priyanshus1:
Good to hear

On Tuesday 5 April 2016, Flavin <notifications@github.com> wrote:

> Any thoughts on this? It really is mostly moving classes around, adding
> some 406 and 409 handling, and javadoc. It isn't all *that* big. [image:
> :grin:]
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/spotify/docker-client/pull/401#issuecomment-205563868>
>


-- 

*Regards    Priyanshu Shukla                            *

*Student       *

*Department of Computer Engineering       Pune Institute of Computer
Technology   *


*India                                 *
PS: To reach please mail at priyanshus1@gmail.com or call +91-90280-80386

mattnworb:
Looks good to me! Sorry for the delay in reviewing. @davidxia what do you think?

I think the point about checkstyle and full classnames in the javadoc is fine, in the end it's the same thing AFAIK, and any ugliness in the javadoc source is outweighed by having more-complete documentation in regards to which exceptions are thrown when.

One request though: I think we should bump the version number up to 3.7 or even 4.0 since the change in exception packages will cause compiler errors in projects using this library.

johnflavin:
I could bump the version. That's no problem. 

Or, if you prefer, I could just not move the exceptions to their own package. I did it for the sake of tidiness, but if you think that's not a good enough reason to break downstream compatibility, I could put them all back where they were.

davidxia:
LGTM. Agree with @mattnworb on full classnames in javadoc. I don't mind bumping the version.

mattnworb:
I think the classes are fine in the new package - there are definitely quite a lot of them

johnflavin:
Ping @davidxia @mattnworb. Anything else I need to do before this can get merged?

davidxia:
Sorry for the delay. I don't think so. @mattnworb Are we all set?

mattnworb:
thanks @johnflavin :smile:

Issue #403, opened 2016-03-30
Does networks translate to NetworkStats fail?


kingsmiler:
## Description

I use DEBUG level for logging, and I run docker.stats("id"), then I can see the networks information from console

`"networks":{"eth0":{"rx_bytes":275523,"rx_packets":2364,"rx_errors":0,"rx_dropped":0,"tx_bytes":407693,"tx_packets":6099,"tx_errors":0,"tx_dropped":0}}`

but I cannot find out from containerStats.toString(), it is null "network=null".

Is there something wrong?

johnflavin:
This is a feature that docker added to their API in version 1.9 (API 1.21), but `docker-client` hasn't yet been updated to properly receive that message.

Prior to v1.9, the container stats would only return information about a single network interface—or maybe it was an aggregate of all the network interfaces; I'm not sure—as `"network"`. So you'd get something like this:
```
"network":{"rx_dropped":0,"rx_bytes":648,"rx_errors":0,"tx_packets":8,"tx_dropped":0,"rx_packets":8,"tx_errors":0,"tx_bytes":648}
``` 
See the [v1.20 Remote API docs](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.20/#get-container-stats-based-on-resource-usage). 

In [API v1.21](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#get-container-stats-based-on-resource-usage) (docker v1.9) they changed the container stats to break out the stats for each network separately as `"networks"`. So now you'd get something like this:
```
"networks":{"eth0":{"rx_bytes":5338,"rx_dropped":0,"rx_errors":0,"rx_packets":36,"tx_bytes":648,"tx_dropped":0,"tx_errors":0,"tx_packets":8},"eth5":{"rx_bytes":4641,"rx_dropped":0,"rx_errors":0,"rx_packets":26,"tx_bytes":690,"tx_dropped":0,"tx_errors":0,"tx_packets":9}}
```

`docker-client` needs to be updated to support that. But it hasn't been done yet. If you wanted to submit a PR for this, it wouldn't take that much work but would be greatly appreciated.

johnflavin:
Actually, I'll just do a PR for this myself. It won't be much work, and I think it will fix a test failure we've been seeing when building against versions 1.9 and 1.10.

davidxia:
Thanks for noticing. I just added a commit for it here
https://github.com/spotify/docker-client/pull/400 under "fix testStats()".

[...]

Issue #402, opened 2016-03-30
How can I get blkio_stats using docker-client?


kingsmiler:
There is no blkio stats in ContainerStats,  is there some way to get it?

johnflavin:
I see in the [API docs](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#get-container-stats-based-on-resource-usage) that `blkio` is a thing that is returned, and I see that the `ContainerStats` class in `docker-client` does not read it. So short answer is "no, you can't get it". But, like my answer to #403, you could submit a PR (or wait for someone else to do so) to fix this.

kingsmiler:
Thanks :)

PR #404, opened 2016-03-30
Update ContainerStats

johnflavin:
Addresses #402: Adds "blkio_stats" message as BlockIoStats class
Addresses #403: Adds "networks" message as ImmutableMap<String, NetworkStats>

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/BlockIoStats.java
+++ b/src/main/java/com/spotify/docker/client/messages/BlockIoStats.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.docker.client.messages;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableList;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+import java.util.Objects;
+
+import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
+import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.NONE;
+
+@JsonAutoDetect(fieldVisibility = ANY, getterVisibility = NONE, setterVisibility = NONE)
+public class BlockIoStats {
+
+  @JsonProperty("io_service_bytes_recursive") private ImmutableList<Object> ioServiceBytesRecursive;
+  @JsonProperty("io_serviced_recursive") private ImmutableList<Object> ioServicedRecursive;
+  @JsonProperty("io_queue_recursive") private ImmutableList<Object> ioQueueRecursive;
+  @JsonProperty("io_service_time_recursive") private ImmutableList<Object> ioServiceTimeRecursive;
+  @JsonProperty("io_wait_time_recursive") private ImmutableList<Object> ioWaitTimeRecursive;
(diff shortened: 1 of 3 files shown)


davidxia:
Oh, thanks for fixing my commit and making it an ImmutableMap.

johnflavin:
I thought it would be good to cherry-pick your commit at the base of this branch, so if you rebase on this later git can hopefully figure out it doesn't need to duplicate that.

johnflavin:
What's up with Travis? [The build for this PR](https://travis-ci.org/spotify/docker-client/builds/119809689) is just sitting around queued, as is [the build for the big testing update](https://travis-ci.org/spotify/docker-client/builds/119805138) merge into master.

davidxia:
I had the same question the other day. @mattnworb told me he thinks it's
queued behind other travis builds for other arbitrary projects.

On Thu, Mar 31, 2016 at 10:40 AM, Flavin <notifications@github.com> wrote:

> What's up with Travis? The build for this PR
> <https://travis-ci.org/spotify/docker-client/builds/119809689> is just
> sitting around queued, as is the build for the big testing update
> <https://travis-ci.org/spotify/docker-client/builds/119805138> merge into
> master.
>
> —
> You are receiving this because you commented.
> <https://github.com/spotify/docker-client/pull/404#issuecomment-203969188>
>



-- 
David Xia
Software Engineer
dxia@spotify.com

mattnworb:
the fact that we launch so many builds in the matrix probably doesn't help either. Perhaps we don't need to test with the different DOCKER_HOST values any more and could remove 50% of them by removing that.

davidxia:
Yea, is there a reason to test both unix and tcp?

On Thu, Mar 31, 2016 at 4:21 PM, Matt Brown <notifications@github.com>
wrote:

> the fact that we launch so many builds in the matrix probably doesn't help
> either. Perhaps we don't need to test with the different DOCKER_HOST values
> any more and could remove 50% of them by removing that.
>
> —
> You are receiving this because you commented.
> <https://github.com/spotify/docker-client/pull/404#issuecomment-204108283>
>



-- 
David Xia
Software Engineer
dxia@spotify.com

johnflavin:
This PR is good to go, ready for review. 

I'll preemptively point out something that I don't like but had to struggle with to get right: all the properties in `BlockIoStats` are `ImmutableList<Object>`. I couldn't figure out a better, more specific, type for them. The docker documentation for these properties is sparse. In my local testing, I could never get `stats` to return anything for any of those properties other than `null` or `[]`. I initially thought they'd be `String`s, but Jackson complained. Then I thought they would be `List<String>`s, but Jackson complained again. So I don't know what to expect there.

johnflavin:
Another thought about Travis: when you guys are working in branches on spotify/docker-client, and you are keeping open a PR to merge that branch, both the branch and the PR trigger full builds every time you push a new commit. So that doubles the CPU time spent on each commit, along with doubling from testing both `unix://` and `tcp://`.

And what's more, Travis doesn't stop or cancel a build if a newer commit is pushed that obsoletes a previous commit. So sometimes builds are queued, taking a spot in line, but no one even wants them or cares about the result.

mattnworb:
@johnflavin good point, I was curious about the separate settings for push and pr. Do you think it would make more sense to just have the build on `pr`, assuming that will trigger a new build if a PR is open and a commit is added to it? I was a little confused when setting that up so any experience/advice you have on Travis would be appreciated! :smile:

davidxia:
I was confused by why there are two builds as well, ie push vs pr. The push build occurs for each new commit on that branch. The PR build isn't the same. It tests that the result of the merge commit from the PR onto master will pass.

So it seems good to have both.

mattnworb:
I updated the Travis config to turn off "build on push" to see if that helps reduce the wait time. I don't think that we are often pushing branches to this repo where we don't already have a PR open. And if you really want to test a branch outside of a PR, you could do so in a fork that is set up on travis.

davidxia:
Ah so "build pushes" build all pushes to all branches? We can't limit it to only master?

mattnworb:
Could probably do that by examining some of the `TRAVIS_` [environment variables](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables) in the build script and exiting if it's not a PR and not master, but I'm not sure if that is very useful then if the PR build tests the merge commit anyway

mattnworb:
also sorry @johnflavin for hijacking the comments in this PR :smile:

mattnworb:
:+1: to the actual change in this PR. I think `List<Object>` is fine if the return value from the API is so ambiguous/poorly documented. Could also do something even more abstract like `List<JsonNode>` or some other Jackson type; I think both are fine.

johnflavin:
Anything else needed to merge this one, @davidxia & @mattnworb?

PR #407, opened 2016-03-31
checkstyle: Ensure local variables are final

davidxia:


davidxia:
@mattnworb @gimaker Same as what we did for helios.

mattnworb:
sure why not :ok_hand:

PR #408, opened 2016-03-31
revert HostConfig.Builder.binds() behavior that appended to list

mattnworb:
Partially revert behavior introduced in 1f4d0c7c where the
`HostConfig.Builder.binds(..)` method appends to the existing list
instead of replaces it.

We found a bunch of internal Spotify code that assumed that this method
always did a replacement, and we found it easier to revert the change
here in docker-client (while adding new methods for appending to the
list).

### DIFF
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -439,7 +439,7 @@ try (OutputStream imageOutput = new BufferedOutputStream(new FileOutputStream(im
   }
 }
 ```
- 
+
 
 ### Get a tarball containing all images.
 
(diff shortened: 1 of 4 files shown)


mattnworb:
@davidxia

davidxia:
:+1:

mattnworb:
@johnflavin just a heads up that we had to revert a small part of something you changed recently, we found it easier to add new methods with the different append (instead of replace) behavior rather than fix all existing use of docker-client that assumes that `.binds(..)` overwrites the value.

johnflavin:
:+1: Sorry about that :grimacing:

mattnworb:
no worries. If we had code that relied on a certain behavior we should have also had tests or documentation that ensured that the behavior was more clearly expressed

mattnworb:
I missed a place in the tests where `.bind()` is used twice in a row, now fixed and the pr build is green

PR #409, opened 2016-03-31
Fix PushPullIT errors

davidxia:


src/test/java/com/spotify/docker/it/PushPullIT.java
```diff
@@ -138,7 +138,7 @@ public void testPushImageToPrivateAuthedRegistryWithoutAuth() throws Exception {
     client.build(Paths.get(dockerDirectory), LOCAL_IMAGE);
 
     exception.expect(ImagePushFailedException.class);
-    exception.expectMessage("no basic auth credentials");
+    exception.expectMessage("Authentication is required");
```
davidxia: The registry seems to have been updated so this string changed: https://hub.docker.com/r/library/registry/tags/. If it changes again, I'll look into fixing the version we pull.

davidxia:
@mattnworb @negz

mattnworb:
Does this fix all the integration tests? If so might as well change the travis script to `verify` instead of `mvn test`

davidxia:
I'm doing some more local testing just to make sure, but yes, that's the
goal.

On Fri, Apr 1, 2016 at 9:48 AM, Matt Brown <notifications@github.com> wrote:

> Does this fix all the integration tests? If so might as well change the
> travis script to verify instead of mvn test
>
> —
> <https://github.com/spotify/docker-client/pull/409#issuecomment-204402221>
>



-- 
David Xia
Software Engineer
dxia@spotify.com

davidxia:
This should be good to go now.

PR #410, opened 2016-04-01
remove DOCKER_HOST from matrix builds

mattnworb:
Don't bother testing with this variation if we think it is not valuable,
as it doubles the number of builds in the matrix.


davidxia:
:+1:

johnflavin:
Could this get merged? Would make testing and opening new PRs quicker.

PR #414, opened 2016-04-15
Standardize equals and hashCode methods

johnflavin:
Continues the work of #396.

* Standardize style of `equals` and `hashCode` methods
  * Replace manual hash calculation with `java.util.Objects.hash`
  * Replace manual `null` and equality checking in `equals` with `java.util.Objects.equals`.
  * Local variable names (`Object o`, `that`) in `equals`
  * Indentation and spacing
  * Method ordering: `equals`, then `hashCode`, then `toString`
* Assorted other changes
  * Replace manual String concatenation in `Event.toString()` with Guava `MoreObjects.toStringHelper()`
  * Fix indentation throughout `Device`, not just in `equals` and `hashCode`
  * Move inner class to the bottom of `Network` instead of top
  * Fix potential bug in `Network`, in which two properties had been missed in `equals` and `hashCode`

### DIFF
--- a/src/main/java/com/spotify/docker/client/DockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DockerClient.java
@@ -571,11 +571,8 @@ public interface DockerClient extends Closeable {
 
       final BuildParam that = (BuildParam) o;
 
-      if (name != null ? !name.equals(that.name) : that.name != null) {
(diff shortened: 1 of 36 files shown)


mattnworb:
👍  thanks!

Issue #412, opened 2016-04-11
Test failure: testEventStream


johnflavin:
## Description
Had a seemingly random test error occur in `DefaultDockerClientTest.testEventStream` (see build [420.10](https://travis-ci.org/spotify/docker-client/jobs/122327781)). I want to note it here so I (or someone else) can remember to fix it later.

The code that produced the error is
```
Event startEvent = eventStream.next();
if (!dockerApiVersionLessThan("1.22")) {
  // For some reason, version 1.22 has an extra null Event. So we read the next one.
  startEvent = eventStream.next();
}
```
It appears that, occasionally, the first event is not null. Or, at least, there is only one available event in the stream, so calling `eventStream.next()` throws an error. I've never seen that error before now and I haven't seen it again (even with [another build of the exact same code](https://travis-ci.org/spotify/docker-client/builds/122339189)), so I have no idea how to reproduce it.

The quick way to fix this would be to amend the conditional to say `if (!dockerApiVersionLessThan("1.22") && startEvent != null)` or something. The longer but better solution is to try to dig into why the event stream is returning a null event in the first place, and open an issue upstream at docker/docker.

## Software:

- docker version: `1.10.3`
- docker-client version: `3.6.9-SNAPSHOT`, being updated to `4.0.0-SNAPSHOT` in #401.

## Full backtrace

```text
testEventStream(com.spotify.docker.client.DefaultDockerClientTest)  Time elapsed: 0.813 sec  <<< ERROR!
java.util.NoSuchElementException
	at com.google.common.collect.AbstractIterator.next(AbstractIterator.java:154)
	at com.spotify.docker.client.DefaultDockerClientTest.testEventStream(DefaultDockerClientTest.java:1409)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
[...]

PR #415, opened 2016-04-16
testEventStream: Only read the next event if it exists

johnflavin:
I saw [another test failure](https://travis-ci.org/johnflavin/docker-client/jobs/123603775) like the one I noted in #412. See that issue for details.

I am not certain this fix will prevent failures in `testEventStream`, as I still have no way to reproduce this failure. But I think it's worth a shot.


mattnworb:
👍

PR #417, opened 2016-04-19
Added assessor for checkDuplicate in NetworkConfig

leonverschuren:
The checkDuplicate field was the only field that didn't have an assessor method.


davidxia:
👍

PR #420, opened 2016-04-23
Fix date parsing to work with fractional seconds with less than nine digits

### LINKED ISSUE #416: Docker date time parsing (e.g. in inspect command) is not working sometimes
## Description

Docker date time parsing (e.g. in inspect command) is not working sometimes.

## How to reproduce

This happens from time to time, but not found a pattern yet. 

## What do you expect

docker inspect command response ContainerInfo should include correct datetime for startAt, finishAt, createAt timestamps parsed. 

## What happened instead

Parsed/deserialized datetime timestamp for ContainerInfo pojo is not correct somtimes. 

## Software:

- docker version: 1.9.1
- docker-client version: 3.6.8

## Full backtrace

Sample output from our environment of one container, from it you can see the docker date time format is not always the pattern expected by com.spotify.docker.client.DockerDateFormat in following code:
```
@Override
  public Date parse(String source) throws ParseException {
    // If the date has nanosecond precision (e.g. 2014-10-17T21:22:56.949763914Z), remove the last
    // six digits so we can create a Date object, which only support milliseconds.
    if (source.matches(".+\\.\\d{9}Z$")) {
      source = source.replaceAll("\\d{6}Z$", "Z");
    }

    return super.parse(source);
  }
```

Output in our environment (`docker inspect 93151d9254cb1f79c577c850a12598027e4568fecc6343327029cdf556463f5b`):

see comment next...

plamentotev:
Fix date parsing to work with fractional seconds with less than nine digits

Docker API returns dates in rfc3339 format with nanoseconds precision but the fractional seconds part is not always nine digits long.

This pull requests changes `DockerDateFormat` so it trims the seconds fraction part to at most three digits.

My understanding of seconds fraction is that they are fractions - this means that any trailing zeros could be omitted(I guess that's why Docker sometimes returns less than nine digits). So `2014-10-17T21:22:56.9Z` should mean 59 seconds and 900 milliseconds but `com.fasterxml.jackson.databind.util.StdDateFormat` converts it to 59 seconds and 9 milliseconds. I think this is wrong but we can live with it.

Fixes #416

### DIFF
--- a/src/main/java/com/spotify/docker/client/DockerDateFormat.java
+++ b/src/main/java/com/spotify/docker/client/DockerDateFormat.java
@@ -41,9 +41,11 @@ public class DockerDateFormat extends StdDateFormat {
   @Override
   public Date parse(String source) throws ParseException {
     // If the date has nanosecond precision (e.g. 2014-10-17T21:22:56.949763914Z), remove the last
-    // six digits so we can create a Date object, which only support milliseconds.
-    if (source.matches(".+\\.\\d{9}Z$")) {
-      source = source.replaceAll("\\d{6}Z$", "Z");
+    // digits so we can create a Date object, which only support milliseconds.
+    // Docker doesn't always return nine digits for the fractional seconds part,
+    // so we need to be more flexible when trimming to milliseconds.
+    if (source.matches(".+\\.\\d{4,9}Z$")) {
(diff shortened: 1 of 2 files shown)


mattnworb:
👍

mattnworb:
thanks for the fix @plamentotev !

PR #422, opened 2016-04-28
Added NetworkID field to AttachedNetwork

leonverschuren:
The field "NetworkID" was missing from AttachedNetwork. This field was added in [Docker API 1.22](https://docs.docker.com/engine/reference/api/docker_remote_api/#v1-22-api-changes)

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -2547,6 +2547,7 @@ public class DefaultDockerClientTest {
     final AttachedNetwork attachedNetwork =
         containerInfo.networkSettings().networks().get(networkName);
     assertThat(attachedNetwork, is(notNullValue()));
+    assertThat(attachedNetwork.networkId(), is(notNullValue()));
     assertThat(attachedNetwork.endpointId(), is(notNullValue()));
     assertThat(attachedNetwork.gateway(), is(notNullValue()));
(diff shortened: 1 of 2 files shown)


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -2547,6 +2547,7 @@ public void testNetworksConnectContainer() throws Exception {
     final AttachedNetwork attachedNetwork =
         containerInfo.networkSettings().networks().get(networkName);
     assertThat(attachedNetwork, is(notNullValue()));
+    assertThat(attachedNetwork.networkId(), is(notNullValue()));
```
mattnworb: there is a compilation error here because there is no method `networkId()`

mattnworb:
@leonverschuren thanks for the contribution! Looks like Travis is failing because of the compilation error above.

leonverschuren:
One job is still failing because it uses an older Docker API version. Should I bump `requireDockerApiVersionAtLeast` to 1.22?

johnflavin:
There are helper methods in `DefautDockerClentTest` to only run things (whole tests or just individual assertions) for certain API versions. Wrap your assertion in an `if` and use... Let me remember... I think the name of the method you should use is `apiVersionGreaterThan("1.22")`.

Look for some examples. This and its converse are used a fair amount throughout the test class.

davidxia:
@leonverschuren, thanks!

PR #425, opened 2016-04-29
Added networks and mounts to container list (API 1.22)

leonverschuren:
Added NetworkSettings and Mounts to the Container class.
NetworkSettings is added in Docker [API 1.22](https://docs.docker.com/engine/reference/api/docker_remote_api/#v1-22-api-changes):
`GET /containers/json now returns the list of networks of containers.`

Mounts is added in Docker [API 1.23](https://docs.docker.com/engine/reference/api/docker_remote_api/#v1-23-api-changes):
`GET /containers/json returns the mount points for the container.`

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/Container.java
+++ b/src/main/java/com/spotify/docker/client/messages/Container.java
@@ -17,12 +17,11 @@
 
 package com.spotify.docker.client.messages;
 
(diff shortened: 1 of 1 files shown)


davidxia:
@leonverschuren Thanks. I added tests for them here https://github.com/spotify/docker-client/pull/435.

PR #426, opened 2016-04-29
Add (documented) info fields

johnflavin:
The `/info` API has seen a lot of changes in recent versions, which I've attempted to capture.

# Changes
* Add many new properties to `Info` class.
* Alphabetize properties in `Info`. There were getting to be too many properties to easily compare to the API docs (in which the example info object is alphabetized).
* Add `Plugin`, `RegistryConfig`, and `IndexConfig` classes to represent some `Info` fields.
* Add tests for all `Info` fields, with fields added in each successive new API version in a new `if` block.

# Caveats
* In my testing I found some undocumented fields returned by `/info`, and I chose not to include them here. I may open an upstream issue about it.
* Most of the test assertions I added don't actually test anything, they just compare to `Matchers.anything()`. I don't have a good way to predict beforehand what most of the fields will be, or even whether they will be null or blank.

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/Info.java
+++ b/src/main/java/com/spotify/docker/client/messages/Info.java
@@ -18,45 +18,105 @@
 package com.spotify.docker.client.messages;
 
 import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


johnflavin:
@mattnworb What's wrong with the codecov reports? I guess they changed how their plugin works, and something is breaking right now.

mattnworb:
@johnflavin not sure. The link in the travis output loads fine: https://codecov.io/github/spotify/docker-client/commit/ea77379f3d1f133b095317c7dc357ddf85c6ebea

johnflavin:
Anyone up for a review?
👉 Poke @mattnworb @davidxia

mattnworb:
LGTM. Sorry for delay, was away for a few weeks.

Issue #423, opened 2016-04-28
Add Status field in ContainerState class to get explicit status of a container


quickwind:
## Description

Currently the ContainerState from ContainerInfo (response of docker inspect) doesn't have a field mapping to status of state json sub-object in following examples:

```
{
        "Status": "created",
        "Running": false,
        "Paused": false,
        "Restarting": false,
        "OOMKilled": false,
        "Dead": false,
        "Pid": 0,
        "ExitCode": 0,
        "Error": "",
        "StartedAt": "0001-01-01T00:00:00Z",
        "FinishedAt": "0001-01-01T00:00:00Z"
    }
{
        "Status": "exited",
        "Running": false,
        "Paused": false,
        "Restarting": false,
        "OOMKilled": false,
        "Dead": false,
        "Pid": 0,
        "ExitCode": 0,
        "Error": "",
        "StartedAt": "2016-04-14T06:49:59.089717628Z",
        "FinishedAt": "2016-04-14T06:50:20.24937831Z"
    }
```

The status should be mapped to get explicit container status, rather than inferring from the rest of the fields...

## How to reproduce

Always the behavior, missing feature. 

## What do you expect

Container status is an explicit field mapped in the ContainerState class. 

## What happened instead

NA

## Software:

- docker version: 1.9.1
- docker-client version: 4.0.2

## Full backtrace

NA

quickwind:
Anyone any comment on this?

johnflavin:
This should be fairly easy to add in a PR. Do you want to do that? If not, I could get to it sometime soon.

quickwind:
@johnflavin PR proposed, please review.

quickwind:
@johnflavin BTW, we are waiting for version 4.0.5 available on maven central repo, it is not there yet, what is the frequency to push to maven central repo? 

Also where I can find the release notes/changelog for a version? In github releases page, there is not much info I can get. 

Thanks.

johnflavin:
Good questions. I don’t know how often releases are made, or when they get pushed to maven central. As far as I know, there are no release notes, just the history contained in the commits, issues, and PRs.

davidxia:
I'll release now.

davidxia:
@quickwind Released as [4.0.6](http://search.maven.org/#artifactdetails%7Ccom.spotify%7Cdocker-client%7C4.0.6%7Cjar). Thanks for your PR.

quickwind:
[...]

PR #430, opened 2016-05-04
Add Status field in ContainerState class to get explicit status

quickwind:
Add Status field in ContainerState class to get explicit status of a container #423


johnflavin:
LGTM. I think you could squash those commits into one, though.

@mattnworb?

mattnworb:
👍  

thanks @quickwind! I'll squash this when merging via the Github option, but like @johnflavin mentioned, it would be helpful to do a squash on your side to condense the commits down before creating a PR (as there is less to review then).

PR #433, opened 2016-05-17
add a note on Docker for Mac

mattnworb:
the DOCKER_HOST value might not be obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,11 @@ If you're looking for how to build and develop it, keep reading.
 docker-client should be buildable on any platform with Docker, JDK7+, and a recent version of
 Maven 3.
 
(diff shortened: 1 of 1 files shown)


mattnworb:
@spotify/helios-team

johnflavin:
Might it be better to change `DefaultDockerClient.defaultEndpoint()` so that it will use `unix:///var/run/docker.sock` for both linux and OS X? Then users with Docker for Mac would not have to set `DOCKER_HOST` at all.

mattnworb:
@johnflavin fair point. I also wanted to document what value to use for people using older docker-client versions as well, prior to any logic that might get added.

davidxia:
👍

PR #434, opened 2016-05-17
Add default endpoint used by Docker for Mac

johnflavin:
I've verified this works with Docker for Mac. No `DOCKER_HOST` required.


mattnworb:
@johnflavin thanks. It looks like Docker for Mac also sets up a symlink at `/var/run/docker.sock -> /var/tmp/docker.sock`, so maybe there is no need to add a new constant?

johnflavin:
Yeah, good point. I don't know if the new constant is necessary. It's a style thing, really. It's your code (in some sense) so I leave the decision to you.

The change to the conditional is necessary, though. Before, it would only use the socket by default for Linux. Now, that should also be the default on OS X.

davidxia:
I opened https://github.com/spotify/docker-client/pull/436 to use the symlink and the correct socket.

Issue #440, opened 2016-05-19
DockerHost.fromEnv() not updated for Docker for Mac


johnflavin:
## Description

The method `DefaultDockerClient.fromEnv()` was updated to use the unix socket on Mac in #436. However, an almost identical method exists in `DockerHost.fromEnv()` and this was not updated in #436. This causes test `DockerHostTest.testFromEnv()` to fail when using Docker for Mac, connected via the unix socket with no environment variables.

## Full backtrace

```text
testFromEnv(com.spotify.docker.client.DockerHostTest)  Time elapsed: 0.001 sec  <<< FAILURE!
java.lang.AssertionError:
Expected: "unix:///var/run/docker.sock"
     but: was "localhost:2375"
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
	at com.spotify.docker.client.DockerHostTest.testFromEnv(DockerHostTest.java:106)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
[...]

PR #441, opened 2016-05-19
Move default Docker constants + methods into DockerHost

davidxia:
Default Docker constants and methods for constructing them were
duplicated in `DockerHost` and `DefaultDockerClient`. Move them all into
`DockerHost` and add tests.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -36,6 +36,7 @@ import com.spotify.docker.client.exceptions.NotFoundException;
 import com.spotify.docker.client.exceptions.PermissionException;
 import com.spotify.docker.client.messages.AuthConfig;
 import com.spotify.docker.client.messages.AuthRegistryConfig;
+
 import com.google.common.base.Strings;
 import com.google.common.io.CharStreams;
 
@@ -131,8 +132,6 @@ import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.Maps.newHashMap;
 import static com.spotify.docker.client.ObjectMapperProvider.objectMapper;
 import static com.spotify.docker.client.VersionCompare.compareVersion;
-import static java.lang.System.getProperty;
-import static java.lang.System.getenv;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static javax.ws.rs.HttpMethod.DELETE;
@@ -189,10 +188,6 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   }
 
   // ==========================================================================
-
-  static final String DEFAULT_UNIX_ENDPOINT = "unix:///var/run/docker.sock";
-  private static final String DEFAULT_HOST = "localhost";
-  private static final int DEFAULT_PORT = 2375;
 
   private static final String UNIX_SCHEME = "unix";
 
@@ -1685,9 +1680,8 @@ public class DefaultDockerClient implements DockerClient, Closeable {
    * @throws DockerCertificateException if we could not build a DockerCertificates object
    */
   public static Builder fromEnv() throws DockerCertificateException {
-    final String endpoint = fromNullable(getenv("DOCKER_HOST")).or(defaultEndpoint());
-    final Path dockerCertPath = Paths.get(fromNullable(getenv("DOCKER_CERT_PATH"))
-                                              .or(defaultCertPath()));
+    final String endpoint = DockerHost.endpointFromEnv();
+    final Path dockerCertPath = Paths.get(DockerHost.certPathFromEnv());
 
     final Builder builder = new Builder();
 
@@ -1702,8 +1696,8 @@ public class DefaultDockerClient implements DockerClient, Closeable {
       final String hostText = hostAndPort.getHostText();
       final String scheme = certs.isPresent() ? "https" : "http";
 
-      final int port = hostAndPort.getPortOrDefault(DEFAULT_PORT);
-      final String address = isNullOrEmpty(hostText) ? DEFAULT_HOST : hostText;
+      final int port = hostAndPort.getPortOrDefault(DockerHost.defaultPort());
+      final String address = isNullOrEmpty(hostText) ? DockerHost.defaultAddress() : hostText;
 
       builder.uri(scheme + "://" + address + ":" + port);
     }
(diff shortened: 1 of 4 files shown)


davidxia:
@johnflavin @mattnworb 

addresses https://github.com/spotify/docker-client/issues/440

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1685,9 +1684,11 @@ public static Builder builder() {
    * @throws DockerCertificateException if we could not build a DockerCertificates object
    */
   public static Builder fromEnv() throws DockerCertificateException {
-    final String endpoint = fromNullable(getenv("DOCKER_HOST")).or(defaultEndpoint());
+    final DockerUtils dockerUtils = new DockerUtils();
+    final String endpoint = fromNullable(getenv("DOCKER_HOST"))
```
johnflavin: This exact series of method calls—`fromNullable(getenv("DOCKER_HOST")).or(dockerUtils.defaultDockerEndpoint());`—is used multiple times in different places. It seems to me this would be better as a single method in `DockerUtils`, something like `getHost()`.

Same with `fromNullable(getenv("DOCKER_CERT_PATH")).or(dockerUtils.defaultCertPath())`.

davidxia: Ah good point

On Thursday, May 19, 2016, Flavin <notifications@github.com> wrote:

> In src/main/java/com/spotify/docker/client/DefaultDockerClient.java
> <https://github.com/spotify/docker-client/pull/441#discussion_r63953389>:
>
> > @@ -1685,9 +1684,11 @@ public static Builder builder() {
> >     * @throws DockerCertificateException if we could not build a DockerCertificates object
> >     */
> >    public static Builder fromEnv() throws DockerCertificateException {
> > -    final String endpoint = fromNullable(getenv("DOCKER_HOST")).or(defaultEndpoint());
> > +    final DockerUtils dockerUtils = new DockerUtils();
> > +    final String endpoint = fromNullable(getenv("DOCKER_HOST"))
>
> This exact series of method calls—
> fromNullable(getenv("DOCKER_HOST")).or(dockerUtils.defaultDockerEndpoint());—is
> used multiple times in different places. It seems to me this would be
> better as a single method in DockerUtils, something like getHost().
>
> Same with
> fromNullable(getenv("DOCKER_CERT_PATH")).or(dockerUtils.defaultCertPath())
> .
>
> —
> <https://github.com/spotify/docker-client/pull/441/files/ac2c911da1dc2013c62a4f2950d13ae9857742da#r63953389>
>


-- 
David Xia
Software Engineer
dxia@spotify.com

mattnworb: or easier to read IMHO:

```java
endpoint = MoreObjects.firstNonNull(System.getenv("DOCKER_HOST"), dockerUtils.defaultCertPath());
```

src/main/java/com/spotify/docker/client/DockerUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import static java.lang.System.getenv;
+
+/**
+ * A class that contains Docker constants and utility functions.
+ */
+class DockerUtils {
+
+  private SystemDelegate systemDelegate;
+
+  DockerUtils setSystemDelegate(final SystemDelegate systemDelegate) {
```
mattnworb: should be `@VisibleForTesting`

src/main/java/com/spotify/docker/client/DockerUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.System.getProperty;
+import static java.lang.System.getenv;
+
+/**
+ * A class that contains Docker constants and utility functions.
+ */
+class DockerUtils {
+
+  private SystemDelegate systemDelegate;
```
mattnworb: if you set this to a default implementation then you don't have to have null checks below

```java
private SystemDelegate systemDelegate = new SystemDelegate() {
  @Override
  String getProperty(String key) {
    return System.getProperty(key);
  }
  @Override
  String getenv(String name) {
    return System.getenv(name);
  }
};
```

src/test/java/com/spotify/docker/client/DockerHostTest.java
```diff
@@ -23,9 +23,9 @@
 
 import static com.google.common.base.Optional.fromNullable;
 import static com.google.common.base.Strings.isNullOrEmpty;
-import static com.spotify.docker.client.DockerHost.DEFAULT_PORT;
-import static com.spotify.docker.client.DockerHost.DEFAULT_HOST;
-import static com.spotify.docker.client.DockerHost.DEFAULT_UNIX_ENDPOINT;
+import static com.spotify.docker.client.DockerUtils.DEFAULT_HOST;
+import static com.spotify.docker.client.DockerUtils.DEFAULT_PORT;
+import static com.spotify.docker.client.DockerUtils.DEFAULT_UNIX_ENDPOINT;
```
mattnworb: these look unused

mattnworb:
1. I find it a little weird though (and probably bug-prone down the road) that `DockerUtils` has both accessible constants for `DEFAULT_HOST` etc, and methods for properties that have to be computed. I think it should all be the latter, e.g. `defaultHost()`, so that if there is ever a need to replace a constant `DOCKER_HOST` with a calculated value then the foundation is already there. In other words, if we have to use methods to access some properties, then make all of the property access via methods.

2. Couldn't all of this logic just go in `DockerHost`, if that class is meant to "represent a dockerd endpoint? It's kind of confusing to me why DefaultDockerClient uses DockerUtils, and DockerHost uses DockerUtils, but DefaultDockerClient (and everyone else) doesn't just use DockerHost.

src/main/java/com/spotify/docker/client/DockerHost.java
```diff
@@ -17,52 +17,73 @@
...
+  /**
+   * An interface to be mocked during testing.
+   */
+  @VisibleForTesting
+  interface SystemDelegate {
+    String getProperty(String key);
+    String getenv(String name);
+  }
+
+  private static SystemDelegate systemDelegate = new SystemDelegate() {
```
davidxia: Having this as a static attribute that you can set during tests seems bad, but if this is an instance attribute, all the currently static methods like [`endpointFromEnv()`](https://github.com/spotify/docker-client/pull/441/files#diff-0d0cb499d121de8c7a4e50c2a2459ec9R156) would have to go into another static class in order to be used by the static factory method `fromEnv()`. Since `setSystemDelegate()` is package local, I thought this is OK although inelegant.

mattnworb: I think you could get away from this whole thing by having a way to pass in the "env" or "system properties" as a `Map<String, String>` in the test, rather than having to swap out a "delegate".

mattnworb:
👍

src/main/java/com/spotify/docker/client/DockerHost.java
davidxia: @mattnworb Just for my own edification, in cases like this where you want to test a static method that calls a final class' final methods, what are the good approaches, if any? I set the static attribute with a package local static method, but what are other ways?

mattnworb: What is here feels like the best thing that I know of for this case. I suppose it is worth minimizing the chain of static method -> static method calls because of this reason.

PR #444, opened 2016-05-25
Remove unexpected default cert path in DockerHost

davidxia:


src/test/java/com/spotify/docker/client/DockerHostTest.java
```diff
@@ -151,7 +151,7 @@ public void testFromEnv() throws Exception {
     final String dockerHostEnvVar = DockerHost.defaultDockerEndpoint();
     final boolean isUnixSocket = dockerHostEnvVar.startsWith("unix://");
     final URI dockerHostUri = new URI(dockerHostEnvVar);
-    final String dockerCertPathEnvVar = DockerHost.defaultCertPath();
+    final String dockerCertPathEnvVar = null;
```
mattnworb: later in this test, this value is compared to `isNullOrEmpty()` - which is always true - and then another value is compared to this one. It would be simpler to just get rid of the reference and test the other value against `isNull()` or whatever the correct matcher is

src/test/java/com/spotify/docker/client/DockerHostTest.java
```diff
@@ -98,7 +98,7 @@ public void testCertPathFromEnv() throws Exception {
     DockerHost.setSystemDelegate(systemDelegate);
 
     assertThat(DockerHost.certPathFromEnv(), equalTo("foo"));
-    assertThat(DockerHost.certPathFromEnv(), equalTo("bar/.docker"));
+    assertThat(DockerHost.certPathFromEnv(), equalTo(null));
```
mattnworb: how can the same method evaluate to `"foo"` and then `null`? What is the point of returning different values from the mock?

you can also use `nullValue()` instead of `equalTo(null)`

johnflavin: Mockito lets you specify different values to return in sequence when you hit a certain method. You can also tell it to throw exceptions. It's pretty nice.

The point is so you can test different paths of a conditional. You’re testing "When this method (not under test) returns `foo`, I expect my code under test to do a, b, and c, but when it returns `null` I expect x, y, and z to happen."

davidxia: Yes, as @johnflavin said, I wanted to test different return values based on different return values from the mock. Thanks for mentioning `nullValue()` though.

davidxia:
Addressed comments here https://github.com/spotify/docker-client/pull/445

src/test/java/com/spotify/docker/client/DockerHostTest.java
mattnworb: Gotcha, that makes sense. In general I end up writing two separate tests for that.

PR #448, opened 2016-06-01
Fix bug in load

johnflavin:
There are a few different `DefaultDockerClient.load` methods with different signatures, which will use either a user-supplied or default `AuthConfig` and `ProgressHandler`. One of the versions of `load` (starting on line 918) was accepting an `AuthConfig`, but `return`ing another version of `load` that used the default `authConfig`. Effectively, that method ignored the user-supplied `AuthConfig`.

This PR fixes that.


mattnworb:
👍  thanks!

PR #455, opened 2016-06-06
deploy snapshots from Travis after master builds

mattnworb:
inspired by https://gist.github.com/neothemachine/4060735


mattnworb:
@davidxia I thought this might be useful to have in place automatically, what do you think?

mattnworb:
of course, can't really test this out until it is merged

davidxia:
👍

PR #456, opened 2016-06-07
only deploy snapshots to sonatype once

mattnworb:
Since we use a matrix build, the `after_success` block runs for each
build combination. Since the build output is the same in each build, it
is wasteful to upload the JAR to sonatype a bunch of extra times.


mattnworb:
@davidxia

davidxia:
nice workaround 👍

PR #459, opened 2016-06-09
Added support for version:BuildTime

deleSerna:
Hi all,
Please review my chanages for supporting version:buildTime

https://docs.docker.com/engine/reference/api/docker_remote_api/#v1-22-api-changes

GET /version now returns the BuildTime field in RFC3339Nano format to
make it consistent with other date/time values returned by the API.

Thanks and Regards,
Nadeesh T V

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/Version.java
+++ b/src/main/java/com/spotify/docker/client/messages/Version.java
@@ -21,14 +21,13 @@ import com.google.common.base.MoreObjects;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-import java.util.Date;
 import java.util.Objects;
 
 public class Version {
 
   @JsonProperty("ApiVersion") private String apiVersion;
   @JsonProperty("Arch") private String arch;
-  @JsonProperty("BuildTime") private Date buildTime;
+  @JsonProperty("BuildTime") private String buildTime;
   @JsonProperty("GitCommit") private String gitCommit;
   @JsonProperty("GoVersion") private String goVersion;
   @JsonProperty("KernelVersion") private String kernelVersion;
@@ -43,8 +42,8 @@ public class Version {
     return arch;
   }
 
-  public Date buildTime() {
-    return buildTime == null ? null : new Date(buildTime.getTime());
+  public String buildTime() {
+    return buildTime;
   }
 
   public String gitCommit() {
--- a/src/main/java/com/spotify/docker/client/messages/Version.java
+++ b/src/main/java/com/spotify/docker/client/messages/Version.java
@@ -21,13 +21,14 @@ import com.google.common.base.MoreObjects;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import java.util.Date;
 import java.util.Objects;
 
 public class Version {
 
   @JsonProperty("ApiVersion") private String apiVersion;
   @JsonProperty("Arch") private String arch;
-  @JsonProperty("BuildTime") private String buildTime;
+  @JsonProperty("BuildTime") private Date buildTime;
   @JsonProperty("GitCommit") private String gitCommit;
   @JsonProperty("GoVersion") private String goVersion;
   @JsonProperty("KernelVersion") private String kernelVersion;
@@ -42,8 +43,8 @@ public class Version {
     return arch;
   }
 
-  public String buildTime() {
-    return buildTime;
+  public Date buildTime() {
+    return buildTime == null ? null : new Date(buildTime.getTime());
   }
 
   public String gitCommit() {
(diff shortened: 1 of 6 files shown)


src/main/java/com/spotify/docker/client/messages/Version.java
```diff
@@ -27,6 +27,7 @@
 
   @JsonProperty("ApiVersion") private String apiVersion;
   @JsonProperty("Arch") private String arch;
+  @JsonProperty("BuildTime") private String buildTime;
```
johnflavin: Should this be a `Date`, rather than a `String`?

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -436,6 +436,7 @@ public void testVersion() throws Exception {
     final Version version = sut.version();
     assertThat(version.apiVersion(), not(isEmptyOrNullString()));
     assertThat(version.arch(), not(isEmptyOrNullString()));
+    assertThat(version.buildTime(), not(isEmptyOrNullString()));
```
johnflavin: Since `buildTime` was added in API 1.22, this assertion will fail when tested against earlier APIs. Wrap this assertion in an `if(versionAtLeast("1.22"))` condition. (I don't remember exactly what the helper method to test the version is called, but you can find other examples around the test class.)

deleSerna: Thanks for the review. I will change it accordingly.

src/main/java/com/spotify/docker/client/messages/Version.java
deleSerna: Hi John, It will be  a string in RFC3339Nano  format. I added a new method
 parseRFC3339Nano()  in client/DockerDateFormat.java which we can use to parse this.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
deleSerna: I updated with the above suggestion. Please review it

mattnworb:
@deleSerna thanks for the contribution! I am confused about something though - in 64f6365, what code is calling `DockerDateFormat.parseRFC3339Nano()`? It seems to me like it is never used. Is that just in case anyone wants to turn the string date/time into a Date-esque object?

I think I would prefer if `Version.buildTime` was a `java.util.Date` as @johnflavin mentioned - I don't see much value in storing the raw string that the Docker Remote API returns.

deleSerna:
@mattnworb  Thanks for the review.
Yes, I added parseRFC3339Nano() because if someone want nanosec precision they can use it. 

Could I store Version.buildTime as org.threeten.bp.ZonedDateTime ( rather than java.util.Date) sothat I could store it upto the nanosec precision and will use parseRFC3339Nano() for parsing?

mattnworb:
Do you really need nanosecond precision in the build time of the docker package though? I would prefer consistency with the type used to represent dates elsewhere in the source code (`java.util.Date`). Millisecond precision seems like it would be fine, since it is unlikely that there is a huge variance in build times of the docker package.

I am also not crazy about having to replace the `threeten` package with the proper `java.util.time` classes if/when this library supports Java 8 as a minimum version.

deleSerna:
@mattnworb Please review the updated changes

mattnworb:
@deleSerna thanks for addressing the comments. The changes in `Version.java` look good to me. My only remaining comment before merging would be that the `threeten` dependency added in pom.xml and the `DockerDateFormat.parseRFC3339Nano` look to be no longer necessary or useful.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -441,6 +441,9 @@ public void testVersion() throws Exception {
     assertThat(version.kernelVersion(), not(isEmptyOrNullString()));
     assertThat(version.os(), not(isEmptyOrNullString()));
     assertThat(version.version(), not(isEmptyOrNullString()));
+    if (dockerApiVersionAtLeast("1.22")) {
+      assertThat(version.buildTime(), not(isEmptyOrNullString()));
```
mattnworb: this needs to be updated (it fails to compile) since `version.buildTime()` is a Date not a String

deleSerna:
@mattnworb  @johnflavin  Thanks for the comments. Please see the updated changes.

mattnworb:
@deleSerna thanks again. Travis seems to fail for some Docker versions because the date format that Docker uses has changed between releases. 

I am sorry to backtrack but maybe it would be easier to store this value as a String, rather than Date, to avoid trying to have to figure out how to parse differently based on the docker version.

deleSerna:
@mattnworb Please see the updated one

deleSerna:
@mattnworb @johnflavin Could you please review and push it for me

mattnworb:
👍 thanks!

PR #460, opened 2016-06-15
Add support for changing authentication on push.

rgrunber:
It's not currently possible to change the authentication credentials (to push to multiple registries) per call to push(). For now users have to create a separate connection per registry and set the authentication credentials when building the connection.

With this change, users will be able to pass any AuthConfig to push, enabling pushes to various registries using the same connection.

This is bringing the same support to push() as exists for pull().

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -1020,7 +1020,19 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   }
 
   @Override
+  public void push(final String image, final AuthConfig authconfig)
+      throws DockerException, InterruptedException {
+    push(image, new LoggingPushHandler(image), authconfig);
+  }
+
+  @Override
   public void push(final String image, final ProgressHandler handler)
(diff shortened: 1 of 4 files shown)


rgrunber:
Accidentally did a forced push on master, and forgot this points to it. I guess on more reason to use a separate branch per PR.

rgrunber:
codecov/patch seems to imply that the changes I made aren't covered by any tests. In fact, if you look in PushPullIT.java, you'd see that I do in fact test out the exact functionality I'm implementing.

As far as I can tell PushPullIT.java isn't being run because integration tests get run just after 'package' phase, which occurs after the 'test' phase, which is mainly meant for unit tests.

rgrunber:
@mattnworb , am I missing something obvious here to get codecov passing ? If the the builds ran 'verify' instead of just up to 'test', I'm pretty sure PushPullIT would run as well.

mattnworb:
@rgruener you're correct, we aren't running the integration tests in Travis for a reason that escapes me at the moment.

mattnworb:
👍 thanks for the contribution, and sorry for the delayed review/response!

Issue #465, opened 2016-06-28
Fix compatibility with 1.24


velo:
I'm willing to fix this, would just need some pointers

## Description
On v1-24 API copy is no longer available, returning 404

https://docs.docker.com/engine/reference/api/docker_remote_api/#v1-24-api-changes
POST /containers/{name:.*}/copy is now removed and errors out starting from this API version.

## How to reproduce
Install docker 1.12 and try to use copyContainer.  https://github.com/spotify/docker-client/blob/master/docs/user_manual.md#copy-files-or-folders-from-a-container
```
C:\Users\froederm>docker -v
Docker version 1.12.0-rc2, build 906eacd, experimental
```

## Software:

- docker version: 1.12.0-rc2
- docker-client version: 1.12.0-rc2

## Full backtrace

```
com.spotify.docker.client.exceptions.ContainerNotFoundException: Container not found: ecefb611d3273459c4d581f749161f6bc90407e0dff4dc0f3961188dd9d0df7e
	at com.spotify.docker.client.DefaultDockerClient.copyContainer(DefaultDockerClient.java:746)
	at com.vizexplorer.crm.pushpull.containers.PushPullContainer.readLog(PushPullContainer.java:81)
	at com.vizexplorer.crm.pushpull.it.PushPullLoggerRule.failed(PushPullLoggerRule.java:37)
	at org.junit.rules.TestWatcher.failedQuietly(TestWatcher.java:84)
	at org.junit.rules.TestWatcher.access$300(TestWatcher.java:46)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:62)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
[...]

PR #466, opened 2016-06-28
copy is removed from 1.24 

velo:
Fix for https://github.com/spotify/docker-client/issues/465

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -34,6 +34,7 @@ import com.spotify.docker.client.exceptions.ImageNotFoundException;
 import com.spotify.docker.client.exceptions.NetworkNotFoundException;
 import com.spotify.docker.client.exceptions.NotFoundException;
 import com.spotify.docker.client.exceptions.PermissionException;
+import com.spotify.docker.client.exceptions.UnsupportedApiVersionException;
 import com.spotify.docker.client.messages.AuthConfig;
 import com.spotify.docker.client.messages.AuthRegistryConfig;
 
@@ -71,6 +72,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.net.HostAndPort;
 
+import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.utils.IOUtils;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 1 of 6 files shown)


KostyaSha:
How would apiVersion check work with swarm?

velo:
@KostyaSha I guess the way it is already....

I made no changes to it

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -729,6 +729,21 @@ public InputStream exportContainer(String containerId)
   @Override
   public InputStream copyContainer(String containerId, String path)
       throws DockerException, InterruptedException {
+
+    final String apiVersion = version().apiVersion();
+    final int versionComparison = compareVersion(apiVersion, "1.20");
+
+    // Version below 1.20
+    if (versionComparison < 0) {
+      return doCopyContainer(containerId, path);
```
johnflavin: This doesn't seem like the right design to me. I would not expect the `copyContainer` method to do different things depending on the version of docker you’re connected to. I would think there would be two different methods for the two different API endpoints: `copyContainer` that wraps `/containers/{id}/copy` and `archiveContainer` that wraps `/containers/{id}/archive`. `DockerClient.copyContainer` should be marked `@Deprecated`.

mattnworb:
I agree with @johnflavin. Also #465 mentions API version 1.24 but this PR checks 1.20 - why is that?

mattnworb:
@KostyaSha I am not very familiar with Swarm, but how does that related to checking the version of the Docker Remote API used to communicate with the docker daemon?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
velo: ok, I can do that, thanks

velo:
This option is available since 1.20 why is why.... but mandatory only on 1.24m anyway, I will apply the asked change witch will pretty much remove this

velo:
@johnflavin @mattnworb done!

src/main/java/com/spotify/docker/client/exceptions/UnsupportedApiVersionException.java
```diff
@@ -0,0 +1,30 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.docker.client.exceptions;
+
+public class UnsupportedApiVersionException extends NotFoundException {
```
mattnworb: why does this extend NotFoundException?

mattnworb: why does this extend NotFoundException? That seems confusing and unnecessary.

mattnworb: Why does this extend NotFoundException? That seems unnecessary.

mattnworb:
Thanks @velo for making the requested updates. My only remaining concern is that the UnsupportApiVersionException extends NotFoundException - why is that? It doesn't seem as if the two are related.

I'd be happy to merge if that issue was taken care of.

src/main/java/com/spotify/docker/client/exceptions/UnsupportedApiVersionException.java
velo: My bad, fixed

velo:
@mattnworb done

mattnworb:
@velo thanks!

quickwind:
@mattnworb When will this be released? Seems no new version since Jun 8....

F-Scippacercola:
Hi,

I was looking to the corrections made in the API due to the deprecation of the copy endpoint.

Instead of just making the interface of copyCointainer fail, when docker version is >= 1.24, I would suggest to forward the invocation of copy to the new archive endpoint, but preserving the interface of the original call, in order to keep the library backward-compatible. 

Moreover, it seems that the new archive endpoint does not return all the exceptions of the new API, such as "NotFoundException", that should be handled by the library.

velo:
Sure, go for it

johnflavin:
I don’t agree that `copy` should silently redirect to `archive` depending on the version of docker you are connected to. In fact, that was the original design of this PR and I argued against it [here](https://github.com/spotify/docker-client/pull/466#discussion_r69149083). My opinion is that the methods in `docker-client` should reflect the docker REST API endpoints they connect to. So when they deprecate `copy` in favor of `archive`, we do the same.

PR #468, opened 2016-06-30
Fix annotation usage for the latest version of httpcore.

mbooth101:
In httpcore 4.4.5, there was a legal requirement to remove these
annotations. See https://issues.apache.org/jira/browse/HTTPCLIENT-1743
for more details. This patch makes use of alternative annotations
to fix the build against httpcore 4.4.5 and newer.

Signed-off-by: Mat Booth <mat.booth@redhat.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,6 +131,11 @@
       <version>4.5</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpcore</artifactId>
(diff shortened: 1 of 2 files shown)


mbooth101:
This is based on the patch that we carry in Fedora Linux: http://pkgs.fedoraproject.org/cgit/rpms/docker-client.git/tree/httpcomponents-annotations.patch

mattnworb:
thanks!

PR #470, opened 2016-07-01
Support usage of archive to copy files from a container

bradbeck:
`DefaultDockerClient.copyContainer(...)` does not work with Docker Remote API version 1.24. This add `DefaultDockerClient.copyFromContainer(...)` to take advantage of `archive` functionality that has been in Docker Remote API since 1.20 and is currently used by `DefaultDockerClient.copyToContainer(...)`.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -810,6 +810,25 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   }
 
   @Override
+  public InputStream copyFromContainer(String containerId, String path)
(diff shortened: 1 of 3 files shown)


zepouet:
You can use the 5.0.3-SNAPSHOT to use this feature.
At the moment, you need to pull the source and run "mvn clean install -DskipTests" to install locally the artefact to use it. Be careful if you have any checkstyle problem, comment the checkstyle plugin into pom.xml

I will push a PR about this feature. It is right to use it but difficult to gather the file.

Thanks a lot for everything ! Spotify is awesome :)

davidxia:
Rebased here https://github.com/spotify/docker-client/pull/493

Issue #464, opened 2016-06-23
docker.load() seems to strip off metadata


talademo:
I am using Docker 1.10.2 and Spotify 5.0.2
When I load an image tarball using docker.load() it succeeds. However, when I look at the meta data (Labels, Config) using "docker inspect <image-name>" via CLI, the metadata is missing (Labels are lost, ...)

Does anyone see similar issues?

johnflavin:
I think you’re running into a part of this library that is not fully implemented and—in my opinion—is confusingly named.

It seems you’re trying to use the [`/images/load` docker API endpoint](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#load-a-tarball-with-a-set-of-images-and-tags-into-docker). The method `DefaultDockerClient.load()` does not wrap `/images/load` endpoint, as the name would imply. Instead, it wraps the [`/images/create?fromSrc=-` docker API endpoint](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#create-an-image). The two endpoints are similar, but, as you've seen, they aren't the same. The key difference is that `/images/load` creates all the image layers contained in the tarball along with all the metadata, whereas `/images/create` makes a single image from the combined contents or the tarball. ...I think.

This issue was also brought up in #154, which is currently closed—erroneously, in my opinion.

I'll eventually get around to implementing `/images/load` in my huge 1.18 API PR #427, but I've not been progressing particularly quickly. It would get done a lot faster if you or someone else could make a smaller PR just focusing on this issue.

davidxia:
Fixed by #494

Issue #154, opened 2015-04-03
Support loading image tarball


stephan-huttenhuis:
Hi guys,

It would be great if it were possible to load an image tarball through the DockerClient.

https://docs.docker.com/reference/api/docker_remote_api_v1.17/#load-a-tarball-with-a-set-of-images-and-tags-into-docker

Thanks

davidxia:
Finally closed by https://github.com/spotify/docker-client/pull/301

stephan-huttenhuis:
The implemented solution does not work for me because an images/create is done instead of an image/load. The image ends up with only a single layer.

johnflavin:
+1 to reopen. #301 implemented `/images/create`, not `/images/load`.

davidxia:
Fixed by #494

PR #471, opened 2016-07-08
Disentangle /images/create and /images/load

johnflavin:
Fix for #154 and #464.

Previously, the various `DockerClient.load(...)` methods all wrapped the `/images/create` docker API endpoint, not the `/images/load` endpoint as the method name implies. In fact, no methods in `DockerClient` wrapped `/images/load`.

This PR:
* Deprecates all existing `DockerClient.load(...)` methods;
* Creates two `DockerClient.create(...)` methods, the implementations of which perform the functions that `DefaultDockerClient.load(...)` used to perform;
* Creates a new `DockerClient.load(ImageStream)` method, the implementation of which calls `/images/load`.


johnflavin:
I'm going to modify one other thing and create a new PR

PR #476, opened 2016-07-21
Change return type of DockerClient.archiveContainer() to plain InputStream

bentmann:
While I appreciate the feature addition from 23f3c29ae7205d2b4f483a1a4df7db6025774f2f, I question the new method signature. Handing out a `TarArchiveInputStream` makes assumptions about the caller's use case and limits the general usability of the new API. Wrapping a plain input stream into a `TarArchiveInputStream` is easy to do in calling code, the opposite however of getting to the plain stream (e.g. to simply write it out to disk) from `TarArchiveInputStream` is not.


mattnworb:
Makes a lot of sense. Thanks for the contribution @bentmann

bentmann:
Thanks for the quick merge!

PR #478, opened 2016-07-25
added --memory-reservation attribute for docker run

chochreiner:
implemented the --memory-reservation attribute to allow for soft memory limits according to https://docs.docker.com/engine/reference/run/#/runtime-constraints-on-resources


mattnworb:
👍  thanks!

PR #483, opened 2016-08-15
deprecate copyContainer and delegate to archiveContainer

stempler:
...if API version is 1.24 or greater. This is intended to prevent breaking existing applications that use `copyContainer`, as it will be apparent only at runtime that the command is not supported.

I thought as long as the behavior is the same, `copyContainer` can be easily backed by `archiveContainer`, giving users some time to migrate.


stempler:
Looking at the build logs I don't think that the changes broke the build:
```
The command "docker version" failed and exited with 1 during .
```

davidxia:
@stempler Thanks for your PR.

I'm a little hesitant to merge this since we've already released the breaking change. Wouldn't this confuse users?

BTW, the test failures are now fixed by https://github.com/spotify/docker-client/pull/496

PR #490, opened 2016-09-12
Adding ShmSize HostConfig parameter support

smecsia:
The `ShmSize` parameter of `HostConfig` is missing in current version of docker-client. However this is a part of [v1.22](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/).


davidxia:
Rebased here #497

PR #491, opened 2016-09-13
Add "Swarm Mode" object model and methods

derekadams:
We needed access to the new APIs added for "Swarm Mode" in 1.12+, so we have added most of the objects and REST methods required to work with services and tasks. We also added the call for inspecting the Swarm. There are still a few missing pieces, but it should be useful to people that want to run against the latest APIs. We based the objects directly on the ones in the Docker .go files (including the package structure). See com.spotify.docker.client.SwarmModeDockerClientTest for coverage.


mattnworb:
Hey @derekadams thanks so much for the contribution! It might take some time to be able to do a good review of this since there is a large amount of code being added.

It looks like travis fails because of a few checkstyle violations in the test code - can you fix that?

I have one small question: why add the Gradle build script? I'd worry that if there are two build scripts in this repo (maven + gradle), the one that is unused by the maintainers / CI / etc will become unmaintained and stale over time.

derekadams:
Hi @mattnworb I will take a look at the checkstyle violations and fix them. I should have run the Maven build to test it out with the correct settings. Regarding the Gradle build, I can remove it. I needed to compile locally against the new code in our project's Gradle build to make sure the transitive dependencies were included, but I can see how having it in there might be confusing and cause maintenance problems.

ketan:
Hi,

I was looking for exactly this PR. Anything I can do to help get this merged?

derekadams:
I got sidetracked after the original updates. The only thing missing right now is that the code coverage plugin is complaining about the lack of unit tests. There is a unit test class in

    com.spotify.docker.client.SwarmModeDockerClientTest

but it's not written to work with the existing build environment. I temporarily removed the test annotations, but ran into the coverage errors. If the test setup code is done like the standard client test (env rather than a hard-coded local address), it should fix the code coverage complaints. I was planning to get back to fixing it, but it may be a few more days before I can do it.

ketan:
@derekadams — was there any specific reason to introduce a `DefaultSwarmModeDockerClient`, when it would have been possible to add swarm methods to `DefaultDockerClient`?

derekadams:
@ketan - Originally, I thought it might be confusing for users to have swarm mode methods in the default client since older versions would not support the functionality. In retrospect, it's probably better to just include the methods in the default client and add comments or annotations to indicate that 1.12 is required. That would definitely make things cleaner.

ketan:
Could someone close this PR, I've opened #499 instead.

davidxia:
Closed in favor of #499

Issue #481, opened 2016-08-03
Add support oom-kill-disable option suggestion


jazzhuang:
## Description

$docker -v
Docker version 1.9.1, build 827708e BUILDTIME:2016-04-11 14:17:18
$docker run --help | grep oom
  --oom-kill-disable=false        Disable OOM Killer

I found there is no option for setting it in HostConfig.java.

Does I miss something?

jazzhuang:
This client has support it yet, ref: https://github.com/docker-java/docker-java/blob/ca6a7f307ce85b1b33713ca38eb44ae1754936f2/src/main/java/com/github/dockerjava/api/model/HostConfig.java#L139

davidxia:
Closed by #498

PR #492, opened 2016-09-16
Add missing OOM attributes to HostConfig

christofferh:
Adds:
- --oom-kill-disable
- --oom-score-adj

 https://docs.docker.com/engine/reference/run/#/runtime-constraints-on-resources

Resolves https://github.com/spotify/docker-client/issues/481

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/HostConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/HostConfig.java
@@ -64,6 +64,8 @@ public class HostConfig {
   @JsonProperty("IpcMode") private String ipcMode;
   @JsonProperty("Ulimits") private ImmutableList<Ulimit> ulimits;
   @JsonProperty("PidMode") private String pidMode;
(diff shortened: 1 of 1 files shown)


src/main/java/com/spotify/docker/client/messages/HostConfig.java
```diff
@@ -207,6 +211,10 @@ public String ipcMode() {
     return ulimits;
   }
 
+  public Boolean oomKillDisable() { return oomKillDisable; }
```
johnflavin: This line (and the other getters added) cause your build to fail for code style.
Change
```java
public Boolean oomKillDisable() { return oomKillDisable; }
```
to
```java
public Boolean oomKillDisable() {
  return oomKillDisable;
}
```

christofferh: Missed that, thanks 👍  
Fixed and squashed.

davidxia:
Rebased here #498

PR #493, opened 2016-09-19
Support usage archive to copy files from a container

davidxia:


davidxia:
@bradbeck @mattnworb

davidxia:
I think this functionality is already in master: https://github.com/spotify/docker-client/blame/master/src/main/java/com/spotify/docker/client/DefaultDockerClient.java#L775

PR #494, opened 2016-09-19
Changes to loading and saving

davidxia:
Fix for #154 and #464.

## Load and Create
Previously, the various `DockerClient.load(...)` methods all wrapped the `/images/create` docker API endpoint, not the `/images/load` endpoint as the method name implies. In fact, no methods in `DockerClient` wrapped `/images/load`.

This PR:
* Deprecates all existing `DockerClient.load(...)` methods;
* Creates two `DockerClient.create(...)` methods, the implementations of which perform the functions that `DefaultDockerClient.load(...)` used to perform;
* Creates a new `DockerClient.load(ImageStream)` method, the implementation of which calls `/images/load`.

## Save
* There was no reason to use the `AuthConfig` when calling `/images/{name}/get`. The method `DockerClient.save(String, AuthConfig)` has been deprecated.
* The other `save` method, `DockerClient.save(String)` has been changed to `DockerClient.save(String...)`. This is to make use of the two related APIs `/images/{name}/get` and `/images/get?names=name1&names=name2...`.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -1050,22 +1050,33 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   }
 
   @Override
-  public InputStream save(final String image)
+  public InputStream save(final String... images)
       throws DockerException, IOException, InterruptedException {
-    return save(image, authConfig);
-  }
-
-  @Override
-  public InputStream save(final String image, final AuthConfig authConfig)
-      throws DockerException, IOException, InterruptedException {
-    final WebTarget resource = resource().path("images").path(image).path("get");
+    WebTarget resource;
+    if (images.length == 1) {
+      resource = resource().path("images").path(images[0]).path("get");
+    } else {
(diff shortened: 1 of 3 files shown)


davidxia:
@mattnworb The client will support both now. I think @johnflavin was just trying to make the method names match the API endpoint names. Does that make sense?

mattnworb:
I think I misunderstood why the existing `DockerClient.load(..)` methods were being marked as deprecated. Is it simply that some of the existing methods are confusing in that they do not match the names of the Docker Remote API methods that they call, so they are deprecated to point people to use the better-named methods?

mattnworb:
also the TravisCI errors look completely unrelated.

davidxia:
@mattnworb 

> Is it simply that some of the existing methods are confusing in that they do not match the names of the Docker Remote API methods that they call, so they are deprecated to point people to use the better-named methods?

Yup

davidxia:
@mattnworb Any idea why travis fails for those two docker versions?

mattnworb:
@davidxia I think some assumption we made in our script that installs a specific docker version in travis's environment must have changed or been broken by them

PR #499, opened 2016-09-22
Add swarm mode support

ketan:
This was based off #491, with a few changes -
* all commits were squashed into one
* the `SwarmModeDockerClient` and `DefaultSwarmModeDockerClient` subclasses
  were removed and pushed into `DockerClient` and `DefaultDockerClient`

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -2,6 +2,7 @@
  * Copyright (c) 2014 Spotify AB.
  * Copyright (c) 2014 Oleg Poleshuk.
  * Copyright (c) 2014 CyDesign Ltd.
+ * Copyright (c) 2016 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +20,19 @@
 
 package com.spotify.docker.client;
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.google.common.base.Supplier;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.io.CharStreams;
+import com.google.common.net.HostAndPort;
 import com.spotify.docker.client.exceptions.BadParamException;
 import com.spotify.docker.client.exceptions.ConflictException;
 import com.spotify.docker.client.exceptions.ContainerNotFoundException;
@@ -34,14 +48,12 @@ import com.spotify.docker.client.exceptions.ImageNotFoundException;
 import com.spotify.docker.client.exceptions.NetworkNotFoundException;
 import com.spotify.docker.client.exceptions.NotFoundException;
 import com.spotify.docker.client.exceptions.PermissionException;
+import com.spotify.docker.client.exceptions.ServiceNotFoundException;
+import com.spotify.docker.client.exceptions.TaskNotFoundException;
 import com.spotify.docker.client.exceptions.UnsupportedApiVersionException;
 import com.spotify.docker.client.exceptions.VolumeNotFoundException;
 import com.spotify.docker.client.messages.AuthConfig;
 import com.spotify.docker.client.messages.AuthRegistryConfig;
-
-import com.google.common.base.Strings;
-import com.google.common.io.CharStreams;
-
 import com.spotify.docker.client.messages.Container;
 import com.spotify.docker.client.messages.ContainerChange;
 import com.spotify.docker.client.messages.ContainerConfig;
@@ -61,23 +73,16 @@ import com.spotify.docker.client.messages.NetworkConfig;
 import com.spotify.docker.client.messages.NetworkCreation;
 import com.spotify.docker.client.messages.ProgressMessage;
 import com.spotify.docker.client.messages.RemovedImage;
+import com.spotify.docker.client.messages.ServiceCreateOptions;
+import com.spotify.docker.client.messages.ServiceCreateResponse;
 import com.spotify.docker.client.messages.TopResults;
 import com.spotify.docker.client.messages.Version;
 import com.spotify.docker.client.messages.Volume;
 import com.spotify.docker.client.messages.VolumeList;
-
-import com.fasterxml.jackson.core.JsonGenerator;
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.node.JsonNodeFactory;
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Optional;
-import com.google.common.base.Preconditions;
-import com.google.common.base.Supplier;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
-import com.google.common.net.HostAndPort;
-
+import com.spotify.docker.client.messages.swarm.Service;
+import com.spotify.docker.client.messages.swarm.ServiceSpec;
+import com.spotify.docker.client.messages.swarm.Swarm;
+import com.spotify.docker.client.messages.swarm.Task;
 import org.apache.commons.compress.utils.IOUtils;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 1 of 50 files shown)


ketan:
I'm at a loss to understand why checkstyle is failing even on master. But the build is apparently green. https://travis-ci.org/spotify/docker-client/branches

Does someone have any pointers?

johnflavin:
I looked at your latest builds. Checkstyle is passing, so I guess you managed to fix whatever it found distasteful.

But all the tests are failing. Like, every one. I think it could be the unix socket connection. All the test failures are trying to connect to `unix://localhost:80`. Typically the socket connections are to `unix:///var/run/docker.sock`. 

(Total speculation follows:) are you testing this code locally with your docker daemon open to http connections? Because if I remember correctly the daemon on the test machines is not; the only way to connect is `/var/run/docker.sock`.

ketan:
Pretty sure I did not change any setup, in fact I'm on a Mac, so nothing
running on port 80. I suspect it is Travis doing something funny.

Will investigate further.

ketan:
Allright, the build is now green, however the coverage is down because of the plethora of classes that represent the JSON data.

If all is good, can I expect a 5.0.3 release?

ketan:
BTW — support for `docker node` and friends is not part of this PR, if you're OK merging this PR, I'll submit another PR for `docker node` support.

src/main/java/com/spotify/docker/client/messages/mount/Driver.java
```diff
@@ -0,0 +1,97 @@
+/*
```
ketan: Looking at the api docs, for a few references to where this class is used, it appears that the class is just a generic thing that describes a driver name, and options for a variety of things, including networks and tasks.

ketan:
@davidxia — I've changed all timestamp fields from `String` to `java.util.Date`.

> In src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java:
> 
> > @@ -3041,6 +3064,198 @@ public void testIpcMode() throws Exception {
>  
>    }
>  
> +  @Test
> +  public void testInspectSwarm() throws Exception {
> +    requireDockerApiVersionAtLeast("1.24", "swarm support");
> +
> +    final Swarm swarm = sut.inspectSwarm();
> +    assertThat(swarm.createdAt(), is(not(isEmptyOrNullString())));
> Should we check not null for the other fields? Seems like we should check for ID at least.
> 
> 

I'm not sure what "value" asserting on fields adds, esp for code that appears to declarative in nature. I've added a test to assert the swarm `id` attribute as you requested.

ketan:
Also, I could not help but notice that the indentation in most of the code is off. Some files are indented using 2 spaces, others with 4 spaces. What's the deal?

ketan:
All tests are green now

mattnworb:
Since a chunk of the code in this PR was originally contributed by @derekadams, is there any way to maintain that info in the commit author field?

ketan:
I'm not sure how to go about putting multiple authors in the author field. I'm happy to attribute the entire commit to him, I don't care so much about my attribution.

Let me know what you'd prefer, I'd really like to get this merged ASAP :)

johnflavin:
This looks like it will do the trick: http://stackoverflow.com/a/31669789/1288474

ketan:
@johnflavin — Done, travis is green.

ketan:
Bump!

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1539,6 +1552,163 @@ public LogStream execStart(final String execId, final ExecStartParameter... para
...
+    final WebTarget resource = resource().path("services");
+    return request(GET, SERVICE_LIST, resource, resource.request(APPLICATION_JSON_TYPE));
+  }
+
+  @Override
+  public List<Service> listServices(Service.Criteria criteria)
+          throws DockerException, InterruptedException {
+    assertAPIVersionIsAbove("1.24");
+    WebTarget resource = resource().path("services");
+    final Map<String, List<String>> filters = new HashMap<String, List<String>>();
```
davidxia: You can use the diamond type `<>` here. The pom.xml specifies Java 1.7.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1539,6 +1552,163 @@ public LogStream execStart(final String execId, final ExecStartParameter... para
...
   @Override
+  public Swarm inspectSwarm() throws DockerException, InterruptedException {
+    assertAPIVersionIsAbove("1.24");
+
+    final WebTarget resource = resource().path("swarm");
+    return request(GET, Swarm.class, resource, resource.request(APPLICATION_JSON_TYPE));
+  }
+
+  @Override
+  public ServiceCreateResponse createService(ServiceSpec spec, ServiceCreateOptions options)
```
davidxia: nitpick: Can you use `final` for all the parameter arguments in this class?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1539,6 +1552,163 @@ public LogStream execStart(final String execId, final ExecStartParameter... para
...
+        case 404:
+          throw new ServiceNotFoundException(serviceId);
+        default:
+          throw e;
+      }
+    }
+  }
+
+  @Override
+  public void updateService(String serviceId, Long version, ServiceSpec spec)
```
davidxia: Same as comment here https://github.com/spotify/docker-client/pull/499/files#diff-163ee817c343ffd4cbc08d4c3412a8a4R1563

src/main/java/com/spotify/docker/client/messages/swarm/Service.java
```diff
@@ -0,0 +1,160 @@
...
+        public String getServiceName() {
+            return serviceName;
+        }
+
+        public void setServiceName(String serviceName) {
+            this.serviceName = serviceName;
+        }
+    }
+
+    public static class CriteriaBuilder {
```
davidxia: For consistency, can you use the builder pattern in `ContainerConfig`? Also make sure the access modifiers for classes and methods are as strict as they can be.

src/main/java/com/spotify/docker/client/messages/swarm/Task.java
```diff
@@ -0,0 +1,277 @@
...
+        public String getDesiredState() {
+            return desiredState;
+        }
+
+        public void setDesiredState(String desiredState) {
+            this.desiredState = desiredState;
+        }
+    }
+
+    public static class CriteriaBuilder {
```
davidxia: For consistency, can you use the builder pattern in `ContainerConfig`? Also make sure the access modifiers for classes and methods are as strict as they can be.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -3267,6 +3290,199 @@ public void testOomScoreAdj() throws Exception {
...
+
+    final Task task = sut.listTasks().get(1);
+
+    final List<Task> tasksWithId = sut.listTasks(Task.find().withTaskId(task.id()).build());
+
+    assertThat(tasksWithId.size(), is(1));
+    assertThat(tasksWithId.get(0), equalTo(task));
+  }
+
+  private ServiceSpec createServiceSpec(String serviceName) {
```
davidxia: nitpick: final for parameter.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -3267,6 +3290,199 @@ public void testOomScoreAdj() throws Exception {
...
+  }
+
+  @Test
+  public void testInspectTask() throws Exception {
+    requireDockerApiVersionAtLeast("1.24", "swarm support");
+
+    final ServiceSpec spec = createServiceSpec(randomName());
+    assertThat(sut.listTasks().size(), is(0));
+    final ServiceCreateResponse response = sut.createService(spec, new ServiceCreateOptions());
+    Thread.sleep(2000); // to give it a while to spin containers
```
davidxia: I'm not a fan of these sleeps. We should use something like https://github.com/awaitility/awaitility, but since we have them littered throughout the tests already, we can do that in another PR.

davidxia:
@ketan Thanks so much for your patience. I took a closer look. I think we should be good to once you address my comments.

src/main/java/com/spotify/docker/client/messages/swarm/RestartPolicy.java
```diff
@@ -0,0 +1,123 @@
...
+
+import java.util.Objects;
+
+import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
+import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.NONE;
+
+@JsonAutoDetect(fieldVisibility = ANY, getterVisibility = NONE, setterVisibility = NONE)
+public class RestartPolicy {
+
+    public static final String RESTART_POLICY_NONE = "none";
```
ketan: The docs here https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/create-a-service specify just these 3 policies.

davidxia:
replaced by https://github.com/spotify/docker-client/pull/514

PR #501, opened 2016-09-23
Use BigInteger for MemoryStats.hierarchicalMemoryLimit

davidxia:
Some computers will have more memory limits higher than
Long can represent.

This was breaking DefaultDockerClientTest.testStats() on some hosts.


davidxia:
@mattnworb

davidxia:
@rohansingh @gimaker

rohansingh:
I don't see the need for this, 640K ought to be enough memory for anybody.

PR #503, opened 2016-09-23
Update docs

johnflavin:
* Update docs to reflect additions and changes to the library, mostly from #427 but also assorted other PRs.
* Minor javadoc updates
* Deprecate `copyContainer` (it should have been deprecated already; it throws an `UnsupportedApiVersionException` for API ≥ 1.24)
* Suppress deprecation warnings for tests of deprecated methods

### DIFF
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -138,7 +138,9 @@ final ContainerInfo info = docker.inspectContainer("containerID");
 
 ### List processes running inside a container
 
-Not implemented. PRs welcome.
+```java
+final TopResults topResults = docker.topContainer("containerID", "ps_args");
+```
 
 ### Get container logs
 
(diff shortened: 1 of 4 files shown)


docs/user_manual.md
```diff
@@ -443,11 +475,22 @@ try (OutputStream imageOutput = new BufferedOutputStream(new FileOutputStream(im
...
+    while ((read = imageInput.read(buffer)) > -1) {
+      // Do stuff with the tar stream of images
+    }
+}
+```
 
 ### Load a tarball with a set of images and tags into docker
 
-Not implemented yet. PRs welcome.
+```
```
davidxia: You can also add `java` here.

src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -1080,7 +1080,9 @@ public static RemoveContainerParam forceKill(final boolean force) {
    *                              if container is not found (404)
    * @throws DockerException      if a server error occurred (500)
    * @throws InterruptedException If the thread is interrupted
+   * @deprecated Replaced by {@link #archiveContainer(String, String)}
```
davidxia: Can we add a note that this was deprecated since 1.20 and will throw `UnsupportedApiVersionException` for 1.24 and above?

davidxia:
@johnflavin Awesome. Thanks for this.

PR #508, opened 2016-09-28
Use official Docker in Docker image for tests

davidxia:
Installing lxc-docker-1.6.2 in Docker caused machines running
Docker 1.12.1 to crash.


davidxia:
@mattnworb @rohansingh

PR #509, opened 2016-09-28
Test with Docker 1.12.1 instead of 1.12.0 on Travis

davidxia:
Update README. Upload artifact tested on 1.12.1 to Sonatype instead
of the one tested on 1.6.0.


davidxia:
@mattnworb

Issue #510, opened 2016-09-29
Support nocopy bind option


dmandalidis:
nocopy options in binds was introduced in 1.24 API version. However, implementation allows to specify it in Bind#to property but I don't know if this was done on purpose....otherwise I could send a PR

johnflavin:
Can you quote from and/or give a link to the docker API documentation for this option?

dmandalidis:
First time it appears on https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/ (search for 'nocopy') but it was really introduced at 1.23. (https://docs.docker.com/engine/reference/api/docker_remote_api/#/v1-23-api-changes)

> POST /containers/create now allows specifying nocopy for named volumes, which disables automatic copying from the container path to the volume.
>

johnflavin:
Ok, I see that it was listed in the API changes for v1.23. However, I can't find "nocopy" in the [API 1.23](https://github.com/docker/docker/blob/master/docs/reference/api/docker_remote_api_v1.23.md) docs at all, and the only mention in the [1.24 docs](https://github.com/docker/docker/blob/master/docs/reference/api/docker_remote_api_v1.24.md#create-a-service) is in reference to creating a swarm, not creating a container. (Which is not to say that we don't want that swarm option. But it may already be included in #499. Not sure.)

I think this is a very doable PR. But I would feel more comfortable with it if you first opened an issue with docker/docker saying their API docs for 1.23, 1.24, and 1.25 should be updated to list this option. That way we can be sure we're doing the correct thing on our end.

dmandalidis:
You 're right it is only mentionned in the API changelog for 1.23, while in the website it is mentioned as a general mode for volume creating (compared to the respective .md). However, I 've already submitted a PR at docker-java/docker-java with a passing integration tests in 1.23 API (https://github.com/docker-java/docker-java/pull/696) while I 'm also using it in real life (through abusing the Bind#to of yours and through shell :) ). Here's the docker issue I 've just opened : https://github.com/docker/docker/issues/27029

johnflavin:
Cool, thanks for opening the docker ticket. If you’re already familiar enough with the nocopy option in the API, then feel free to get started on a PR for it. I wouldn’t want to do the PR myself yet, because I’m not 100% certain how the option is implemented. Nor would I want to merge it until docker updates the docs (but I can’t merge anything here anyway, so that point is moot). But none of that precludes you from working on it now.

dmandalidis:
[...]

PR #511, opened 2016-09-29
Support nocopy option in volume binds (#510)

dmandalidis:


src/main/java/com/spotify/docker/client/messages/HostConfig.java
```diff
@@ -1036,7 +1038,12 @@ public String toString() {
       } else if (from == null || from.equals("")) {
         return to;
       } else if (readOnly == null || !readOnly) {
-        return from + ":" + to;
+          if (noCopy == null || !noCopy) {
```
johnflavin: Are `noCopy` and `readOnly` mutually exclusive? As in, you can only ever have `from:to:ro` or `from:to:nocopy` but never `from:to:ro:nocopy`?

dmandalidis: Yeap (doesn't make sense after all)

`mandas@docker-1:~$ docker volume create --name foo
foo

mandas@docker-1:~$ docker run -v foo:/tmp:ro:nocopy busybox
docker: Error response from daemon: Invalid bind mount spec "foo:/tmp:ro:nocopy": Invalid volume specification: 'foo:/tmp:ro:nocopy'.
See 'docker run --help'.
`

dmandalidis: Phew! This option is cursed, it is supported in bind mounts although the man page says it so :

https://github.com/docker/docker/issues/27046

johnflavin: I checked out [the volume options](https://docs.docker.com/engine/reference/run/#/volume-shared-filesystems) after reading docker/docker#27046. It isn't worded very clearly, but it does say you can set `ro` and `nocopy` together, along with any of the other options.
> The **comma-delimited** `options` are are `[rw|ro]`, `[z|Z]`, `[[r]shared|[r]slave|[r]private]`, and `[nocopy]`.

Emphasis mine.


```bash
$ docker volume create --name foo
foo
$ docker run -v foo:/tmp:ro:nocopy busybox
docker: Error response from daemon: Invalid bind mount spec "foo:/tmp:ro:nocopy": Invalid volume specification: 'foo:/tmp:ro:nocopy'.
See 'docker run --help'.
$ docker run -v foo:/tmp:ro,nocopy busybox
[no error]
```

dmandalidis: So probably, 'ro' has to do with the operation and 'nocopy' with the initialization...I see

dmandalidis:
Ok but I think this goes to another PR (I could take care of this)

PR #514, opened 2016-10-07
Add swarm mode support

davidxia:


davidxia:
Replaces https://github.com/spotify/docker-client/pull/499

ketan:
Could you apply this patch along with this? `Integer` is not `Long` enough to represent anything over 2GB :)


```patch
diff --git a/src/main/java/com/spotify/docker/client/messages/swarm/Resources.java b/src/main/java/com/spotify/docker/client/messages/swarm/Resources.java
index 469a78e..dc50688 100644
--- a/src/main/java/com/spotify/docker/client/messages/swarm/Resources.java
+++ b/src/main/java/com/spotify/docker/client/messages/swarm/Resources.java
@@ -29,16 +29,16 @@ import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.NONE;
 public class Resources {
 
     @JsonProperty("NanoCPUs")
-    private Integer nanoCpus;
+    private long nanoCpus;
 
     @JsonProperty("MemoryBytes")
-    private Integer memoryBytes;
+    private Long memoryBytes;
 
-    public Integer nanoCpus() {
+    public Long nanoCpus() {
         return nanoCpus;
     }
 
-    public Integer memoryBytes() {
+    public Long memoryBytes() {
         return memoryBytes;
     }
 
@@ -46,12 +46,12 @@ public class Resources {
 
         private Resources resources = new Resources();
 
-        public Builder withNanoCpus(int nanoCpus) {
+        public Builder withNanoCpus(long nanoCpus) {
             resources.nanoCpus = nanoCpus;
             return this;
         }
 
-        public Builder withMemoryBytes(int memoryBytes) {
+        public Builder withMemoryBytes(long memoryBytes) {
             resources.memoryBytes = memoryBytes;
             return this;
         }
```

davidxia:
@ketan Thanks for catching. I'm wondering whether to use `long` or `Long`. We mix primitives and boxed types throughout :/ @dflemstr, you're good with Jackson right? Is the difference that primitives with Jackson annotations will get default values and boxed types will be null when deserializing from a JSON string that's missing that field?

e.g.

`{"bar": 2}`

```java
@JsonProperty("foo")
private int foo;
```

-> `foo = 0;`

VS

`{"bar": 2}`

```java
@JsonProperty("foo")
private Integer foo;
```

-> `foo = null;`

mattnworb:
@davidxia irrespective of which json library is used, a primitive `int foo` will have to default to `0` if not set

davidxia:
@mattnworb PTAL

src/main/java/com/spotify/docker/client/exceptions/UnsupportedApiVersionException.java
```diff
@@ -23,8 +23,8 @@ public UnsupportedApiVersionException(final String version, final Throwable caus
     super("Unsupported Api Version: " + version, cause);
   }
 
-  public UnsupportedApiVersionException(final String containerId) {
-    this(containerId, null);
+  public UnsupportedApiVersionException(final String version) {
+    this(version, null);
```
mattnworb: thanks for catching this

ketan:
Thanks for merging!

Issue #488, opened 2016-09-09
Question -> docker run --rm


diemol:
## Description

Hi, 

When using the command line, I start the container like this: "docker run --rm {other options}...".

So far, I am able to start the container without any issues through the docker client, nevertheless I have not found how to add the "-- rm" option.

I saw this [issue](https://github.com/spotify/docker-client/pull/220), but I don't know if it is related to what I want to achieve.

If the option does not exist yet, I could implement it via PR. I just wonder if the [HostConfig](https://github.com/spotify/docker-client/blob/v5.0.2/src/main/java/com/spotify/docker/client/messages/HostConfig.java) class is the right place to do it. 
I see that most of the HostConfig parameters seen [here](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.25/#/create-a-container) are mapped in that class.

Please let me know if it is possible to add the option with an existing method or if I could go ahead and implement it.

Thanks!

johnflavin:
I looked through the docker documentation, and I couldn't find a way to do it. You can check for yourself: https://github.com/docker/docker/tree/master/docs/reference/api. 

If there is no documentation on how to do it, then either
1. There is no way, it doesn't exist.
2. There is a way, but it isn't documented.

I strongly suspect that 2 is the more likely option. In that case, you could open an issue with docker/docker to document the API for the `--rm` option. Once they document that, we'll know where to add the option on this side, and maybe you could do a PR for that.

diemol:
I read again the link I provided for the docker API doc, and the field is there, it is called `AutoRemove`.
The thing is that it is added in the version 1.25 which has not been released (I think).
So, I guess we just need to wait until the new API version is released.

johnflavin:
Yes, it does appear that `HostConfig.AutoRemove` was added in docker/docker#20848, and will be released in API 1.25. It looks like, in all existing docker versions, `--rm` is a bit of client-side trickery that removes the container when the client detects that it has died. Since the client was responsible, there was no daemon API to make it happen. Now with the newer version they moved that function to the daemon, thus it can be exposed through the API.

I think you could add that field to the `HostConfig` class before the Docker version is released without harming anything. You could even add tests for it, wrapped in a conditional `if(apiVersionAtLeast("1.25")) {}` or whatever the helper method is.

diemol:
Great, I will add it and send the PR around next week. 
Thanks.

diemol:
PR sent https://github.com/spotify/docker-client/pull/526

[...]

PR #526, opened 2016-10-12
Adding "autoRemove" field for HostConfiguration.

diemol:
Hi, 

Regarding the discussion related to this [question](https://github.com/spotify/docker-client/issues/488) I had some weeks ago, here is the PR addressing that field I was asking for and that will be added in the [API v.1.25](https://github.com/docker/docker/blob/master/docs/reference/api/docker_remote_api_v1.25.md).

I added also a small unit test to check the field.

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/HostConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/HostConfig.java
@@ -99,6 +99,8 @@ public class HostConfig {
   private Boolean oomKillDisable;
   @JsonProperty("OomScoreAdj")
   private Integer oomScoreAdj;
+  @JsonProperty("AutoRemove")
+  private Boolean autoRemove;
 
   private HostConfig() {
   }
@@ -135,6 +137,7 @@ public class HostConfig {
     this.shmSize = builder.shmSize;
     this.oomKillDisable = builder.oomKillDisable;
     this.oomScoreAdj = builder.oomScoreAdj;
+    this.autoRemove = builder.autoRemove;
   }
 
   public List<String> binds() {
@@ -257,6 +260,10 @@ public class HostConfig {
     return oomScoreAdj;
   }
 
+  public Boolean autoRemove() {
+    return autoRemove;
+  }
+
   @Override
   public boolean equals(final Object o) {
     if (this == o) {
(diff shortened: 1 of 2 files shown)


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -3356,6 +3356,26 @@ public void testOomScoreAdj() throws Exception {
...
+        .image(BUSYBOX_LATEST)
+        .hostConfig(HostConfig.builder()
+                        .autoRemove(true) // Default is false
+                        .build())
+        .build();
+
+    final ContainerCreation container = sut.createContainer(config, randomName());
+    final ContainerInfo info = sut.inspectContainer(container.id());
+
+    assertThat(info.hostConfig().autoRemove(), is(true));
```
johnflavin: In this test, you are just setting the option on container creation and verifying that it was indeed set correctly. A better test would be to start and stop the container, verifying that it does get removed.

diemol: You are right, the test I wrote was quite naive. Anyway, I improved it a bit and just made the commit.
Thanks for the feedback.

diemol:
Hi @mattnworb, 

To test it locally I just faked the behaviour as if the flag was implemented. But I think this does not make sense in the CI and overall, honestly, I would just wait until it gets released.

Right now I have a workaround for that missing feature, so I don't need it urgently, it is more a nice to have.

Thanks for the nice work you are doing over there, this docker-client simplified my life a lot in the project I am working, so the least I could do is to five something back :smile:

src/main/java/com/spotify/docker/client/messages/HostConfig.java
```diff
@@ -1042,6 +1053,15 @@ public Integer oomScoreAdj() {
       return oomScoreAdj;
     }
 
+    public Boolean autoRemove() {
+      return autoRemove;
+    }
+
+    public Builder autoRemove(final Boolean autoRemove) {
```
mattnworb: On second thought I am happy to merge this now, in advance of the 1.25 API being released, but would you mind adding a javadoc comment here that this field only has relevance for versions >= 1.25?

diemol: Sure, I'll do it between today and tomorrow!

davidxia:
I've added the javadoc here https://github.com/spotify/docker-client/pull/538

diemol:
Ah cool @davidxia, you were faster! Thanks!

PR #531, opened 2016-11-01
fix AssertionError in DefaultLogStream

mattnworb:
this is the same bug as https://github.com/spotify/helios/pull/1018


mattnworb:
@davidxia @dflemstr

PR #532, opened 2016-11-01
ignore swarm exceptions in DefaultDockerClientTest teardown

mattnworb:
if the docker instance being tested against is not a part of a swarm,
then listServices will throw an exception as the remote api seems to
return HTTP 406 in this case.

Avoid extra errors during tearDown - if the daemon is not a part of a
docker swarm then there is nothing swarm-related to clean up.


mattnworb:
@davidxia

PR #533, opened 2016-11-04
Add "pidsLimit" to HostConfig

CGNonofr:
Introduced in Docker api 1.11

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/HostConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/HostConfig.java
@@ -99,6 +99,8 @@ public class HostConfig {
   private Boolean oomKillDisable;
   @JsonProperty("OomScoreAdj")
   private Integer oomScoreAdj;
+  @JsonProperty("PidsLimit")
+  private Integer pidsLimit;
 
   private HostConfig() {
   }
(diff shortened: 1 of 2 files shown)


CGNonofr:
Do i have to make a unit test considering that this will only work with docker api >= 1.11 ?

davidxia:
@CGNonofr Thanks for your PR. Yes, please add a test.

CGNonofr:
I made a test and I don't understand why the build failed.
- for docker 1.10.3 the issue doesn't seem to be related to my pull request
- for docker 1.12.1 the test fails, but I don't know why nor how to fix it since it works on the same docker version on my computer!

CGNonofr:
According to the documentation (https://docs.docker.com/engine/reference/commandline/create/), pidsLimit only works with a kernel >= 4.3, i think this is the issue!
What can we do?

davidxia:
@CGNonofr You're right. Travis CI has kernel "Linux testing-gce-afb0b56d-a60a-4bf3-8b8d-bae07747d4c5 3.19.0-66-generic #74~14.04.1-Ubuntu SMP Tue Jul 19 19:56:11 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux". I've added to your PR here https://github.com/spotify/docker-client/pull/537.

PR #541, opened 2016-11-14
Add "TmpFs" to HostConfig

CGNonofr:


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -3381,6 +3381,28 @@ public void testPidsLimit() throws Exception {
 
     assertThat(info.hostConfig().pidsLimit(), is(100));
   }
+  
+  @Test
+  public void testTmpfs() throws Exception {
+    requireDockerApiVersionAtLeast("1.22", "PidsLimit");
```
davidxia: Second param is wrong.

PR #546, opened 2016-11-18
replace pojo boilerplate with AutoValue

davidxia:


mattnworb:
@davidxia my plan for #534 was to go back and create test cases for all classes first to verify nothing is broken by any changes. I was working on a script to generate test cases from the existing code to make it not so tedious. I think the same approach should still be taken - the message serialization would be so easy to slip a bug into.

davidxia:
@mattnworb Great idea. Want to continue that work on this branch? Do you want to do the test cases while I continue converting to AutoValue?

mattnworb:
Well once there is a script to generate test cases from the existing source code it should be easy to also generate auto value using classes too, so I wouldn't spend time converting them by hand.

davidxia:
Added a couple of JSON deserialization tests.

davidxia:
Added `CHANGELOG.md` detailing most of the breaking changes.

davidxia:
@mattnworb Can I haz review? :)

cc @gimaker @rohansingh since this is a big change. The motivation is that AutoValue will make this project a lot more maintainable and less tedious to add new value classes. Let me know if you have questions.

CHANGELOG.md
```diff
@@ -0,0 +1,66 @@
...
+### Misc
+
+```
+-    final Ipam ipam =
+-        Ipam.builder().driver("default").config("192.168.0.0/24", "192.168.0.0/24", "192.168.0.1")
+-            .build();
++    final IpamConfig ipamConfig =
++        IpamConfig.create("192.168.0.0/24", "192.168.0.0/24", "192.168.0.1");
++    final Ipam ipam = Ipam.create("default", singletonList(ipamConfig));
+```
```
mattnworb: it is not entirely clear to me what this part of the document is documenting?

pom.xml
```diff
@@ -136,6 +136,29 @@
...
+      <groupId>com.intellij</groupId>
+      <artifactId>annotations</artifactId>
+      <version>12.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>annotations</artifactId>
+      <version>3.0.1</version>
+      <scope>provided</scope>
+    </dependency>
```
mattnworb: a bit of an aside but don't some of these annotation packages overlap? For example there is a `Nullable` in `com.intellij:annotations` and also `>com.google.code.findbugs:jsr305`. Are we using all of these?

CHANGELOG.md
davidxia: This is meant to be a diff documenting the breaking changes to `Ipam` and `IpamConfig`. I can add a sentence to clarify.

pom.xml
davidxia: Where do you see `@Nullable` in `com.google.code.findbugs:jsr305`? My IntelliJ can't seem to find it.

As for `com.google.code.findbugs:annotations`, it provides `edu.umd.cs.findbugs.annotations.SuppressFBWarnings;`

CHANGELOG.md
mattnworb: I think that would help.

pom.xml
mattnworb: seems to be a `javax.annotation.Nullable` in `com.google.code.findbugs:jsr305:3.0.0` to me - but I suppose any Nullable annotation is fine. I was just curious if each of these are actually needed.

davidxia: Ah yes the javax one. I'm not really sure of the difference between that one and intellij's. I guess there's no functional difference? The intellij one had nicer javadocs, but I'll remove it so we don't have the extra dependency.

davidxia:
@gimaker I suppose that's friendlier to the user. I think I was just a bit lazy at first. I'll try to add deprecated methods when they're straight forward to do so.

davidxia:
@gimaker I've marked the old methods as deprecated now https://github.com/spotify/docker-client/blob/bd8f0307de8049c39df3b1bafa24b0b0e1c6652e/CHANGELOG.md. There are a few breaking changes when required or not straight forward.

davidxia:
I'll make this a release candidate for now in case people have lingering thoughts.

davidxia:
See https://github.com/spotify/docker-client/tree/7.0.0-rc1

PR #549, opened 2016-12-02
Update Events

johnflavin:
* Add event fields from docker updates
* Document the version each had been added
* Deprecate older fields
* Update event tests to assert on new fields
* Refactor event tests (now explicitly test event streaming and polling modes)

### DIFF
--- a/src/main/java/com/spotify/docker/client/DockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DockerClient.java
@@ -2229,8 +2229,9 @@ public interface DockerClient extends Closeable {
      *
      * @param until Return events up until this Unix timestamp.
      * @return {@link EventsParam}
+     * @since API 1.18
      */
-    public static EventsParam until(Long until) {
+    public static EventsParam until(final Long until) {
       return new EventsParam("until", String.valueOf(until));
     }
 
@@ -2239,8 +2240,9 @@ public interface DockerClient extends Closeable {
      *
      * @param since Return events since this Unix timestamp.
      * @return {@link EventsParam}
+     * @since API 1.18
      */
-    public static EventsParam since(Long since) {
+    public static EventsParam since(final Long since) {
       return new EventsParam("since", String.valueOf(since));
     }
 
@@ -2250,8 +2252,9 @@ public interface DockerClient extends Closeable {
      * @param name  Name
      * @param value Value
      * @return {@link EventsParam}
+     * @since API 1.18
      */
-    private static EventsParam filter(String name, String value) {
+    private static EventsParam filter(final String name, final String value) {
       return new EventsFilterParam(name, value);
     }
 
@@ -2259,6 +2262,7 @@ public interface DockerClient extends Closeable {
      * Show only certain events. For example, "event=pull" for image pull events.
      * @param event Type of event to show
      * @return EventsParam
+     * @since API 1.21
      */
     public static EventsParam event(final String event) {
       return filter("event", event);
@@ -2268,6 +2272,7 @@ public interface DockerClient extends Closeable {
      * Show events for an image.
      * @param image An image tag or id
      * @return EventsParam
+     * @since API 1.18
      */
     public static EventsParam image(final String image) {
       return filter("image", image);
@@ -2277,6 +2282,7 @@ public interface DockerClient extends Closeable {
      * Show events for a container.
      * @param container A container name or id
      * @return EventsParam
+     * @since API 1.18
      */
     public static EventsParam container(final String container) {
       return filter("container", container);
@@ -2286,6 +2292,7 @@ public interface DockerClient extends Closeable {
      * Show events for a volume.
      * @param volume A volume name or id
      * @return EventsParam
+     * @since API 1.22
      */
     public static EventsParam volume(final String volume) {
       return filter("volume", volume);
@@ -2295,6 +2302,7 @@ public interface DockerClient extends Closeable {
      * Show events for a network.
      * @param network A network name or id
      * @return EventsParam
+     * @since API 1.22
      */
     public static EventsParam network(final String network) {
       return filter("network", network);
@@ -2304,6 +2312,7 @@ public interface DockerClient extends Closeable {
      * Show events for a daemon.
      * @param daemon A daemon name or id
      * @return EventsParam
+     * @since API 1.24
      */
     public static EventsParam daemon(final String daemon) {
       return filter("daemon", daemon);
@@ -2313,9 +2322,22 @@ public interface DockerClient extends Closeable {
      * Show events of a given type. For instance, "type=image" for all image events.
      * @param type A type of event. Possible values: container, image, volume, network, or daemon
      * @return EventsParam
+     * @deprecated Use {@link #type(EventType)}.
+     * @since API 1.22
      */
+    @Deprecated
     public static EventsParam type(final String type) {
       return filter("type", type);
+    }
+
+    /**
+     * Show events of a given type. For instance, "type=image" for all image events.
+     * @param type A type of event. Possible values: container, image, volume, network, or daemon
+     * @return EventsParam
+     * @since API 1.22
+     */
+    public static EventsParam type(final EventType type) {
+      return filter("type", type.getName());
     }
 
     /**
@@ -2324,6 +2346,7 @@ public interface DockerClient extends Closeable {
      * @param label The label to filter on
      * @param value The value of the label
      * @return EventsParam
+     * @since API 1.21
      */
     public static EventsParam label(final String label, final String value) {
       return isNullOrEmpty(value) ? filter("label", label) : filter("label", label + "=" + value);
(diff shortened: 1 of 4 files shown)


davidxia:
Looks like two event tests still fail on Docker 1.9.1. https://travis-ci.org/spotify/docker-client/jobs/186413456

johnflavin:
I fixed the tests for 1.9. Current build fails on 1.10, but it looks like a random, transient failure. Unrelated to my updates.

davidxia:
@johnflavin Thanks for the PR. I think the 1.10.3 build keeps failing for some reason though.

src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -2332,10 +2355,33 @@ public static EventsParam label(final String label, final String value) {
...
+     * @since API 1.21
      */
     public static EventsParam label(final String label) {
       return label(label, null);
     }
+
+    /**
+     * Valid event types for EventsParam.type
+     */
+    public enum EventType {
```
davidxia: Can we also test this?

src/main/java/com/spotify/docker/client/messages/Event.java
```diff
@@ -91,7 +143,50 @@ public String toString() {
...
+        .add("action", action)
+        .add("actor", actor)
         .add("time", time)
+        .add("timeNano", timeNano)
         .toString();
   }
+
+  public static class Actor {
+    @JsonProperty("ID") private String id;
+    @JsonProperty("Attributes") private ImmutableMap<String, Object> attributes;
```
davidxia: Could this just be `ImmutableMap<String, String>` or do we expect nested JSON objects?

johnflavin:
Yes, that does look like a real failure. I spun up a 1.10.3 docker-machine vm, but I wasn't able to reproduce this.

I added assertions to the test that the stream does have events so that, at the very least, the test can `FAIL` with a message rather than just `ERROR` with an exception.

If this continues to happen on the travis machines, I do have a suspicion about the cause. In the past I have had issues with docker events not behaving like I expected; the cause turned out to be an mismatch between the clocks on the docker vm and the host machine (see docker/docker#25579 for more). I'm not sure if that is happening on the travis machines, but that is where I would investigate.

johnflavin:
Hey, all green! ✅

davidxia:
Nice. What changed to make it pass?

johnflavin:
Short answer: nothing. I'm pretty sure that event errors like that, where for some reason docker just doesn't give back events, are machine-dependent and not always reproducible. Like I said in [my comment above](https://github.com/spotify/docker-client/pull/549#issuecomment-269655905) it could very well be that the clock on the particular docker vm that they were using is out of sync with the host clock. Docker re-syncs occasionally, but in my experience it doesn't happen reliably and I have to manually sync to make the errors stop.

src/main/java/com/spotify/docker/client/messages/Event.java
johnflavin: It looks like it can be `<String, String>`. I'm not sure why I thought it was possible to get an `Object` there, but their [swagger.yaml](https://github.com/docker/docker/blob/master/api/swagger.yaml#L5336) says `Attributes` will always have `String` values.

I'll fix that.

src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -2248,15 +2250,17 @@ public static EventsParam since(Long since) {
...
-    private static EventsParam filter(String name, String value) {
+    private static EventsParam filter(final String name, final String value) {
       return new EventsFilterParam(name, value);
     }
 
     /**
      * Show only certain events. For example, "event=pull" for image pull events.
      * @param event Type of event to show
      * @return EventsParam
+     * @since API 1.21
```
johnflavin: Yes, I think you are correct. Can fix.

src/main/java/com/spotify/docker/client/DockerClient.java
johnflavin: Let me think this through. I'm not sure exactly what you’re asking to test.

Do you mean we should test whether the `filter=type=<>` param works correctly, and that if we ask for, say, only image events, then that's all we get back? That's doable. A test could do a bunch of stuff—pull an image, create a container, create a volume, etc.—then ask docker for the different types of events one-by-one, and verify that we only get back what we expect.

Or do you mean to test the validity of the `enum` itself, that it has all the values that it needs to have? I don't know how that would be tested. I'm kind of trusting that the API fully documents all the valid values.

davidxia: Sorry, I meant the former.

johnflavin: Ok. That might take a little time, but it shouldn't be too complicated.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1873,10 +1898,149 @@ public void testEventStreamPolling() throws Exception {
...
+    final HostConfig hostConfig = HostConfig.builder()
+            .binds(Bind.from(volume).to(mountPath).build())
+            .build();
+    final ContainerConfig config = ContainerConfig.builder()
+            .image(BUSYBOX_LATEST)
+            .hostConfig(hostConfig)
+            .build();
+
+    // Wait once to clean our event "palette" of events from other tests
+    Thread.sleep(1000);
```
davidxia: This is probably fine for now, but is there a way to check there are no leftover events? Perhaps we can use [awaitility](https://github.com/awaitility/awaitility) in the future here.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1873,10 +1898,149 @@ public void testEventStreamPolling() throws Exception {
...
+    final Date start = new Date();
+    final long startTime = start.getTime() / 1000;
+
+    sut.pull(BUSYBOX_LATEST);
+    sut.createVolume(volume);
+    final ContainerCreation container = sut.createContainer(config, containerName);
+    final String containerId = container.id();
+    sut.startContainer(containerId);
+    Thread.sleep(1000);
```
davidxia: Is this to wait for the container to exit?

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1896,15 +2060,10 @@ private EventStream getImageAndContainerEventStream(final EventsParam... eventsP
...
-                                     final Event createEvent,
-                                     final Event startEvent,
-                                     final String containerId,
-                                     final String imageName) throws Exception {
-    // pull event assertions
+  private void imagePullEventAssertions(final Event pullEvent) throws Exception {
     assertThat(pullEvent.time(), notNullValue());
     if (dockerApiVersionAtLeast("1.22")) {
-      assertEquals("image", pullEvent.type());
+      assertEquals(IMAGE.getName(), pullEvent.type());
```
davidxia: Perhaps make this method take more parameters like `containerEventAssertions()` below so we don't have to hard-code `IMAGE` and `BUSYBOX_LATEST`?

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
johnflavin: I think I could do that. There is a case to avoid: when you are streaming events, if you ask for the next one and there is none, the thread will hang while it waits for more to come over the stream. But I can careful in my tests to only assert that there are no events when I have polled for events and the stream is already closed. Also, setting timeouts on the tests will help.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
johnflavin: That is correct. Writing these tests, I got into problems when I tried to remove a container that was still running. I also tried to stop the container first, but sometimes it was stopped already so that caused its own problems.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
johnflavin: Can do

johnflavin: I left the `IMAGE.getName()` hard-coded. The method is a helper for assertions about image events only, so they will always have that `event.type()`. But I generalized the image name and the type of action.

johnflavin:
I just had a thought based on [your comment](https://github.com/spotify/docker-client/pull/549#discussion_r942881570). Currently the event type enum is in `DockerClient`; it is used as a parameter to filter events to those of certain types. However, the actual `type` member of `Event` just returns a `String`. 

It seems it would make more sense to move `enum Type` from `DockerClient` into `Event`, and change `String type` to `Type type`.

Maybe I'll give that a go and see if it works.

johnflavin:
👆 ✅

johnflavin:
Ugh. I just realized that the best version of the `Events` model object is not just one class with enumerated types. It is a different class for each type, with enumerated `action`s.

johnflavin:
Test failure appears to be a travis hiccup, unrelated to PR.

davidxia:
@johnflavin I've seen similar spurious pull failures. I usually click the "restart job" button for the individual failed jobs to fix them.

johnflavin:
I don't have that button. Must be a Travis feature that only repo owners get.

johnflavin:
Oops. Didn't see the approval before I rebased.

davidxia:
That's OK. Your last commit is straight forward.

mattnworb:
lgtm 👍  thanks!

Issue #530, opened 2016-11-01
LogStream throws EOFException if the stream contains less than 8 bytes


bayanr:
## Description

When trying to read the output of a command launched through `DockerClient.execStart()`, an `EOFException` is thrown if the output of the command is less than 8 characters using `v6.0.0:shaded`

## How to reproduce
The following unit test demonstrates the problem.

``` java
@Test
public void readStdoutOfCommandWithOutputLessThan8Chars() throws DockerCertificateException, DockerException, InterruptedException, IOException {
  DockerClient client = createDockerClient();
  String simpleEcho = client.execCreate("container_id", new String[] { "echo" ,"Hello" }, DockerClient.ExecCreateParam.attachStdout(true), DockerClient.ExecCreateParam.attachStderr(true)).id();

  LogStream logStream = client.execStart(simpleEcho, DockerClient.ExecStartParameter.TTY);
  ByteArrayOutputStream stdOut = new ByteArrayOutputStream();
  ByteArrayOutputStream stdErr = new ByteArrayOutputStream();

  logStream.attach(stdOut, stdErr);
  System.out.println("======== OUTPUT =======");
  System.out.println(stdOut);
}
```

Produces the following trace:

```
java.lang.RuntimeException: java.io.EOFException
    at com.google.common.base.Throwables.propagate(Throwables.java:160)
    at com.spotify.docker.client.LogStream.computeNext(LogStream.java:61)
    at com.spotify.docker.client.LogStream.computeNext(LogStream.java:35)
    at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:143)
    at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:138)
    at com.spotify.docker.client.LogStream.attach(LogStream.java:142)
    at Test.readStdoutOfCommandWithOutputLessThan8Chars(Test.java:61)
    // JUnit noise omitted
Caused by: java.io.EOFException
    at com.spotify.docker.client.LogReader.nextMessage(LogReader.java:59)
    at com.spotify.docker.client.LogStream.computeNext(LogStream.java:59)
    ... 31 more
```

Whereas the following test

``` java
@Test
public void readStdoutOfCommandWithOutputMoreThan8Chars() throws DockerCertificateException, DockerException, InterruptedException, IOException {
  DockerClient client = createDockerClient();
[...]

PR #551, opened 2016-12-09
Handle case when TTY is on and output less than 8 bytes. (#530)

rgrunber:
There is no need to throw an EOFException when TTY is enabled and the
output is less than 8 bytes as this is a valid case. Continue reading
the header information and decide what kind of stream it is.

I haven't created a separate test for this since one can easily reproduce by taking DefaultDockerClientTest#testLogsTty() and changing '.cmd("sh", "-c", "ls")' to '.cmd("echo", "123")' .

Note that calling header.get(), header.getInt(0) is still safe in this case because our header byte buffer will always have a size of 8 (since the underlying byte array has length 8)

### DIFF
--- a/src/main/java/com/spotify/docker/client/LogReader.java
+++ b/src/main/java/com/spotify/docker/client/LogReader.java
@@ -25,7 +25,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.Closeable;
-import java.io.EOFException;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


rgrunber:
Failing only on Docker 1.11.2 but the issues seem unrelated (network unreachable, and a few errors on docker pull)

mattnworb:
I re-ran the build and the test failure on 1.11.2 indeed looks unrelated

PR #553, opened 2016-12-16
Small code style cleanup

davidxia:


src/main/java/com/spotify/docker/client/messages/AuthRegistryConfig.java
```diff
@@ -26,6 +26,7 @@
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import jersey.repackaged.com.google.common.collect.ImmutableMap;
```
mattnworb: did you mean to import this repackaged class?

davidxia: Oops. Thanks.

Issue #419, opened 2016-04-23
Hanging on exiting Java process


pnathan:
Hi, I have a sample program that's hanging on exit. I have a notion there's not closing properly, but I don't _think_ so

```
package docker;
import com.spotify.docker.client.exceptions.DockerException;
import com.spotify.docker.client.messages.Container;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import com.spotify.docker.client.*;

import java.util.List;

public class Example {
   private static final Logger LOGGER = LoggerFactory.getLogger(MuxDemux.class);
   public static void main(String[] args) throws DockerException, InterruptedException {
      final DockerClient docker = new DefaultDockerClient("unix:///var/run/docker.sock");

      final List<Container> containers = docker.listContainers();

      LOGGER.info("Found " + Integer.toString(containers.size())) ;
      for (int i = 0; i < containers.size(); i++) {

         LOGGER.info("Detected container id: " + containers.get(i).id());
      }


     docker.close();
     LOGGER.info("Going home now") ;
   }
}
```

This code gets the container IDs, prints them, logs "Going home now", and then the program hangs; I have to control-C it.

My expectation is that docker.close() appropriately closes all waiting connections.

Two questions: (1) should this work (i.e., is this PEBKAC (always a possibility. :-) ), and if not, (2) what should I do to fix it?
- docker version: docker 1.10.3 on Debian
- docker-client version: 4.0.2
- Java 1.7

plamentotev:
if it hangs for about a minute then this is the same issue as #405

davidxia:
Should be fixed by #558.

Issue #405, opened 2016-03-30
docker.pull keeps JVM alive for a long time


RichoDemus:
## Description

After having run docker.pull() and docker.close, a thread called jersey-client-async-executor-0 is left alive for quite some time, preventing the JVM from terminating
## How to reproduce

I have created a small repo to reproduce this: https://github.com/RichoDemus/docker-pull-bug  
but I'm basically just doing

```
final DockerClient docker = DefaultDockerClient.fromEnv().build();  
docker.pull("alpine:latest");  
docker.close();  
```
## What do you expect

I expect that after doing docker.close and exiting the main method, the program should terminate right away
## What happened instead

The process stays alive for about a minute extra
## Software:
- docker version: 1.10.3
- docker-client version: 3.6.6
## Full backtrace

N/A

johnflavin:
This could be related to a Jersey issue ([2878](https://java.net/jira/browse/JERSEY-2878)). Jersey was keeping connections with an `InputStream` alive after their `close()` methods were called. It was reportedly fixed in version 2.21; `docker-client` uses 2.19.

But I don’t know enough about what is going on in the `pull()` method to know if this Jersey issue applies. Nor do I know what JUnit is doing differently to quickly close the connection during the test. :confused:

plamentotev:
I get this behavior will all commands not only pull. For example with the following code:
```java
        try(DockerClient dockerClient = DefaultDockerClient.builder()
                .uri("http://localhost:2375/")
                .build()) {
            System.out.println(dockerClient.version());
        }
``` 
the program hangs for about a minute after the version is printed to the console. Of course the reason is the connection pool used  by jersey. It uses non-daemon threads so this keeps the application alive. Still not sure why there are alive threads in the pool after the `dockerClient` is closed.

I've updated the jersey version to 2.21 and I don't see any difference too.

plamentotev:
I found it. The offender is [JERSEY-2969 - Client#close() not shutting down Async Executor](https://java.net/jira/browse/JERSEY-2969). it's fixed in 2.22.2. updating jersey dependency to 2.22.2 solved the problem.

nfleury:
@plamentotev  I was facing the same issue. Switching jersey-client to 2.22.2 fixed the issue. Would be nice to update this lib version ?

tobiaskohlbau:
Today I ran into the same issue. Are there any plans on updating jersey-client to 2.22.2 in the mainline version?
Would be glad if I maven version fixes this issue.

[...]

PR #558, opened 2016-12-23
upgrade jersey-client to 2.22.2

mattnworb:
addresess #405, #419


davidxia:
Would be good to verify with https://github.com/RichoDemus/docker-pull-bug.

davidxia:
I tested with https://github.com/RichoDemus/docker-pull-bug, and it works.

PR #561, opened 2016-12-24
Use ObjectMapperProvider in all classes

davidxia:


davidxia:
I noticed the same thing. I don't think we do anything with the context, so I'll change this.

PR #572, opened 2016-12-31
Do not allow attaching to a non-running container

### LINKED ISSUE #288: docker-client should be smarter about attaching to a stopped container
Ever since we use a no timeout client for attaching to a container (see https://github.com/spotify/docker-client/commit/b76fc6387289b1d310a22501e033c07c5895b7ad), if you attach to a stopped container, the docker-client just hangs forever. How can we be smart about this?

davidxia:
This is how the Docker CLI behaves. Attaching to a non-running container
hangs indefinitely. Fixes #288

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -174,6 +174,12 @@
       <version>2.8.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <version>2.0.0</version>
+      <scope>test</scope>
+    </dependency>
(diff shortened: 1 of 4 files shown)


davidxia:
cc @rgrunber

rgrunber:
This looks fine to me, so 👍 . I would only add that similar logic may need to be added in place for execCreate/execStart, but it'd be fine to do all that in a separate issue/pull request.

davidxia:
Thanks. I've updated the javadoc. I made a similar one for `execCreate` here https://github.com/spotify/docker-client/pull/574.

PR #574, opened 2017-01-03
Prevent exec start on a non-running container

davidxia:
Currently, trying to do run `execStart()` will make docker-client throw
a `DockerRequestException` with a generic 500 error message.

Instead have docker-client check if the container is running and throw
an `IllegalStateException` otherwise.


davidxia:
I didn't do anything for `execCreate` since that method doesn't know about the container ID for which the exec was created. So even though you can't start an exec for a container that's not running, I think the way it works right now (`ExecNotFoundException` is thrown) is OK.

PR #577, opened 2017-01-04
Inspect healthcheck

CGNonofr:


davidxia:
@CGNonofr Thanks for your PR. Which version(s) of Docker's API support this? Could you also add a test to `DefaultDockerClientTest` that checks the docker daemon actually provides this info?

CGNonofr:
This feature was added in v1.24. Unfortunately, you have to create an image with a Dockerfile containing the `HEALTHCHECK [...] CMD [...]` syntax. Since this api doesn't allow to create an image from a Dockerfile, I see no way to write this test.

davidxia:
Is this documented anywhere in the docker API docs?

CGNonofr:
I don't think.. it's mainly used by docker-compose but I don't found any related documentation.

johnflavin:
You can use `docker-client` to build an image from a Dockerfile. You need to create a test resource directory containing your Dockerfile, and use the `DockerClient.build(...)` method. See `DefaultDockerClientTest.testBuildImageIdWithBuildargs()` for a test that does this.

johnflavin:
`HEALTHCHECK` is in the Dockerfile documentation. https://docs.docker.com/engine/reference/builder/

CGNonofr:
Thanks!
Test added.
(I meant in https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/inspect-a-container)

CGNonofr:
The code coverage test is failing but I don't know what I should do to fix it!

davidxia:
rebased here https://github.com/spotify/docker-client/pull/583

PR #582, opened 2017-01-06
replace pojo boilerplate with AutoValue

davidxia:


davidxia:
@mattnworb @gimaker Going to merge this by EOD if no objections.

PR #585, opened 2017-01-12
Inspect healthcheck

davidxia:


davidxia:
Let me also add a test.

PR #586, opened 2017-01-13
attempt to include HTTP error bodies in exception

mattnworb:
When a HTTP error (4xx, 5xx, IOException) occurs in communciating with
the Docker Remote API, DefaultDockerClient attempts to read the response
body (if one exists) and passes it to the DockerRequestException which
is thrown.

The message field that is passed to the DockerRequestException is unused
in the "message" passed up to the superclass (which is available via
`getMessage()` and `Exception.toString()`).

This change would include the "message" passed to DockerRequestException
in the superclass's message if possible, to help improve troubleshooting
and debugging.

Also remove the five unused constructors in DockerRequestException.

### DIFF
--- a/src/main/java/com/spotify/docker/client/exceptions/DockerRequestException.java
+++ b/src/main/java/com/spotify/docker/client/exceptions/DockerRequestException.java
@@ -21,24 +21,25 @@
 package com.spotify.docker.client.exceptions;
 
 import java.net.URI;
+import javax.annotation.Nullable;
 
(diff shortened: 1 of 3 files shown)


davidxia:
travis seemed to have gotten stuck

mattnworb:
@davidxia test added. I am not going to test that the DefaultDockerClient passes the proper field to the Exception - I don't think that can be done - but rather that it is included in the Exception.getMessage()

PR #595, opened 2017-01-20
Adding Equals and Hash code to Param objects

Hazz223:
I've added the Equals and Hash code to these objects, as they currently fail when checking for equality. This became apparent when trying to mock them in Testing. 

On master this test fails:
` assertEquals(DockerClient.ListImagesParam.allImages(), DockerClient.ListImagesParam.allImages()); `
This test now passes. 

Question; does this require unit tests? Also, would you like this added to the Change log?


johnflavin:
I hadn't really noticed how all these classes had proliferated. Seems like it might make sense to make an abstract `Param` class with `name` and `value` + `equals` and `hashCode`, then change all these to concrete `Param` types with their unique methods in them.

johnflavin:
And I wouldn't think unit tests of these are necessary. They already have decent test coverage by virtue of being used in various ways to test the API functions.

Hazz223:
Sure, i can convert these to an abstract class. 
Any idea why that might have failed? I've got a few failing tests on my machine, but i think that's more due to environment than the actual code.

Hazz223:
I'm aware some this will currently fail due to the Header checker. I'm not familier with this plugin, and not sure what I've done wrong. Will investigate and find out what's wrong with the header for the Param abstract class

johnflavin:
I'm guessing that it's just the block of comments at the top of each file has to be correct. But I'm not sure either.

I wouldnt make a whole new file just for this one class. I would make a static inner class of DockerClient, similar to all the existing `*Params` classes.

johnflavin:
You don't have to push up to your branch to run the tests. Before it executes the tests, Travis is running `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V` (or, at least, that's what it says in the logs). You could just run that locally, if you've got your docker all set up properly. Or there is probably some better way to run the check-file-headers plugging directly.

Hazz223:
Actually, the pushes were more so I could continue this else where and not leave it on my laptop for the weekend ha! 

Actually been wondering about the structure of the DockerClient class. Is there a reason it's one file? Any reason why this isn't broken down?

Hazz223:
Pull request fails the coverage check as it adds the equals and hascode. I'm happy to add in tests, but I'm not sure how much gain will come from them. Advice?

mattnworb:
and thanks @Hazz223 for the contribution!

PR #597, opened 2017-01-24
Allow specifying tar InputStream when copying files to container

plamentotev:
Currently if you want to copy files to container the only option
is to provide a directory that will be archived and uploaded.

In some cases the files are already archived in tar format.
In those cases it's beneficial to be be able to provide InputStream instead.

p.s. I'm modifying the API so I guess I have to bump the version too, but not sure what the versioning policy is. What the version should be - 7.1.0?


mattnworb:
also: thanks for the contribution!

PR #600, opened 2017-01-26
Fix for issue 552

krizsan:


krizsan:
Builds fails in DefaultDockerClientTest, code that I have not modified in this pull request.

krizsan:
I'm giving up after having spent more time trying to fix the build here than it took to fix the actual issue.

krizsan:
Fixes in PR 612 fixed the failing tests! Thanks a lot for that!

src/main/java/com/spotify/docker/client/LogReader.java
```diff
@@ -77,4 +91,49 @@ public void close() throws IOException {
...
+        int count = LogReader.this.read(in, buff, off, len);
+        return count;
+      }
+    });
+
+    try {
+      int readCountInt = readCount.get(readTimeoutInMilliseconds, TimeUnit.MILLISECONDS);
+      return readCountInt;
+    } catch (Exception theException) {
+      System.currentTimeMillis();
```
mattnworb: what is the point of this `catch`?

src/main/java/com/spotify/docker/client/LogReader.java
```diff
@@ -77,4 +91,49 @@ public void close() throws IOException {
...
+    }
+
+    readStreamExecutorService.shutdown();
+    readStreamExecutorService = Executors.newSingleThreadExecutor();
+    return 0;
```
mattnworb: why shutdown and re-initialize the executor in case of exception, and then return 0?

src/main/java/com/spotify/docker/client/LogReader.java
```diff
@@ -77,4 +91,49 @@ public void close() throws IOException {
     // where close is a no-op.
     copy(stream, nullOutputStream());
   }
+
+  protected synchronized int readWithTimeout(final InputStream in, final byte[] buff,
```
mattnworb: why make this synchronized? because you are re-assigning the readStreamExecutorService field?

src/test/java/com/spotify/docker/it/LogReaderHangTest.java
```diff
@@ -0,0 +1,142 @@
...
+   * Performs setup before one test method.
+   *
+   * @throws Exception If error occurred creating test Docker image.
+   */
+  @Before
+  public void setupBeforeOneTest() throws Exception {
+    createDockerLogReaderHangTestImage();
+
+    ivanDockerContainer =
+        new IvanDockerContainer(dockerClient, LOGHANG_DOCKERIMAGE);
```
mattnworb: couldn't all of the logic in the IvanDockerContainer class be in this test class instead? doesn't seem worthwhile to put that logic in its own class if it is only used in one test.

mattnworb: It is also not the most obvious of names for what the class is doing

krizsan:
I agree that it would be better to handle timeouts at the HTTP client / socket level.

PR #602, opened 2017-01-29
Adding NetworkConnection abstraction to allow aliases & more parameters on connection

loicortola:
To this day, only **connectToNetwork(String containerId, String networkId)** was supported to add a network.
This did not allow to add additional configuration (specifically in my case, provide **aliases** when connecting a container to a network).

This use-case could not be solved unless [EndpointConfig (see API Docs)](https://docs.docker.com/engine/api/v1.25/) support was added.

This pull request adds support for EndpointConfig while keeping existing APIs.

NB: An additional method was added to the DockerClient interface with the following signature:
```java
  /**
   * Connects a docker container to a network, with extended configuration.
   * This is useful when you want to set specific details (aliases, gateway, etc...)
   * for your container inside the network.
   *
   * @param networkId          The id of the network to connect.
   * @param networkConnection  The target connection parameters.
   * @throws NotFoundException          if either container or network is not found (404)
   * @throws DockerException            if a server error occurred (500)
   * @throws InterruptedException       if the thread is interrupted
   */
  void connectToNetwork(String networkId, NetworkConnection networkConnection)
    throws DockerException, InterruptedException;
```

The DefaultDockerClient has been implemented, one test was added in  **src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java** to make sure the use-case works out well.

Would you be able to merge this PR?

Thanks everyone for the awesome work behind this API. It's evolved to become the best java API for docker with great improvements from what it was two years ago ;)

Best wishes!

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -93,6 +93,7 @@ import com.spotify.docker.client.messages.ImageSearchResult;
 import com.spotify.docker.client.messages.Info;
 import com.spotify.docker.client.messages.Network;
 import com.spotify.docker.client.messages.NetworkConfig;
+import com.spotify.docker.client.messages.NetworkConnection;
 import com.spotify.docker.client.messages.NetworkCreation;
 import com.spotify.docker.client.messages.ProgressMessage;
 import com.spotify.docker.client.messages.RegistryAuth;
@@ -159,7 +160,6 @@ import org.glassfish.jersey.internal.util.Base64;
 import org.glassfish.jersey.jackson.JacksonFeature;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 
 public class DefaultDockerClient implements DockerClient, Closeable {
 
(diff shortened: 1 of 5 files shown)


loicortola:
Btw: checked the tests, the failing tests have nothing to do with the actual PR

loicortola:
Up, anyone around?

loicortola:
@mattnworb any thoughts / update on this?

loicortola:
see #609 to solve test issues.
Will wait for #609 to be merged before merging this one.

src/main/java/com/spotify/docker/client/messages/AttachedNetwork.java
```diff
@@ -27,14 +27,20 @@
...
 import javax.annotation.Nullable;
 
+
 @AutoValue
 @JsonAutoDetect(fieldVisibility = ANY, getterVisibility = NONE, setterVisibility = NONE)
 public abstract class AttachedNetwork {
 
   @Nullable
+  @JsonProperty("Aliases")
+  public abstract ImmutableList<String> aliases();
```
mattnworb: is there a difference between a null list of aliases and an empty list? Meaning does the Docker Remote API treat the two differently?

loicortola: I don't think so. 
Actually, I don't think aliases are ever null when supported by the API (at least the container id).
For retrocompatibility purposes, I put it as nullable so that it doesn't break for earlier versions.

loicortola:
@mattnworb ready to merge?

mattnworb:
Thanks again for the contribution!

PR #616, opened 2017-02-06
Support listing services by label (fixes #615)

### LINKED ISSUE #615: Support filtering services by label
```DockerClient#listServices``` does not support filtering by labels

### PR DESCRIPTION
**Pull request by dmandalidis**

### CODE REVIEW

File: src/main/java/com/spotify/docker/client/messages/swarm/Service.java [hunk 1]
```diff
@@ -71,10 +71,37 @@
     @Nullable
     public abstract String serviceName();
 
+    /**
+     * Filter by label.
+     */
+    @Nullable
+    public abstract Label label();
```
johnflavin: Can a service really only have one label? All the other objects that have labels can have an arbitrary number of them.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1684,7 +1683,13 @@ public void updateService(final String serviceId, final Long version, final Serv
     if (criteria.serviceName() != null) {
       filters.put("name", Collections.singletonList(criteria.serviceName()));
     }
+    if (criteria.label() != null) {
+      filters.put("label", Collections.singletonList(
+          String.format("%s=%s", criteria.label().name(), criteria.label().value())
```
johnflavin: Usually with the docker APIs, you can filter on labels as "name=value" but also just "name" with any value. Is that not the case with services?

src/main/java/com/spotify/docker/client/messages/swarm/Service.java
```diff
@@ -71,10 +71,37 @@
...
+      public abstract String name();
+      
+      public abstract String value();
+      
+      public static Builder builder() {
+        return new AutoValue_Service_Criteria_Label.Builder();
+      }
+      
+      @AutoValue.Builder
+      public abstract static class Builder {
```
johnflavin: Does Label really need its own Builder? Seems like it would be easier to just have a static method `Service.Builder.label(String, String)` that makes a new Label and adds it to the Service Builder. Maybe also a second one that takes only one String if you just want a Label with name and no value.

src/main/java/com/spotify/docker/client/messages/swarm/Service.java
dmandalidis: fixed

src/main/java/com/spotify/docker/client/messages/swarm/Service.java
dmandalidis: from a first glance it doesn't seem so trivial to address it (ImmutableMap doesn't take null values)

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
dmandalidis: fixed

src/main/java/com/spotify/docker/client/messages/swarm/Service.java
```diff
@@ -71,6 +74,12 @@
     @Nullable
     public abstract String serviceName();
 
+    /**
+     * Filter by label.
+     */
+    @Nullable
+    public abstract ImmutableMap<String, String> labels();
```
mattnworb: this contains the labels associated with the Service, right? That were created when the service was created? i.e. this isn't the labels used in a "list services" filter

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4210,6 +4210,26 @@ public void testListServicesFilterByName() throws Exception {
...
+    
+    final ServiceSpec spec = createServiceSpec(serviceName, labels);
+    sut.createService(spec);
+
+    final List<Service> services = sut.listServices(Service.find().addLabel("foo", "bar").build());
+    
+    assertThat(services.size(), is(1));
+    assertThat(services.get(0).spec().labels().get("foo"), is("bar"));
+    
+    final List<Service> notFoundServices = sut.listServices(Service.find().addLabel("bar", "foo").build());
```
mattnworb: checkstyle seems to be complaining that this line is too long, can you fix?

dmandalidis: fixed

PR #620, opened 2017-02-08
Make readOnly mount option optional (fixes #618)

dmandalidis:


davidxia:
Thanks. Fixed checkstyle error here https://github.com/spotify/docker-client/pull/633

PR #623, opened 2017-02-09
Fix 622: create Healthcheck message class

johnflavin:


mattnworb:
something is screwy with the one travis build that fails, but the others look ok ✅

PR #643, opened 2017-02-22
Add "Stop a container" section to user manual

giftig:
The heading existed in the TOC but the code example was missing. I've
filled in a sensible example from what I gleaned from reading the
source.

### DIFF
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -189,6 +189,12 @@ docker.resizeTty("containerID", height, width);
 docker.startContainer("containerID");
 ```
 
+### Stop a container
+
+```java
+docker.stopContainer("containerID", 10); // kill after 10 seconds
+```
+
 ### Restart a container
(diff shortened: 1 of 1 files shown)


docs/user_manual.md
```diff
@@ -189,6 +189,12 @@ docker.resizeTty("containerID", height, width);
 docker.startContainer("containerID");
 ```
 
+### Stop a container
+
+```java
+docker.stopContainer("", 10); // kill after 10 seconds
```
davidxia: Instead of empty string, we should indicate the first param needs to be a container ID.

giftig: That's exactly what I meant to do but I was focusing on the parameter I wasn't sure of first and must've gotten. Sorry, my bad, will fix.

giftig:
@davidxia thanks, fixed.

PR #646, opened 2017-02-22
Make travis test on Docker 1.13.1 as well

davidxia:


mattnworb:
👍  i think it might make sense to start phasing out older versions as well

.travis.sh
```diff
@@ -83,7 +83,7 @@ end script
       sudo restart docker
     fi
 
-    if [[ "$DOCKER_VERSION" =~ ^1\.12\..* ]]; then
```
mattnworb: might be more maintainable long-term to add this as a param to the matrix list in travis.yml, like `ENABLE_SWARM=1`

PR #660, opened 2017-03-03
Fix container name regex

### LINKED ISSUE #658: Why there is a restriction for the container name?
In file `docker-client/src/main/java/com/spotify/docker/client/DefaultDockerClient.java` there is a restriction for the container name.

    private static final Pattern CONTAINER_NAME_PATTERN = Pattern.compile("/?[a-zA-Z0-9_-]+");

For example, in docker envrionment, the `.` is allowed for the container name. e.g `a.b.c-123` is a valid container name in docker.

johnflavin:
Bring our container name regex into alignment with docker's, as read from their code [here](https://github.com/docker/docker/blob/dba271a42ab4841dbcf2e953491e9ee728cd8e16/api/names.go#L6). Fixes #658.

Changes:

* Change name matcher from "one or more characters in the set `[a-zA-Z0-9_-]`" to "one character in the set `[a-zA-Z0-9]` followed by one or more characters in the set `[a-zA-Z0-9_.-]`". 
* Remove the optional `/` at the start of the name. I think this may have been added for testing purposes, as the name returned by docker when you inspect the container does have a leading `/`. However, it doesn't make sense to include it here. We have better ways to check that condition in tests, and `/` is not a valid character when naming a container.
* Enforce that these characters are the whole name by adding the anchors `^...$`. I am pretty sure this will not actually change anything about the match—i.e. if something would have matched or not without these anchors, it still will match or not—since the `java.util.regex.Matcher#match` method we use is already anchored by default. But, still, I feel that it's good to make the anchors obvious in case anyone checks. And it is consistent with docker's regex linked above, which is explicitly anchored on both ends.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -226,7 +226,8 @@ public class DefaultDockerClient implements DockerClient, Closeable {
       LogsResponseReader.class,
       ProgressResponseReader.class);
 
-  private static final Pattern CONTAINER_NAME_PATTERN = Pattern.compile("/?[a-zA-Z0-9_-]+");
+  private static final Pattern CONTAINER_NAME_PATTERN =
+          Pattern.compile("^[a-zA-Z0-9][a-zA-Z0-9_.-]+$");
 
   private static final GenericType<List<Container>> CONTAINER_LIST =
       new GenericType<List<Container>>() {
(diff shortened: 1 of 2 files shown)


davidxia:
Thanks. Perhaps add a test for the new regex?

johnflavin:
I added a test with a few assertions. It isn't exactly comprehensive, since there is a huge set of names that match and a practically infinite set of names that do not. So I just threw in a few that I thought would be demonstrative.

PR #663, opened 2017-03-05
Create secret support (fixes #662)

### LINKED ISSUE #662: Support for create secret

dmandalidis:


dmandalidis:
@johnflavin it is possible to add a secret to a swarm service, but the way you do it is undocumented (see https://github.com/docker/docker/issues/31179). The way it works is still unclear to me, so I assumed that it would take some time to block this PR from merging. I 'll implement that part of the API as well.

davidxia:
@dmandalidis Thanks for your PR. Could you also add a section to the user_manual.md?

davidxia:
We are waiting on docker to provide clearer instructions on how to use secrets once they're created right?

dmandalidis:
@davidxia Nope, this PR is for creating secrets, use secrets is another PR which I 'll work as well

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2069,6 +2071,28 @@ public VolumeList listVolumes(ListVolumesParam... params)
...
+
+    try {
+      return request(POST, SecretCreateResponse.class, resource,
+                     resource.request(APPLICATION_JSON_TYPE),
+                     Entity.json(secret));
+    } catch (DockerRequestException dre) {
+      switch (dre.status()) {
+        case 409:
+          throw new DockerException("Name conflicts with an existing object.", dre);
+        case 503:
```
davidxia: Is this correct? According to the [API docs](https://docs.docker.com/engine/api/v1.25/#operation/SecretCreate), this should be 406, but the docs have been known to be wrong.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4257,6 +4260,18 @@ public void testCreateService() throws Exception {
     final ServiceCreateResponse response = sut.createService(spec);
     assertThat(response.id(), is(notNullValue()));
   }
+  
+  @Test
+  public void testCreateSecret() throws Exception {
+    requireDockerApiVersionAtLeast("1.25", "swarm support");
+    
+    final String secretData = Base64.encodeAsString("testdata".getBytes());
```
davidxia: Could we also test what happens when the data isn't Base64-url-safe-encoded, with labels, collision with an existing secret, etc?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
johnflavin: The docs you linked say that both 406 and 409 are possible error responses.

davidxia: Oh, I meant that for the error when server isn't part of a swarm, the docs say the code will be 406 instead of 503 here.

davidxia: And I don't see a 503 but a general 500.

johnflavin: Ah, sorry. I misunderstood. I blame the GitHub UI which shows all the code above the line note but nothing after, which is kinda the opposite of what you would probably want.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
davidxia: Thanks

davidxia:
In the process of fixing up this PR, I ended up adding support for all secret ops. https://github.com/spotify/docker-client/pull/685

PR #667, opened 2017-03-06
MemoryStats compatible with Docker 1.13.x

icodethereforeiam:


davidxia:
@icodethereforeiam Thanks. Could you write a test for this as well or at least describe how to repro?

PR #674, opened 2017-03-13
Add builder for Device

### LINKED ISSUE #673: No builder for Device
## Description
Currently I am using v6.2.5 but was looking to upgrade to the most recent versions.  However it seems with the change to use AutoValue it's no longer possible to construct a Device? 

I currently use `HostConfig.builder().devices(new Device(...), new Device(...))` to run a container with an attached device.  With the AutoValue'd version Device is abstract so can't be initialized, and the AutoValue_Device isn't accessible.  Looks like a builder needs to be setup as well but I wasn't sure if this was intentional and in new versions should be adding devices some other way. 

https://github.com/spotify/docker-client/blob/master/src/main/java/com/spotify/docker/client/messages/Device.java
https://github.com/google/auto/blob/master/value/userguide/builders.md

davidxia:
fixes #673

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/Device.java
+++ b/src/main/java/com/spotify/docker/client/messages/Device.java
@@ -28,24 +28,49 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 
+import javax.annotation.Nullable;
+
 @AutoValue
 @JsonAutoDetect(fieldVisibility = ANY, getterVisibility = NONE, setterVisibility = NONE)
 public abstract class Device {
 
+  @Nullable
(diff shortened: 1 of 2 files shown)


src/main/java/com/spotify/docker/client/messages/Device.java
```diff
@@ -28,24 +28,49 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 
+import javax.annotation.Nullable;
+
 @AutoValue
 @JsonAutoDetect(fieldVisibility = ANY, getterVisibility = NONE, setterVisibility = NONE)
 public abstract class Device {
 
+  @Nullable
```
davidxia: dockerd didn't complain when I didn't set this or `pathInContainer`. I'm not sure if it makes sense for these to be nullable though. Should we enforce in code?

PR #677, opened 2017-03-14
Test on v17.03.0-ce

johnflavin:
Test on latest docker release.


davidxia:
Perhaps we can just drop testing on 1.8.3 which was released 2015-10-12? My motivation for not testing on too many versions is that travis is really flaky and often fails to pull images or, less frequently, install docker. The fewer versions we test on, the less likely a PR is to fail.

johnflavin:
Makes sense.

davidxia:
Want to rebase on master? I saw a failure that https://github.com/spotify/docker-client/pull/681 fixes.

johnflavin:
This has already been rebased on master with #681. So maybe we do need #666.

PR #678, opened 2017-03-14
Add PKCS8 and multi CA support

shiroari:
Hi,

This PR adds support for PKCS#8 private keys in key.pem file. Also I replaced `generateCertificate` call with `generateCertificates` to read all certificates from ca.pem file in order to support multiple certificate authority. In order to make testing simpler I introduced ssl factory interface and moved ssl test that can be run without docker to DockerCertificatesTest.
This PR is important when one uses external tools to manage and rotate certificates (e.g. ipa).


src/test/java/com/spotify/docker/client/DockerCertificatesTest.java
```diff
@@ -0,0 +1,149 @@
...
+    verify(factory).newSslContext(keyStore.capture(), password.capture(), trustStore.capture());
+
+    final KeyStore.PrivateKeyEntry pkEntry = (KeyStore.PrivateKeyEntry) keyStore.getValue()
+        .getEntry("key", new KeyStore.PasswordProtection(password.getValue()));
+
+    assertNotNull(pkEntry.getPrivateKey());
+  }
+
+  @Test
+  public void loadReadPrivateKeyPkcs8() throws Exception {
```
davidxia: minor: did you mean `testReadPrivateKeyPkcs8` instead of `loadReadPrivateKeyPkcs8`?

davidxia:
cc @rohansingh @mattnworb Since I'm not very familiar with these certs.

src/test/java/com/spotify/docker/client/DockerCertificatesTest.java
shiroari: Yes, you right. It is wrong here. I can replace it, but I would like to know how do you prefer committing the fix. I personally don't like "dangling" commits with oneline fixes and I usually squash it and use "force", but then I think all approvals will disappear and you will have to review whole commit again.

davidxia: I'm fine with either approach. I'll make sure your commits are nice and tidy :)

shiroari:
Fixed comment from @davidxia

davidxia:
checkstyle :(

shiroari:
Well, I fixed checkstyle, but one build failed. Is that something with stability?

PR #681, opened 2017-03-15
Fix autoremove test

johnflavin:
I have seen a lot of failed tests on 1.13 from `testAutoRemoveWhenSetToTrue`. I fixed that test.

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -4170,7 +4170,7 @@ public class DefaultDockerClientTest {
     assertThat(info.hostConfig().storageOpt().get("size"), is("20G"));
   }
 
-  @Test(expected = ContainerNotFoundException.class)
+  @Test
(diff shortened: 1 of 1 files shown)


davidxia:
Thanks. This was annoying. Just curious how this fixes it though.

johnflavin:
I think the problem was this line:
```
assertThat(info.state().running(), equalTo(true));
```
The container wasn't created with any long-running command, so there was no guarantee that should be running when it was inspected. This also wasn't even necessary to assert, because that wasn't the purpose of the test. So I just removed that line.

The next problem, which just made it less obvious what was going on and harder to debug, is that the test was annotated with `@Test(expected = ContainerNotFoundException.class)`. That meant that, since the test failed with an assertion error, not with the expected exception, the message we got in the logs was `java.lang.AssertionError: Expected exception: com.spotify.docker.client.exceptions.ContainerNotFoundException`. The real cause of the failure was being masked by the exception annotation. So I changed that to use the `@Rule ExpectedException`. Now, the exception should only be thrown by that one line where we expect it to be thrown.

johnflavin:
Once this gets merged, I'll rebase my other open PRs on it.

PR #685, opened 2017-03-16
Support all secret operations (fixes #662)

### LINKED ISSUE #662: Support for create secret

davidxia:


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4279,6 +4283,79 @@ public void testCreateService() throws Exception {
...
+    final SecretSpec secretSpecConflict = SecretSpec.builder()
+        .name("asecret")
+        .data(secretData)
+        .labels(labels)
+        .build();
+
+    try {
+      sut.createSecret(secretSpecConflict);
+      fail("Should fail due to secret name conflict");
+    } catch (DockerException ex) {
```
johnflavin: This should be `ConflictException`

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4279,6 +4283,79 @@ public void testCreateService() throws Exception {
...
+    assertThat(secrets.size(), equalTo(1));
+    assertThat(secrets, hasItem(secret));
+
+    sut.deleteSecret(secretId);
+    assertThat(sut.listSecrets().size(), equalTo(0));
+
+    try {
+      sut.inspectSecret(secretId);
+      fail("Should fail because of non-existant secret ID");
+    } catch (DockerException ex) {
```
johnflavin: Should be `NotFoundException`

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4279,6 +4283,79 @@ public void testCreateService() throws Exception {
...
+      // Ignored
+    }
+
+    try {
+      sut.deleteSecret(secretId);
+      fail("Should fail because of non-existant secret ID");
+    } catch (DockerException ex) {
```
johnflavin: Should be `NotFoundException`

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2069,6 +2074,71 @@ public VolumeList listVolumes(ListVolumesParam... params)
...
+    final WebTarget resource = resource().path("secrets").path(secretId);
+
+    try {
+      return request(GET, Secret.class, resource, resource.request(APPLICATION_JSON_TYPE));
+    } catch (final DockerRequestException ex) {
+      switch (ex.status()) {
+        case 404:
+          throw new NotFoundException("Secret " + secretId + " not found.", ex);
+        case 406:
+          throw new DockerException("Server not part of swarm.", ex);
```
johnflavin: Should this be a new `Exception` class?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2069,6 +2074,71 @@ public VolumeList listVolumes(ListVolumesParam... params)
...
+    final WebTarget resource = resource().path("secrets").path("create");
+
+    try {
+      return request(POST, SecretCreateResponse.class, resource,
+                     resource.request(APPLICATION_JSON_TYPE),
+                     Entity.json(secret));
+    } catch (final DockerRequestException ex) {
+      switch (ex.status()) {
+        case 406:
+          throw new DockerException("Server not part of swarm.", ex);
```
johnflavin: Should this be a new `Exception` class?

davidxia:
Thanks for reviewing.

> Instead of checking for Exceptions in try/catch blocks, use the ExpectedException exception instance.
I can't since I want to test for multiple thrown exceptions in the test method. Once an exception is caught by the `ExpectedException` rule, the test ends. :(

> Check for specific Exception classes, not the general DockerException.
👍 
> Possibly make the 406 error its own Exception class.
👍

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
davidxia: What should we call it? `NonSwarmNodeException`?

davidxia:
updated

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4279,6 +4284,79 @@ public void testCreateService() throws Exception {
...
+    final SecretSpec secretSpecConflict = SecretSpec.builder()
+        .name("asecret")
+        .data(secretData)
+        .labels(labels)
+        .build();
+
+    try {
+      sut.createSecret(secretSpecConflict);
+      fail("Should fail due to secret name conflict");
+    } catch (ConflictException ex) {
```
johnflavin: You’re catching the "correct" exception here according to the spec. But your tests are failing from a docker bug: docker/docker#31909. You'll need to code around it for version 1.13.1 and 17.03.0-ce. 

Adding an additional `catch` clause here for `DockerException` that verifies the docker version is one of these two should be enough.

src/main/java/com/spotify/docker/client/messages/swarm/SecretSpec.java
```diff
@@ -0,0 +1,78 @@
...
+  @JsonCreator
+  static SecretSpec create(
+      @JsonProperty("Name") String name,
+      @JsonProperty("Labels") Map<String, String> labels,
+      @JsonProperty("Data") String data) {
+    return builder()
+        .name(name)
+        .labels(labels)
+        .data(data)
+        .build();
```
mattnworb: why use `builder()` instead of `new AutoValue_SecretSpec`? This won't catch if you add a required property to the SecretSpec class but forget to add it to the static factory method here.

davidxia: If you have a builder, AutoValue creates a private constructor. You have to use the builder.

davidxia:
Test failure is unrelated. Not sure why travis-ci often fails for no reason.

PR #686, opened 2017-03-16
Fix testInspectTask

### LINKED ISSUE #672: Potential race condition in tests?
See

https://travis-ci.org/spotify/docker-client/jobs/208527822
https://travis-ci.org/spotify/docker-client/jobs/208475102
https://travis-ci.org/spotify/docker-client/jobs/208527822
https://travis-ci.org/spotify/docker-client/jobs/209783076
https://travis-ci.org/spotify/docker-client/jobs/208527822
https://travis-ci.org/spotify/docker-client/jobs/209833577
https://travis-ci.org/spotify/docker-client/jobs/210681087
https://travis-ci.org/spotify/docker-client/jobs/211064815
https://travis-ci.org/spotify/docker-client/jobs/211102812
https://travis-ci.org/spotify/docker-client/jobs/211064816

johnflavin:
This fixes failures we have been seeing in `DefaultDockerClientTest#testInspectTask` on docker version 1.12.1. 

#### BACKGROUND
The test would create a service with some task spec, go get the list of all the tasks, and for whichever one is first in that list it would inspect it (which is the ostensible purpose of the test).

But under some unknown conditions, the `Task` that got returned as part of the list would be ever-so-slightly different from the `Task` that got returned by inspecting by ID. Namely, the `Version` of the two instances would be off by one. `Version` is a field that the swarm manager checks when it wants to write some changes to the state of some objects or the system; in your request to update stuff you have to include the current `Version`, which gets incremented after your requested changes are made. In this way they can avoid concurrent modifications, because if two different requests come in to change something, both having the same `Version`, one will get applied, the "correct" `Version` will change, and the other will then be rejected for having the wrong `Version`. 

I guess there is some bug in docker 1.12.1 that makes the `Version` increment upon reading the state of the `Task`s? That would explain the failures we have seen.

#### FIX
Instead of getting two ostensibly identical `Task` objects and asserting equality, I modified the test to assert on the individual properties of the `Task` object. Some of them I knew would be equal to other values I could compare to, for instance from the `ServiceSpec` that was used to create the `Task`s. Others I could assert were `null`. Some of them I did not know what to compare to, so I left assertions commented out. (Alternatively, I could have compared against `Matchers.anything()`.)

Fixes #672

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -242,6 +242,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
+import javax.annotation.Nullable;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
@@ -4751,7 +4752,15 @@ public class DefaultDockerClientTest {
     final List<Task> tasksWithServiceName =
             sut.listTasks(Task.find().serviceName(spec.name()).build());
     assertThat(tasksWithServiceName.size(), is(greaterThanOrEqualTo(1)));
-    assertThat(task, isIn(tasksWithServiceName));
+    final Set<String> taskIds = Sets.newHashSet(
+            Lists.transform(tasksWithServiceName, new Function<Task, String>() {
+              @Nullable
+              @Override
+              public String apply(@Nullable final Task task) {
+                return task == null ? null : task.id();
+              }
+            }));
+    assertThat(task.id(), isIn(taskIds));
   }
 
   @SuppressWarnings("ConstantConditions")
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -84,6 +84,7 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.isEmptyOrNullString;
 import static org.hamcrest.Matchers.isIn;
 import static org.hamcrest.Matchers.lessThan;
+import static org.hamcrest.Matchers.lessThanOrEqualTo;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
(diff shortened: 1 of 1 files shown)


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4564,14 +4564,34 @@ public void testRemoveService() throws Exception {
...
+    assertThat(inspectedTask.version().index(), allOf(notNullValue(), greaterThan(0L)));
+    final Date now = new Date();
+    assertThat(inspectedTask.createdAt(), allOf(notNullValue(), greaterThan(start), lessThan(now)));
+    assertThat(inspectedTask.updatedAt(), allOf(notNullValue(), greaterThan(start), lessThan(now)));
+    assertThat(inspectedTask.slot(), allOf(notNullValue(), greaterThan(0)));
+    assertThat(inspectedTask.status(), notNullValue());
+    assertThat(inspectedTask.name(), nullValue());
+    assertEquals(serviceCreateResponse.id(), inspectedTask.serviceId());
+    assertEquals(serviceSpec.taskTemplate(), inspectedTask.spec());
+    if (serviceSpec.labels() == null || serviceSpec.labels().isEmpty()) {
```
davidxia: What about something like `either( is( empty() ) ).or( is( nullValue() ) ) )`?

davidxia: Wait, why do we need the assertTrue on 4589 at all?

davidxia: How about `either( is( empty() ) ).or( is( nullValue() ) ) ).or(equalTo(inspectedTask.labels()))?`

johnflavin: I tried both of those. First problem: `empty()` only works for `Collection`s, not `Map`s. Hamcrest does have [`IsMapWithSize#anEmptyMap`](https://github.com/hamcrest/JavaHamcrest/blob/142c8a95d3bd4dddaef1bb09eb6875c56646f7a3/hamcrest-library/src/main/java/org/hamcrest/collection/IsMapWithSize.java#L57) but it is not in Hamcrest 1.3.

Even for lists, though, it is not actually easy to combine checks for `null` or `empty`. My IDE complains about both `either( empty() ).or( nullValue() )` and `anyOf( empty(), nullValue() )` because Hamcrest's generics are weird. There is probably a way to get those to work by specifying the classes in some `<>` diamonds, but I don't know how to do it.

johnflavin: And I want to `assertTrue` on 4589 because I'm saying "I gave you a `ServiceSpec` with either null or empty labels, so the `Task` should also have null or empty labels."

davidxia: Ah OK

davidxia:
Hm `testListTaskWithCriteria()` might have the same issue: https://travis-ci.org/spotify/docker-client/jobs/211860339

davidxia:
Looks like recent two versions fail because the task has image name that includes a hash. e.g. https://travis-ci.org/spotify/docker-client/jobs/211860341

```
Failed tests:   testInspectTask(com.spotify.docker.client.DefaultDockerClientTest): expected:<TaskSpec{containerSpec=ContainerSpec{image=alpine, labels={}, command=[ping, -c1000, localhost], args=null, env=null, dir=null, user=null, groups=null, tty=null, mounts=null, stopGracePeriod=null}, resources=null, restartPolicy=null, placement=null, networks=null, logDriver=null}> but was:<TaskSpec{containerSpec=ContainerSpec{image=alpine:latest@sha256:58e1a1bb75db1b5a24a462dd5e2915277ea06438c3f105138f97eb53149673c4, labels={}, command=[ping, -c1000, localhost], args=null, env=null, dir=null, user=null, groups=null, tty=null, mounts=null, stopGracePeriod=null}, resources=null, restartPolicy=null, placement=null, networks=null, logDriver=null}>
```

johnflavin:
Yeah, I noticed that. I just need to compare those `ContainerSpec`s property-by-property. There is already a special `Matcher` for this `alpine` name comparison, I think.

PR #687, opened 2017-03-18
Test on 1.12.6 instead of 1.12.1

davidxia:
There might be a bug with Task.Version in 1.12.1
that is fixed in a later minor release.
See https://github.com/spotify/docker-client/issues/672.
In any case, we should test on the latest minor release
for each major release. The other major releases we test
on are at their latest minor releases.


davidxia:
cc @johnflavin

johnflavin:
Fine with me. I think we took care of all existing `Task` comparison tests, but any new tests on those objects would keep failing in the same way. If bumping 1.12 up to a new version can prevent that, sounds good. 

And it makes sense to me to test on the highest minor version of each major version. If we can't test on every minor version, might as well get the ones likely to contain fewer bugs.

PR #688, opened 2017-03-18
Rerun failing tests up to two times on travis-ci

davidxia:
Travis builds often fail due to errors pulling images needed
for tests. Retry up to two times before failing a test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_install:
   - docker version
   - docker info
 
-script: mvn clean test
+script: mvn clean -Dsurefire.rerunFailingTestsCount=1 test
 
 after_success:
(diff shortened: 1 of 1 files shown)


davidxia:
cc @johnflavin

mattnworb:
With this I think a failing test would be run a total of 3 times, is that intended?

johnflavin:
Wow. This is a lot like what I was working on last night: run all the tests using a subclass of `DefaultDockerClient` that overrides only the `pull`s, in which it passes the `super` method as a `Callable` to a method that re-tries up to ten times. 

Yours is much simpler and better :+1:

davidxia:
Oh i meant 1. updated.

Issue #689, opened 2017-03-20
listServices method throwing exception


markmclaren:
## Description

Whilst attempting to use the listServices method, an exception is produced

## How to reproduce

```
DockerClient docker = DefaultDockerClient.fromEnv().build();
List<Service> services = docker.listServices();
docker.close();
```

## What do you expect

I expected it to work without an exception.

## What happened instead

Exception thrown.

## Software:

- `docker version`: 1.13.1
- Spotify's docker-client version: 8.1.2

## Full backtrace

```
Exception in thread "main" com.spotify.docker.client.exceptions.DockerRequestException: Request error: GET https://blah.com:12345/services: 200
	at com.spotify.docker.client.DefaultDockerClient.propagate(DefaultDockerClient.java:2165)
	at com.spotify.docker.client.DefaultDockerClient.request(DefaultDockerClient.java:2094)
	at com.spotify.docker.client.DefaultDockerClient.listServices(DefaultDockerClient.java:1680)
	at edu.bristol.webfarmrepdockerclientexperiment.Test.main(Test.java:21)
Caused by: javax.ws.rs.client.ResponseProcessingException: com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of com.spotify.docker.client.messages.mount.Driver, problem: abstract types either need to be mapped to concrete types, have custom deserializer, or be instantiated with additional type information
 at [Source: org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream@582571ea; line: 1, column: 4164] (through reference chain: java.util.ArrayList[2]->com.spotify.docker.client.messages.swarm.Service["Spec"]->com.spotify.docker.client.messages.swarm.ServiceSpec["TaskTemplate"]->com.spotify.docker.client.messages.swarm.TaskSpec["ContainerSpec"]->com.spotify.docker.client.messages.swarm.ContainerSpec["Mounts"]->java.util.ArrayList[1]->com.spotify.docker.client.messages.mount.Mount["VolumeOptions"]->com.spotify.docker.client.messages.mount.VolumeOptions["DriverConfig"])
	at org.glassfish.jersey.client.JerseyInvocation.translate(JerseyInvocation.java:869)
	at org.glassfish.jersey.client.JerseyInvocation.access$800(JerseyInvocation.java:92)
	at org.glassfish.jersey.client.JerseyInvocation$6.completed(JerseyInvocation.java:836)
[...]

PR #690, opened 2017-03-20
Fix for "listServices method throwing exception #689"

markmclaren:
Fix for:

```java
DockerClient docker = DefaultDockerClient.fromEnv().build();
List<Service> services = docker.listServices();
docker.close();
```

throwing:

```
Exception in thread "main" com.spotify.docker.client.exceptions.DockerRequestException: Request error: GET https://blah.com:12345/services: 200
	at com.spotify.docker.client.DefaultDockerClient.propagate(DefaultDockerClient.java:2165)
	at com.spotify.docker.client.DefaultDockerClient.request(DefaultDockerClient.java:2094)
	at com.spotify.docker.client.DefaultDockerClient.listServices(DefaultDockerClient.java:1680)
	at com.blah.Test.main(Test.java:16)
Caused by: javax.ws.rs.client.ResponseProcessingException: com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of com.spotify.docker.client.messages.mount.Driver, problem: abstract types either need to be mapped to concrete types, have custom deserializer, or be instantiated with additional type information
```


src/main/java/com/spotify/docker/client/messages/mount/Driver.java
```diff
@@ -24,19 +24,24 @@
...
 import com.google.common.collect.ImmutableMap;
 import java.util.Map;
 
+import javax.annotation.Nullable;
+
 @AutoValue
 @JsonAutoDetect(fieldVisibility = ANY, getterVisibility = NONE, setterVisibility = NONE)
 public abstract class Driver {
 
+  @Nullable
```
johnflavin: Does `name` need to be `@Nullable`? Can you have a `DriverConfig` that is non-null but has a null name?

markmclaren: I'm not really sure about the @Nullable on name.  Might there be an occasion when options are entered against the default driver? (I'm no Docker expert)

davidxia:
replaced by https://github.com/spotify/docker-client/pull/695

PR #693, opened 2017-03-21
Add hostname into ContainerSpec, fixing issue: 692

Noodle05:
Test: Modified testInspectService to create service with hostname on ContainerSpec, and compare with inspect value.

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/swarm/ContainerSpec.java
+++ b/src/main/java/com/spotify/docker/client/messages/swarm/ContainerSpec.java
@@ -43,7 +43,7 @@ public abstract class ContainerSpec {
   public abstract String image();
 
   /**
-   * @since API 1.25
+   * @since API 1.26
    */
   @Nullable
   @JsonProperty("Hostname")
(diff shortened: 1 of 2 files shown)


src/main/java/com/spotify/docker/client/messages/swarm/ContainerSpec.java
```diff
@@ -43,6 +43,10 @@
   public abstract String image();
 
   @Nullable
+  @JsonProperty("Hostname")
+  public abstract String hostname();
```
johnflavin: Add a little bit of javadoc here with the property `@since API 1.25`.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4549,6 +4551,7 @@ public void testInspectService() throws Exception {
     final Matcher<String> imageMatcher = dockerApiVersionLessThan("1.25")
                                          ? is("alpine") : startsWith("alpine:latest@sha256:");
     assertThat(service.spec().taskTemplate().containerSpec().image(), imageMatcher);
+    assertThat(service.spec().taskTemplate().containerSpec().hostname(), is(hostname));
```
johnflavin: Wrap this check inside `if (dockerApiVersionGreaterThan("1.24")) {...}`.

Noodle05:
After compare Docker API, looks like hostname been added since API version 1.26. I cannot find it in API version 1.25

https://docs.docker.com/engine/api/v1.25/#operation/ServiceCreate

Will change it accordingly.

PR #706, opened 2017-03-28
Handle empty sub-directories when compressing contents.

gavinbunney:
This fixes an issue where empty subdirectories were being ignored (as they didn’t have their `visitor.visitFile` method called). This better aligns with the default unix tar implementation, whereby all directories are created as archive entries, regardless if they have contents or not.

### DIFF
--- a/src/main/java/com/spotify/docker/client/CompressedDirectory.java
+++ b/src/main/java/com/spotify/docker/client/CompressedDirectory.java
@@ -265,6 +265,27 @@ class CompressedDirectory implements Closeable {
     }
 
     @Override
+    public FileVisitResult preVisitDirectory(Path dir,
+                                             BasicFileAttributes attrs) throws IOException {
+      if (Files.isSameFile(dir, root)) {
+        return FileVisitResult.CONTINUE;
+      }
+
+      final Path relativePath = root.relativize(dir);
+
+      if (exclude(ignoreMatchers, relativePath)) {
+        return FileVisitResult.CONTINUE;
+      }
+
+      final TarArchiveEntry entry = new TarArchiveEntry(dir.toFile());
+      entry.setName(relativePath.toString());
+      entry.setMode(getFileMode(dir));
+      tarStream.putArchiveEntry(entry);
+      tarStream.closeArchiveEntry();
+      return FileVisitResult.CONTINUE;
+    }
+
+    @Override
     public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
 
       final Path relativePath = root.relativize(file);
--- a/src/test/java/com/spotify/docker/client/CompressedDirectoryTest.java
+++ b/src/test/java/com/spotify/docker/client/CompressedDirectoryTest.java
@@ -20,6 +20,7 @@
 
 package com.spotify.docker.client;
 
+import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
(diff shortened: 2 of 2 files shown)


src/test/java/com/spotify/docker/client/CompressedDirectoryTest.java
```diff
@@ -74,9 +78,31 @@ public void testFileWithIgnore() throws Exception {
         final String name = entry.getName();
         names.add(name);
       }
-      assertThat(names, containsInAnyOrder("Dockerfile", "bin/date.sh", "subdir2/keep.me",
-                                           "subdir2/do-not.ignore", "subdir3/do.keep",
-                                           ".dockerignore"));
+      assertThat(names, containsInAnyOrder("Dockerfile", "bin/", "bin/date.sh", "subdir2/",
+                                           "subdir2/keep.me", "subdir2/do-not.ignore",
+                                           "subdir3/do.keep", ".dockerignore"));
```
davidxia: Did you mean to also include `subdir3/`?

src/main/java/com/spotify/docker/client/CompressedDirectory.java
```diff
@@ -265,6 +265,27 @@ private Visitor(final Path root, ImmutableList<DockerIgnorePathMatcher> ignoreMa
...
+
+      final Path relativePath = root.relativize(dir);
+
+      if (exclude(ignoreMatchers, relativePath)) {
+        return FileVisitResult.CONTINUE;
+      }
+
+      final TarArchiveEntry entry = new TarArchiveEntry(dir.toFile());
+      entry.setName(relativePath.toString());
+      entry.setMode(getFileMode(dir));
```
davidxia: I'm not that familiar with `TarArchiveEntry`, but I guess we don't need `entry.setSize()`?

src/test/java/com/spotify/docker/client/CompressedDirectoryTest.java
```diff
@@ -74,9 +78,31 @@ public void testFileWithIgnore() throws Exception {
...
+         TarArchiveInputStream tarIn = new TarArchiveInputStream(gzipIn)) {
+
+      final List<String> names = new ArrayList<>();
+      TarArchiveEntry entry;
+      while ((entry = tarIn.getNextTarEntry()) != null) {
+        final String name = entry.getName();
+        names.add(name);
+      }
+      assertThat(names,
+          containsInAnyOrder("emptySubDir/"));
```
davidxia: just use `contains()`?

davidxia:
@rohansingh @mattnworb Any reason we were skipping empty dirs? Will this break anything?

src/test/java/com/spotify/docker/client/CompressedDirectoryTest.java
gavinbunney: That's an interesting one, since `subdir3` is in the [.dockerignore](https://github.com/spotify/docker-client/blob/master/src/test/resources/dockerDirectoryWithIgnore/.dockerignore) file it'll ignore that directory, but still include the `subdir3/do.keep` file. I think it's ok to handle it that way as the directory is still 'created' by the other file handler

src/test/java/com/spotify/docker/client/CompressedDirectoryTest.java
gavinbunney: Lazy copy & paste :) Will fix

src/main/java/com/spotify/docker/client/CompressedDirectory.java
gavinbunney: Yep indeed as the directory has no size, just an index in the archive

PR #707, opened 2017-03-31
Add support for network filters

### LINKED ISSUE #694: Support for network labels
docker-client lacks support for network labels and searching networks based on network labels.
I'd like to contribute this. Would this be appreciated?

albers:
This PR adds support for filters to `DockerClient.listNetworks()` as suggested in #694.
Closes #694. Needs discussion.

### Naming Issues
I was very unsure how to name the factory methods for the filter params because there are some inconsistencies in the existing list methods:

    ListContainersParam
      withExitStatus(int exitStatus)
      withStatusXXX()
      withLabel(String label, String value)
    ListImagesParam
      danglingImages()
      withLabel(String label, String value)
    EventsParam
      event(String event)
      image(String image)
      container(String container)
      volume(String volume)
      network(String network)
      type(Event.Type type)
      label(String label, String value)
    ListVolumesParam
      name(String name)
      dangling()
      driver(String driver)

So I could have named the methods `withDriver` or `driver` , `withLabel` or `label` etc.

I'd like to suggest a third variant: Given that the factory methods are syntactical sugar, they offer most value if they can be used with static imports, i.e.

    client.listNetworks(withLabel("stage", "production"), custom())

instead of

    client.listNetworks(ListNetworkParam.withLabel("stage", "production"), ListNetworkParam.custom())

This does not work well if the various `XXXListParam` types have name collisions, e.g. `.withLabels()` and you want to use several of them in one class. One solution would be to name the methods like `withNetworkLabel`, `withImageLabel`, `withContainerLabel` and so on.

In this PR, I used such names. I'm OK with other naming schemes, just let me know and I'll update the PR.

### Consistency

IMHO consistency in the names of the existing factory methods should be improved. New methods should be added and ill-named ones should be deprecated.

If you agree, just give me a ping and I'll create a follow-up PR.

### Refactoring

I stepped into some repetitions in the filter methods that I resolved in additional commits. The third commit deserves special attention because it changes the type hierarchy.

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/Network.java
+++ b/src/main/java/com/spotify/docker/client/messages/Network.java
@@ -26,11 +26,11 @@ import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.NONE;
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonValue;
 import com.google.auto.value.AutoValue;
 
 import com.google.common.collect.ImmutableMap;
 
-import java.util.Locale;
 import java.util.Map;
 import javax.annotation.Nullable;
 
@@ -125,13 +125,20 @@ public abstract class Network {
    */
   public enum Type {
     /** Predefined networks that are built-in into Docker. */
-    BUILTIN,
-    /** Custom networks created by users. */
-    CUSTOM;
+    BUILTIN("builtin"),
+    /** Custom networks that were created by users. */
+    CUSTOM("custom");
     
-    @Override
-    public String toString() {
-      return name().toLowerCase(Locale.ENGLISH);
+    private final String name;
+
+    @JsonCreator
+    Type(final String name) {
+      this.name = name;
+    }
+
+    @JsonValue
+    public String getName() {
+      return name;
     }
   }
 }
--- a/src/main/java/com/spotify/docker/client/messages/Network.java
+++ b/src/main/java/com/spotify/docker/client/messages/Network.java
@@ -29,6 +29,8 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 
 import com.google.common.collect.ImmutableMap;
+
+import java.util.Locale;
 import java.util.Map;
 import javax.annotation.Nullable;
 
@@ -115,6 +117,21 @@ public abstract class Network {
         @JsonProperty("IPv4Address") final String ipv4Address,
         @JsonProperty("IPv6Address") final String ipv6Address) {
       return new AutoValue_Network_Container(endpointId, macAddress, ipv4Address, ipv6Address);
+    }
+  }
+  
+  /**
+   * Docker networks come in two kinds: built-in or custom. 
+   */
+  public enum Type {
+    /** Predefined networks that are built-in into Docker. */
+    BUILTIN,
+    /** Custom networks created by users. */
+    CUSTOM;
+    
+    @Override
+    public String toString() {
+      return name().toLowerCase(Locale.ENGLISH);
     }
   }
 }
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -1941,8 +1941,32 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   }
 
   @Override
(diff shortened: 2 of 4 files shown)


johnflavin:
Re: failures on test [#707](https://travis-ci.org/spotify/docker-client/builds/217245108)

It looks to me like your test is failing because not all filters are available in all docker API versions. I bet that docker rolled out support for network filters in whatever version you’re using as your `assumeDockerApiVersionAtLeast("X")`, but then in later versions added additional filters. You'll have to check out the various older docker API documentation to see which filters were added when. Then in your test, wrap the later-introduced filters in `if(dockerApiVersionGreaterThan("Y"))`.

albers:
Thanks @johnflavin . I think I found it: I used the `driver` filter in a helper function which was triggered for _all_ new tests.

albers:
@johnflavin can you please trigger re-execution of the tests? The failures are unrelated.

johnflavin:
I don't have the ability to do that. But we're familiar with these kinds of test failures, even if we don't know the cause. It's no big deal.

albers:
@johnflavin Do failing tests block the review process?
Would you like me to re-trigger builds with dummy pushes until all tests succeed?

johnflavin:
No, we don't require that the tests all pass. It would be nice if we could, but the test failures seem to be somewhat beyond our control. Or, at the very least, we live with them for now. (If you're interested in another PR opportunity, figuring out why these tests fail and fixing things would be a big help.)

src/main/java/com/spotify/docker/client/messages/Network.java
```diff
@@ -117,4 +119,19 @@ static Container create(
...
+   */
+  public enum Type {
+    /** Predefined networks that are built-in into Docker. */
+    BUILTIN,
+    /** Custom networks created by users. */
+    CUSTOM;
+    
+    @Override
+    public String toString() {
+      return name().toLowerCase(Locale.ENGLISH);
```
johnflavin: This isn't consistent with the style of other enums around `docker-client`. But, to be fair, the other enums aren't consistent with each other either. Could you do one or more of these things:
1. Make this look more like `Event.Type`
    1. Use an explicit `String name` rather than converting the enum's name to lowercase, and 
    2. Annotate the getter for that name with `@JsonValue`.
1. If you don't want to do that, consider changing from `Locale.ENGLISH` to `Locale.ROOT`. This is used for the `AttachParam` enum [here](https://github.com/spotify/docker-client/blob/master/src/main/java/com/spotify/docker/client/DefaultDockerClient.java#L1553).

Related: I don't know if this `Type` will ever be deserialized from JSON, or will only ever be serialized out. If we will ever deserialize this it needs a creator method annotated with `@JsonCreator`. Again see `Event.Type`.

albers: @johnflavin `Network.Type` is not used in JSON. It is just syntactical sugar so that you don't have to worry about the possible values for `type`. The enum is converted to a string almost immediately. But usage might change at some point, so it's a good idea to make it JSON friendly. I'll adjust it to `Event.Type`.

albers:
@johnflavin @davidxia Do you think that consistency of the static factory methods in`ListContainersParam`, `ListImagesParam`, `EventsParam` and `ListVolumesParam` should be increased as discussed above?

If not, I would update this PR so that the method names of `ListNetworkParam` match the scheme used in `ListContainersParam` and `ListImagesParam` (i.e. `withXXX()` for filters that are likely to return multiple values, `byXXX` for those that match one item (unless partial keys are specified)).

johnflavin:
I do favor increased naming consistency. But I am not sure I care one way or the other about naming the methods `FooParam.withLabels()` vs `FooParam.withFooLabels()`. The former means you can't do a static method import because lots of params classes will have `withLabels()` methods; the latter means you can do a static method import but the method names are all longer and more cumbersome.  They both have positive and negative tradeoffs. So follow your heart, I guess.

albers:
It's really hard to decide. Names like `withImageLabels` and `withNetworkLabels` imply that labels for images and networks are different things, which is wrong.
Perhaps it's best to use names like `withLabels` here.

albers:
The PR was updated according to the review comments. PTAL and tell me whether there are any issues or if I should squash the commits.

albers:
@johnflavin Is there anything I can do to get this merged?

mattnworb:
thank you also for addressing and improving the consistency with the label classes

> IMHO consistency in the names of the existing factory methods should be improved. New methods should be added and ill-named ones should be deprecated.
> If you agree, just give me a ping and I'll create a follow-up PR.

I'm not sure if there is more to do here, but I'd welcome any additional improvements with these names 

> It's really hard to decide. Names like withImageLabels and withNetworkLabels imply that labels for images and networks are different things, which is wrong.
> Perhaps it's best to use names like withLabels here.

+1

PR #711, opened 2017-04-03
Add new fields to ContainerMount

### LINKED ISSUE #709: Get volume names when inspecting a container
## Description

I am trying to remove all volumes used by a container, but AFAIK, that cannot be done. ContainerInfo.mount() returns a list of ContainerMount, but DockerClient.removeVolume() accepts either a volume name or a Volume object. As ContainerMount doesn't expose the name attribute and it is not a subclass of Volume, that information cannot be used to remove the volume

## How to reproduce

It is (if I'm not wrong), a defect on the API.

## What do you expect

I would like to have a way to translate from a ContainerMount to a Volume or, at least, to get the name of the mount. Ideally, I would like to be access to other mount attributes like its type.

## What happened instead

ContainerMount does not expose some attributes you can find with docker inspect, and specially the name attribute is needed to call DockerClient.removeVolume

## Software:

- `docker version`: Client: 17.03.0-ce, Server: 17.03.0-ce
- Spotify's docker-client version: 8.2.1

johnflavin:
Fixes #709

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/ContainerMount.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerMount.java
@@ -28,10 +28,20 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 
+import javax.annotation.Nullable;
+
 
 @AutoValue
 @JsonAutoDetect(fieldVisibility = ANY, getterVisibility = NONE, setterVisibility = NONE)
 public abstract class ContainerMount {
+
+  @Nullable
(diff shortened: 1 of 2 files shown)


davidxia:
Looks like real test failures?

johnflavin:
Fixed that test and rebased. Sorry that I missed that test failure; I forgot that I had this PR open.

johnflavin:
Looks like there are different test failures on the different docker versions. I'll have to go through them all and make my assertions more carefully.

johnflavin:
Tests are green, other than the usual inexplicable failures. Ready for review.

davidxia:
cc @mattnworb in case you want to take a look.

PR #712, opened 2017-04-06
Added functionality to list all existing nodes in the swarm

coenvl:
A function added to the interface, as well as the implementation in the default client, and all required messages according to the docker api.


johnflavin:
Looks good to me. Could you write a test for it in `DefaultDockerClientTest`?

coenvl:
I just pushed the tests. Nothing fancy, just to make sure all fields are serialized correctly.

PR #713, opened 2017-04-06
added killing containers with other signals

### LINKED ISSUE #614: Allow custom signal when "killing" container
The `POST /containers/(id or name)/kill` endpoint long supports a `signal` query parameter to control what signal is send to the container. It would be nice if that query parameter was exposed in the `DockerClient.killContainer()` API somehow.

Of particular interest/value is the ability to send SIGQUIT to the container which triggers a thread dump from Java processes.

MarcoLotz:
fixes #614
added Signal types for killing containers

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -127,6 +127,7 @@ import java.net.URLEncoder;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -659,17 +660,17 @@ public class DefaultDockerClient implements DockerClient, Closeable {
 
   private void containerAction(final String containerId, final String action)
       throws DockerException, InterruptedException {
-    containerAction(containerId, action, new QueryParameter[] {});
+    containerAction(containerId, action, new ArrayList<Map.Entry<String,String>>());
   }
 
   private void containerAction(final String containerId, final String action,
-                               final QueryParameter... queryParameters)
+                               final List<Map.Entry<String,String>> queryParameters)
           throws DockerException, InterruptedException {
     try {
       final WebTarget resource = resource()
               .path("containers").path(containerId).path(action);
-      for (QueryParameter queryParameter : queryParameters) {
-        resource.queryParam(queryParameter.getField(), queryParameter.getValue());
+      for (Map.Entry<String, String> queryParameter : queryParameters) {
+        resource.queryParam(queryParameter.getKey(), queryParameter.getValue());
       }
       request(POST, resource, resource.request());
     } catch (DockerRequestException e) {
@@ -679,35 +680,6 @@ public class DefaultDockerClient implements DockerClient, Closeable {
         default:
           throw e;
       }
-    }
-  }
-
-  /**
-   * Query parameters used for {@link containerAction}.
-   */
-  private class QueryParameter {
-    private String field;
-    private String value;
-
-    public QueryParameter(String field, String value) {
-      this.field = field;
-      this.value = value;
-    }
-
-    public void setField(String field) {
-      this.field = field;
-    }
-
-    public String getField() {
-      return field;
-    }
-
-    public void setValue(String value) {
-      this.value = value;
-    }
-
-    public String getValue() {
-      return value;
     }
   }
 
@@ -760,8 +732,12 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   public void killContainer(final String containerId, final Signal signal)
       throws DockerException, InterruptedException {
     checkNotNull(containerId, "containerId");
-    QueryParameter signalParam = new QueryParameter("signal", signal.getName());
-    containerAction(containerId, "kill", signalParam);
+    Map.Entry<String, String> signalParam =
+            new AbstractMap.SimpleEntry<String, String>("signal", signal.getName());
+    List<Map.Entry<String, String>> queryParameterList = new ArrayList<>();
+    queryParameterList.add(signalParam);
+
+    containerAction(containerId, "kill", queryParameterList);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -1003,6 +1001,39 @@ void restartContainer(String containerId, int secondsToWaitBeforeRestart)
...
+   * @throws DockerException      if a server error occurred (500)
+   * @throws InterruptedException If the thread is interrupted
+   */
+  void killContainer(final String containerId, final TerminationSignals signal)
+      throws DockerException, InterruptedException;
+
+  /**
+   * Supported parameters for {@link #killContainer(String, TerminationSignals)}).
+   */
+  enum TerminationSignals {
```
johnflavin: This should be named `TerminationSignal`, singular.

MarcoLotz:
@johnflavin changed.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -715,10 +715,31 @@ public void restartContainer(String containerId, int secondsToWaitBeforeRestart)
...
     containerAction(containerId, "kill");
   }
 
   @Override
+  public void killContainer(final String containerId, final TerminationSignal signal)
+      throws DockerException, InterruptedException {
+    checkNotNull(containerId, "containerId");
+
+    try {
+      final WebTarget resource = noTimeoutResource()
```
mattnworb: why use the `noTimeout` client here?

seems like there is also room to re-use containerAction, by refactoring it to have an overload where you can pass additional query params, to maintain the idea that most of these methods that call `POST /containers/:id/:action` flow through the same method

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1366,6 +1366,72 @@ public void testRestartContainer() throws Exception {
...
+    {
+      final ContainerInfo containerInfo = sut.inspectContainer(containerId);
+      assertThat(containerInfo.state().running(), equalTo(true));
+    }
+
+    sut.killContainer(containerId);
+
+    // Should not be running
+    {
+      final ContainerInfo containerInfoLatest = sut.inspectContainer(containerId);
```
mattnworb: just curious: why wrap this in a block if you use different names for the ContainerInfo instances and therefore there is no name collision?

src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -990,9 +990,7 @@ void restartContainer(String containerId, int secondsToWaitBeforeRestart)
 
   /**
    * Kill a docker container.
-   * Note: This implementation deviates from the Docker Remote API. The
-   * latter accepts the kill signal as an argument. This implementation does
-   * not accept the signal argument; instead, the default SIGKILL is sent.
```
mattnworb: good improvement here, thanks

src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -1003,6 +1001,39 @@ void restartContainer(String containerId, int secondsToWaitBeforeRestart)
...
+  /**
+   * Supported parameters for {@link #killContainer(String, TerminationSignal)}).
+   */
+  enum TerminationSignal {
+    SIGINT("SIGINT"),
+    SIGQUIT("SIGQUIT"),
+    SIGTERM("SIGTERM"),
+    SIGKILL("SIGKILL"),
+    SIGSTOP("SIGSTOP");
+
```
MarcoLotz: Well spotted. Totally agree with you. Any other signal? I can add all (http://www.comptechdoc.org/os/linux/programming/linux_pgsignals.html) just to be future friendly.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
MarcoLotz: Good point. Will take a look into it.

src/main/java/com/spotify/docker/client/DockerClient.java
albers: Here's what Dockers bash completion uses:
```bash
        local signals=(
                SIGCONT
                SIGHUP
                SIGINT
                SIGKILL
                SIGQUIT
                SIGSTOP
                SIGTERM
                SIGUSR1
                SIGUSR2
        )
```
This is quite an arbitrary subset of the available signals, but nobody ever complained so far.

albers: I think it's best to add all signals because else there would be no way to specify a value missing in the enum. (for bash completion, it's different because the user can enter any value.)

If the enum offeres a subset of the available signals, there should be an overloaded `String` variant that allows passing the missing signals. I prefer the first solution.

MarcoLotz:
@albers: added all the signals and changed naming.

@mattnworb:
i) added overloaded definition to containerAction. Originally I had defined a nested class for the query parameters and used varargs in the containerAction signature. I've changed to Map.Entry to avoid private nested classes. Changed from varargs of Map.Entry<String, String> to a List of Map.Entry<String,String> because was getting many warnings on type inference. The only solution would have been to add a SuppressWarnings on it. Should you decide that going for the nested Query Parameters class and vargs is a better solution (or anything like it) let me know and I will update it.

ii) Updated the restartContainer method to use the new containerAction. Should any further refactoring be required in order to use the new containerAction method, probably would be better to do in another pull request

johnflavin:
Instead of a `List<Map.Entry<String, String>>`, couldn't you accomplish the same thing by just having the method take a `Map<String, String>`?

MarcoLotz:
Well noted. To use a map, we would have to assume that each query parameter field is unique. It may be valid now for docker API, but there's no spec forcing query parameters fields to be unique (http://stackoverflow.com/questions/1746507/authoritative-position-of-duplicate-http-get-query-keys). That's why I used a List instead. I totally agree that Map would be a lot cleaner.

johnflavin:
You’re correct that for general HTTP APIs, query params aren't guaranteed to have unique keys. But they are unique in docker's API (at least as it exists so far). When docker's API needs to allow multiple values for one parameter, as can be the case with `filter` params, they want the params in the form `name=[val1, val2,...]` rather than `name=val1&name=val2&...`.

mattnworb:
other places in the docker-client codebase use Jersey's `MultivaluedMap` for just this purpose

MarcoLotz:
@johnflavin @mattnworb awesome! This makes things a lot cleaner and simpler. Will do the changes.

davidxia:
@mattnworb Is this good to go? Do we want `Signal` to be an inner enum or in a separate file?

mattnworb:
@davidxia doesn't make much difference to me.

Issue #714, opened 2017-04-07
DockerClient.inspectService throws NPE on new service


markvr:
## Description

Docker 17.04 seems to have removed UpdateStatus from the `service inspect` JSON response on new services.  This causes an NPE in spotify-client.

## How to reproduce

    final String serviceName = "tomcat";
    final TaskSpec taskSpec = TaskSpec.builder()
            .containerSpec(ContainerSpec.builder().image("tomcat").build())
            .build();
    final ServiceSpec spec = ServiceSpec.builder().name(serviceName)
            .taskTemplate(taskSpec).mode(ServiceMode.withReplicas(1L))
            .build();

    final ServiceCreateResponse response = client.createService(spec);

    final Service inspectService = client.inspectService(serviceName);
    System.out.println(inspectService);

```
Caused by: java.lang.NullPointerException: Null updateStatus
	at com.spotify.docker.client.messages.swarm.AutoValue_Service.<init>(AutoValue_Service.java:52)
	at com.spotify.docker.client.messages.swarm.Service.create(Service.java:138)
```

## Software:

Client:
 Version:      17.03.0-ce
 API version:  1.26
 Go version:   go1.7.5
 Git commit:   60ccb22
 Built:        Thu Mar  2 01:11:00 2017
 OS/Arch:      windows/amd64

Server:
 Version:      17.04.0-ce
 API version:  1.28 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   4845c56
 Built:        Wed Apr  5 18:55:10 2017
 OS/Arch:      linux/amd64
 Experimental: false

- Spotify's docker-client version: 8.3.1

## Full backtrace
```

cd C:\work\temp\docker-spotify-test; JAVA_HOME=C:\\programs\\jdk1.8.0_102 M2_HOME=C:\\programs\\maven-3.3.9 cmd /c "\"\"C:\\programs\\maven-3.3.9\\bin\\mvn.cmd\" -Dexec.args=\"-classpath %classpath uk.ac.bris.adt.docker.spotify.test.Test\" -Dexec.executable=C:\\programs\\jdk1.8.0_102\\bin\\java.exe -Dmaven.ext.class.path=\"C:\\Program Files\\NetBeans 8.2\\java\\maven-nblib\\netbeans-eventspy.jar\" -Dfile.encoding=UTF-8 org.codehaus.mojo:exec-maven-plugin:1.2.1:exec\""
[...]

PR #718, opened 2017-04-11
Add docker 17.04 CE to the build matrix to check swarm api breakage

ketan:
See #714

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
     - DOCKER_VERSION=1.12.6 ENABLE_SWARM=1
     - DOCKER_VERSION=1.13.1 ENABLE_SWARM=1 UPLOAD_SONATYPE=1
     - DOCKER_VERSION=17.03.0~ce ENABLE_SWARM=1
+    - DOCKER_VERSION=17.04.0~ce ENABLE_SWARM=1
 
 before_install:
   # check what version of docker is installed beforehand
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -4586,8 +4586,8 @@ public class DefaultDockerClientTest {
     try {
       sut.deleteSecret(secretId);
       fail("Should fail because of non-existant secret ID");
-    } catch (DockerRequestException ex) {
-      // Ignored; Docker should return status code 404, but it doesn't for some reason.
+    } catch (DockerRequestException | NotFoundException ex) {
+      // Ignored; Docker should return status code 404, but it doesn't for some reason, so we catch 2 different exceptions
     }
   }
 
--- a/src/main/java/com/spotify/docker/client/messages/swarm/Service.java
+++ b/src/main/java/com/spotify/docker/client/messages/swarm/Service.java
@@ -57,6 +57,7 @@ public abstract class Service {
   public abstract Endpoint endpoint();
 
   @JsonProperty("UpdateStatus")
+  @Nullable
   public abstract UpdateStatus updateStatus();
 
   @AutoValue


.travis.yml
```diff
@@ -23,6 +23,7 @@ env:
     - DOCKER_VERSION=1.12.6 ENABLE_SWARM=1 UPLOAD_SONATYPE=1
     - DOCKER_VERSION=1.13.1 ENABLE_SWARM=1
     - DOCKER_VERSION=17.03.0~ce ENABLE_SWARM=1
+    - DOCKER_VERSION=17.04.0~ce ENABLE_SWARM=1
```
davidxia: To keep the build matrix limited, I'm fine with replacing 17.03 with 17.04. Thoughts @mattnworb?

ketan: I guess there was a fundamental behavioral change between the two docker versions. So I kept it AS-IS to ensure that any further changes to the client don't break something that was working with older versions of docker.

mattnworb: assuming Docker is changing the API version for each API change (or that they are not sneaking changes or fixes in) then I think we should test against each API version. Looks to me [like API 1.27 == 17.03.1 and API 1.26 == 17.03.0](https://docs.docker.com/engine/api/v1.27/#section/Versioning), so it sounds like we should either test against both 17.03.0 and 13.03.1 _or_ 13.03.0 and 13.04.0.

mattnworb: I rather drop old versions off of the start of this list rather than replace the newer versions

davidxia: Good point. I'm also fine dropping 1.9.1.

johnflavin: I want to throw this idea out for consideration: The Docker team is moving to a much more rapid release schedule, with a major `yy.mm.0` release every month and possibly minor version releases as well. So should we continue the policy to only keep the latest six major releases, or should we keep more? Is it a good idea to only test for compatibility with the releases from the last six months, rather than the years worth of releases we have now?

mattnworb: I'm not really sure to be honest. I am not crazy about throwing out tests for old versions as we should try to ensure we are still compatible there, but I don't want to miss things in new versions either. On the other hand the more versions we add, the longer Travis builds take.

Perhaps we could start by continually adding new versions and stretching out the number of versions we test to be more than six, and then we can reduce the number when/if we find the long build times are too cumbersome.

ketan: Would it be OK to merge this PR and punt on the supported releases in a separate issue?

davidxia: @mattnworb Sounds good.

davidxia:
I'm fine with merging once the tests pass on 17.04. Right now `testSecretOperations()` fails.

ketan:
> I'm fine with merging once the tests pass on 17.04. Right now testSecretOperations() fails.

I believe this failure is because of [this change](https://github.com/spotify/docker-client/commit/62289ec9ed828b5654986f755484eac63cd4eebe). I've "fixed" this in 694747cd65d26e2f2a21260cbc1b3178dc38d829, but it seems to be related to a change in API contract.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4586,8 +4586,8 @@ public void testSecretOperations() throws Exception {
     try {
       sut.deleteSecret(secretId);
       fail("Should fail because of non-existant secret ID");
-    } catch (DockerRequestException ex) {
-      // Ignored; Docker should return status code 404, but it doesn't for some reason.
+    } catch (DockerRequestException | NotFoundException ex) {
+      // Ignored; Docker should return status code 404, but it doesn't for some reason, so we catch 2 different exceptions
```
davidxia: Can you clarify this comment to say it doesn't return 404 in versions < 17.04?

davidxia:
closed in favor of #728

PR #719, opened 2017-04-11
Improve consistency in ListXXXParameters classes

albers:
See the discussion in #707 for naming issues in the static factory methods of several `ListXXXParameters` classes.

In this PR, for several short-named methods (e.g. `driver`) aliases with more consistent names (e.g. `withDriver`) are **added**.
I leave it to discussion whether the old methods should get deprecated or not. Currently they are not. 

In cases where you want to do imports for a bunch of static methods from several `ListXXXParameters` classes, it might be helpful to be able to choose from several names for the same function. So it might be a good idea not to deprecate the methods.

Following this argument further on, we could even create more aliases like `withNetworkLabel`, `withImageLabel`, `withContainerLabel` as suggested but dropped in #707.

On the other hand, the choice between several names for the same thing might be confusing to the user. I tried to mitigate this by adding some `@see` javadoc references between the related methods.

WDYT?

Notes:
* For filter parameters that target exactly one instance as event source, I chose the prefix `for` instead of `with` (e.g. `forContainer`).
* This PR also **changes** two **unreleased** method names in `ListNetworksParam`

### DIFF
--- a/src/main/java/com/spotify/docker/client/DockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DockerClient.java
@@ -2510,6 +2510,18 @@ public interface DockerClient extends Closeable {
      * Show events with a label value.
      *
      * @param label The label to filter on
+     * @return EventsParam
+     * @see #withLabel(String)
+     * @since API 1.21
+     */
+    public static EventsParam label(final String label) {
+      return label(label, null);
+    }
+    
+    /**
+     * Show events with a label value.
+     *
+     * @param label The label to filter on
      * @param value The value of the label
      * @return EventsParam
      * @see #label(String, String)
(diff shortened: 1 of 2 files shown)


albers:
@johnflavin @mattnworb PTAL

davidxia:
I prefer not having aliases but just one method for each. As a user, I'd be confused whether similarly named methods are actually different. I also vote for no breaking changes and using `@Deprecated` if we have to remove something.

albers:
@davidxia This actually leaves no space for improving naming consistency here. Just renaming the methods would be a breaking change, and a transition phase (with deprecating the old names) would use aliases.

mattnworb:
thanks @albers for the PR and taking the effort to improve this. 

I think I would prefer to have aliases and deprecate any existing/confusingly-named methods rather than make a breaking change. I'd be comfortable with making a breaking change and bumping the version number but only if we were to tackle this across the entire library, so we could release a new major version where we tried to address all inconsistencies.

@davidxia as @albers mentioned it seems like we have to do one or the other - can you clarify what you meant?

albers:
@mattnworb To be fair, there is a third alternative: leave everything as-is. There is no pressing need to improve consistency here. If you prefer to go this way, I'd be fine with closing this PR.

albers:
Looks like we don't reach a consensus here, so I'm closing this.

PR #721, opened 2017-04-11
Upload 1.13.1 artifact to sonatype

davidxia:


davidxia:
yea there's no difference. This is just for fun.

davidxia:
better bookkeeping, rather

PR #729, opened 2017-04-25
Support more swarm operations

davidxia:
init, join, leave, update, unlockKey, unlock

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -60,6 +60,7 @@ import static com.spotify.docker.client.messages.swarm.RestartPolicy.RESTART_POL
 import static java.lang.Long.toHexString;
 import static java.lang.String.format;
 import static java.lang.System.getenv;
+import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonList;
 import static org.apache.commons.lang.StringUtils.containsIgnoreCase;
 import static org.awaitility.Awaitility.await;
@@ -83,7 +84,6 @@ import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.isEmptyOrNullString;
-import static org.hamcrest.Matchers.isEmptyString;
 import static org.hamcrest.Matchers.isIn;
 import static org.hamcrest.Matchers.lessThan;
 import static org.hamcrest.Matchers.lessThanOrEqualTo;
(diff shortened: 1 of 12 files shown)


tiagoalves83:
@davidxia When this will be released ? Maybe you could add https://github.com/spotify/docker-client/pull/733 ?

davidxia:
I'm still adding tests and sure thing.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4672,6 +4682,104 @@ public void testInspectSwarm() throws Exception {
...
+    assertThat(updatedSwarm.id(), equalTo(swarm.id()));
+    final SwarmSpec newSpec = updatedSwarm.swarmSpec();
+    assertThat(newSpec, equalTo(updatedSpec));
+
+    // Return swarm back to old settings
+    sut.updateSwarm(updatedSwarm.version().index(), swarm.swarmSpec());
+  }
+
+  @Test
+  public void testUnlockKey() throws Exception {
```
davidxia: I'm not sure how to test [unlocking the manager with the key](https://docs.docker.com/engine/swarm/swarm_manager_locking/). I can initialize the swarm with autolocking enabled, but how do I lock the manager? By restarting docker somehow?

tiagoalves83: Maybe:
```
Runtime.getRuntime().exec("docker-machine create -d virtualbox m1");
Runtime.getRuntime().exec("docker $(docker-machine config m1) swarm init --autolock");
// Grab the key somehow ...
Runtime.getRuntime().exec("docker-machine ssh m1 sudo /etc/init.d/docker restart");
// Build DockerClient ...
client.unlockSwarm(key);
```

It's ugly, but I believe this is testable.

davidxia: 😬

tiagoalves83: Restart docker from within java using "Runtime.getRuntime().exec()"

davidxia:
@mattnworb This one is also ready for review.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
mattnworb: this seems like the perfect thing to test with a unit test. We should really be prioritizing testing how docker-client interacts with the Docker Remote API rather than the integration test of `(how docker-client interacts with Docker Remote API + how docker daemon behaves/if it implements that API)`.

for instance in a unit test you could just check that the client sends HTTP requests that are correct according to the API spec. We should really not care so much that the daemon also implements the API spec as advertised.

davidxia:
Agreed. I'm going to leave that for another refactoring effort though :)

mattnworb:
I think the easiest way to start going down that path is to add unit tests (instead of, or in addition to, integration tests) in new PRs and functionality, rather than attempting a big-bang refactoring of test code some day in the future.

PR #732, opened 2017-04-29
Validate HostConfig.extraHosts

### LINKED ISSUE #699: extraHosts(): check parameters to shield from docker daemon panic
With docker version `1.12.6, build 78d1802` and docker-client 8.1.3-SNAPSHOT from ~1 week ago, I ran into [docker issue #20997](https://github.com/docker/docker/issues/20997) when supplying wrong parameters to the `hostConfig`'s `extraHosts` parameter.

E.g., when wrongly passing in `extraHosts("host", "alias")` instead of `...("host:alias")`, as I mistakenly did, I'd prefer to get an IllegalArgumentException because the host alias is not properly formatted.

Although it is a docker daemon bug (producing an `out of bounds` error instead of returning the proper cause (illegal argument), it would save devs some time when the library could check the parameters first.

davidxia:
Extra hosts args must contain ":".

fixes #699

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/HostConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/HostConfig.java
@@ -22,6 +22,7 @@ package com.spotify.docker.client.messages;
 
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.NONE;
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Strings.isNullOrEmpty;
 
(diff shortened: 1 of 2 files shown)


davidxia:
cc @grexe

davidxia:
@mattnworb Did you get a chance to take a look?

src/main/java/com/spotify/docker/client/messages/HostConfig.java
```diff
@@ -627,7 +628,22 @@ public Builder hostPidMode() {
     
     public abstract Builder storageOpt(Map<String, String> tmpfs);
 
-    public abstract HostConfig build();
+    abstract HostConfig autoBuild();
```
mattnworb: should leave a comment about why this exists, it isn't very obvious

src/main/java/com/spotify/docker/client/messages/HostConfig.java
```diff
@@ -627,7 +628,22 @@ public Builder hostPidMode() {
...
+      validateExtraHosts(hostConfig.extraHosts());
+      return hostConfig;
+    }
+  }
+
+  private static void validateExtraHosts(final List<String> extraHosts) {
+    if (extraHosts != null) {
+      for (final String extraHost : extraHosts) {
+        checkArgument(extraHost.contains(":"),
+            String.format("extra host arg '%s' must contain a ':'", extraHost));
```
mattnworb: no need to call String.format yourself, there is an overload of checkArgument that does:

```
public static void checkArgument(boolean expression,
                 @Nullable
                 String errorMessageTemplate,
                 @Nullable
                 Object... errorMessageArgs)
```

davidxia: thanks. Didn't notice this.

src/main/java/com/spotify/docker/client/messages/HostConfig.java
davidxia: Done.

davidxia:
Addressed both comments. How's it look now?

PR #734, opened 2017-05-03
Fix some frequent travis test failures

davidxia:
Travis builds often fail because the docker daemon
isn't running after we restart it. Sleep to give it some
time to restart.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -43,6 +43,7 @@ case "$1" in
     # restart the service for the /etc/default/docker change we made after
     # installing the package
     sudo restart docker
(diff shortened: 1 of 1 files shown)


.travis.sh
```diff
@@ -43,6 +43,7 @@ case "$1" in
     # restart the service for the /etc/default/docker change we made after
     # installing the package
     sudo restart docker
+    sleep 5
```
davidxia: I should actually do `docker ps` in a loop up to some time out to actually ensure docker is ready.

PR #736, opened 2017-05-03
[travis] Fix a frequent test failure

davidxia:
caused by docker not being ready yet.

Travis builds often fail because the docker daemon
isn't running after we restart it. Sleep to give it some
time to restart.


davidxia:
hmm yea you're right. I'll just `sleep 10` then.

mattnworb:
I mean, both are fine 😄

Issue #717, opened 2017-04-11
No changelog?


ketan:
I don't see the changelog file being maintained which makes it really difficult to when upgrading the library to see what has changed between versions.

Could I ask the maintainers to keep the CHANGELOG up-to-date, preferably when merging PR, if it makes it easy to maintain the file better?

Thanks

davidxia:
Hey, sorry about this. We can definitely keep the changelog up to date going forward.

mattnworb:
We'll make sure that new PRs keep the changelog up to date.

PR #741, opened 2017-05-09
remove DockerClient methods with unused RegistryAuth parameters

mattnworb:
There are several overloads of the `load()` and `save()` methods in the
`DockerClient` interface that has a `RegistryAuth` parameter that is
never used in the DefaultDockerClient implementation.

This change removes those methods as they don't serve any purpose (and
have been marked as deprecated for a while).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 8.5.0
+
+### Removal of deprecated methods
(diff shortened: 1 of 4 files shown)


davidxia:
Can you add a note to the CHANGELOG along with bumping the minor or major version saying these are removed? See https://github.com/spotify/docker-client/issues/717

mattnworb:
@davidxia I've added a new commit

PR #749, opened 2017-05-17
Adds support for RegistryAuthSupplier

mavenraven:
Note that the integration test I added currently isn't running in CI.


mavenraven:
@johnflavin I removed the event tests you introduced around here https://github.com/spotify/docker-client/commit/06ef7c844c4c7e65b86b67d4635ea43953c1752c since they keep breaking in CI. Any idea on why they're flaking?

src/main/java/com/spotify/docker/client/NoOpRegistryAuthSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+
+package com.spotify.docker.client;
+
+import com.spotify.docker.client.exceptions.DockerException;
+import com.spotify.docker.client.messages.RegistryAuth;
+import com.spotify.docker.client.messages.RegistryAuthSupplier;
+
+/**
+ * Created by dmorhovich on 5/15/17.
+ */
```
mattnworb: Would prefer if this had at least a one line summary of what the class did - we don't need to have authors and dates in the file headers as we get that from Git.

src/main/java/com/spotify/docker/client/messages/DockerConfigReader.java
```diff
@@ -0,0 +1,117 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DockerConfigReader {
+  private static final Logger LOG = LoggerFactory.getLogger(DockerConfigReader.class);
+
+  private static final ObjectMapper MAPPER = ObjectMapperProvider.objectMapper();
+
+
+  public RegistryAuth fromComfig(Path configPath, String serverAddress) throws IOException {
```
mattnworb: typo: `fromConfig`?

src/main/java/com/spotify/docker/client/messages/GCloudProcess.java
```diff
@@ -0,0 +1,31 @@
...
+ * -/-/-
+ */
+
+package com.spotify.docker.client.messages;
+
+import java.io.IOException;
+
+public class GCloudProcess {
+
+  public Process runGcloudDocker() throws IOException {
```
mattnworb: if there is no state, this should probably be a static method

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -335,7 +336,7 @@ public ClientBuilder get() {
 
   private final URI uri;
   private final String apiVersion;
-  private final RegistryAuth registryAuth;
+  private RegistryAuthSupplier registryAuthSupplier;
```
mattnworb: should be `final`

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -414,7 +415,11 @@ protected DefaultDockerClient(final Builder builder) {
         .property(ApacheClientProperties.CONNECTION_MANAGER, cm)
         .property(ApacheClientProperties.REQUEST_CONFIG, requestConfig);
 
-    this.registryAuth = builder.registryAuth;
+    this.registryAuthSupplier = builder.registryAuthSupplier;
+
+    if (this.registryAuthSupplier == null) {
+      this.registryAuthSupplier = new NoOpRegistryAuthSupplier(null);
```
mattnworb: would be better to have a no-arg constructor to NoOpRegistryAuthSupplier that can internally set the field to null, rather than the caller

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1363,6 +1367,7 @@ public String build(final Path directory, final String name, final String docker
 
     WebTarget resource = noTimeoutResource().path("build");
 
+    final RegistryAuth registryAuth = registryAuthSupplier.authFor(name);
```
mattnworb: I don't think this behavior is correct. When building an image, you are supposed to send all the RegistryAuths that the client is configured for: https://docs.docker.com/engine/api/v1.28/#operation/ImageBuild

It might be that the image you are building `FROM` has different RegistryAuth details than the name of the image you are building

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1798,8 +1803,7 @@ public void unlock(final UnlockKey unlockKey) throws DockerException, Interrupte
   @Override
   public ServiceCreateResponse createService(ServiceSpec spec)
       throws DockerException, InterruptedException {
-
-    return createService(spec, registryAuth);
+    return createService(spec, registryAuthSupplier.authFor(""));
```
mattnworb: it is not documented in the authFor method that you can pass special String values to get special RegistryAuth instances - why not just have a 2nd method in the interface for this use case?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2663,11 +2670,23 @@ public RegistryAuth registryAuth() {
...
+
+    public Builder registryAuthSupplier(final RegistryAuthSupplier registryAuthSupplier) {
+      this.registryAuthCSupplierCalled = true;
+      this.registryAuthSupplier = registryAuthSupplier;
       return this;
     }
 
     public DefaultDockerClient build() {
+      if (registryAuthCalled && registryAuthCSupplierCalled) {
+        throw new RuntimeException("LOGIC ERROR: DefaultDockerClient does not support being built "
```
mattnworb: also I think you could avoid the need for these two booleans by moving this check into the `registryAuth(RegistryAuth)` and `registryAuthSupplier(RegistryAuthSupplier)` methods - you can just throw an exception of `this.registryAuthSupplier` is already set

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2663,11 +2670,23 @@ public RegistryAuth registryAuth() {
      * @return Builder
      */
     public Builder registryAuth(final RegistryAuth registryAuth) {
```
mattnworb: mark as `@Deprecated`?

src/main/java/com/spotify/docker/client/messages/RegistryAuth.java
```diff
@@ -102,7 +96,8 @@ public final String toString() {
    */
   @SuppressWarnings("unused")
   public static Builder fromDockerConfig() throws IOException {
```
mattnworb: can we make this `@Deprecated` if we would like for people to stop using it and to use the RegistryAuthSupplier instead?

src/main/java/com/spotify/docker/client/messages/GoogleContainerRegistryAuthSupplier.java
```diff
@@ -0,0 +1,58 @@
...
+                                             Path configPath) {
+    this.dockerCfgReader = dockerCfgReader;
+    this.googleContainerRegistryCredRefresher = googleContainerRegistryCredRefresher;
+    this.configPath = configPath;
+  }
+
+  @Override
+  public RegistryAuth authFor(String imageName) throws DockerException {
+    try {
+      String registryName = "https://" + imageName.split("/")[0];
```
mattnworb: what would happen if I configure my DefaultDockerClient to use GoogleContainerRegistryAuthSupplier but then the client tries to pull an image from a non-GCR source? This doesn't seem to check to see if the image looks to be gcr.io related.

src/main/java/com/spotify/docker/client/messages/GoogleContainerRegistryCredRefresher.java
```diff
@@ -0,0 +1,46 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.docker.client.messages;
```
mattnworb: this is the wrong package for this class - `messages` is supposed to be POJOs related to messages returned from/sent to the API

src/main/java/com/spotify/docker/client/messages/DockerConfigReader.java
```diff
@@ -0,0 +1,117 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.docker.client.messages;
```
mattnworb: please put in a different package

src/main/java/com/spotify/docker/client/messages/GCloudProcess.java
```diff
@@ -0,0 +1,31 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.docker.client.messages;
```
mattnworb: please put in a different package

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
mavenraven: It can't be final, at least how I wrote it with the null check

mattnworb:
```java
this.registryAuthSupplier = builder.registryAuthSupplier != null ? builder.registryAuthSupplier : new NoOpRegistryAuthSupplier(null);
```

mattnworb: you just need to check the `builder.registryAuthSupplier` against null, not `this.registryAuthSupplier`

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
mavenraven: done

src/main/java/com/spotify/docker/client/NoOpRegistryAuthSupplier.java
mavenraven: yeah, that's an intellij thing, deleted

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
mavenraven: got it

src/main/java/com/spotify/docker/client/messages/DockerConfigReader.java
mavenraven: done

src/main/java/com/spotify/docker/client/messages/DockerConfigReader.java
mavenraven: fixed

src/main/java/com/spotify/docker/client/messages/GCloudProcess.java
mavenraven: done

src/main/java/com/spotify/docker/client/messages/GCloudProcess.java
mavenraven: needs to be virtual so I can mock it via mockito

src/main/java/com/spotify/docker/client/messages/GoogleContainerRegistryCredRefresher.java
mavenraven: moved

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
mavenraven: Are we missing a test? Also, from my understanding, RegistryAuth only represents creds for one registry, not multiple, so I'm not sure how that would work on master right now

src/main/java/com/spotify/docker/client/messages/GoogleContainerRegistryAuthSupplier.java
mavenraven: This could be needed, but I'd argue to make this a separate story/pr since it's adding a feature to this workflow.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
mavenraven: done

src/main/java/com/spotify/docker/client/messages/RegistryAuth.java
mavenraven: done

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
mavenraven: done

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2536,7 +2540,10 @@ public static Builder fromEnv() throws DockerCertificateException {
     private DockerCertificatesStore dockerCertificatesStore;
     private boolean dockerAuth;
     private RegistryAuth registryAuth;
+    private RegistryAuthSupplier registryAuthSupplier;
     private Map<String, Object> headers = new HashMap<>();
+    private boolean registryAuthCalled;
+    private boolean registryAuthCSupplierCalled;
```
mavenraven: variable removed

mavenraven: variable removed

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
mavenraven: changed to add a new method to the RegistryAuthSupplier interface

src/main/java/com/spotify/docker/client/DockerConfigReader.java
```diff
@@ -0,0 +1,117 @@
...
+  public RegistryAuth.Builder parseDockerConfig(final Path configPath, String serverAddress)
+      throws IOException {
+    checkNotNull(configPath);
+    final RegistryAuth.Builder authBuilder = RegistryAuth.builder();
+    final JsonNode authJson = this.extractAuthJson(configPath);
+
+    if (isNullOrEmpty(serverAddress)) {
+      final Iterator<String> servers = authJson.fieldNames();
+      if (servers.hasNext()) {
+        serverAddress = servers.next();
```
davidxia: This will return the first credential in the file. We need the one, if any, that equals `serverAddress`.

src/main/java/com/spotify/docker/client/gcr/GCloudProcess.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.spotify.docker.client.gcr;
+
+import java.io.IOException;
+
+public class GCloudProcess {
+
+  public Process runGcloudDocker() throws IOException {
+    return Runtime.getRuntime().exec("gcloud docker -a");
```
davidxia: nice

src/main/java/com/spotify/docker/client/DockerConfigReader.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    if (serverAuth.has("email")) {
+      authBuilder.email(serverAuth.get("email").asText());
+    }
+
+    return authBuilder;
+  }
+
+  public Path defaultConfigPath() {
```
davidxia: could be static.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
davidxia: keep doing this!

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1179,17 +1184,16 @@ public InputStream saveMultiple(final String... images)
...
+      resource.queryParam("names", urlEncode(image));
     }
 
     return request(
         GET,
         InputStream.class,
         resource,
-        resource.request(APPLICATION_JSON_TYPE).header("X-Registry-Auth", authHeader(registryAuth))
+        resource.request(APPLICATION_JSON_TYPE).header("X-Registry-Auth", authHeader(
+            registryAuthSupplier.authFor(images[0])))
```
davidxia: This uses auth for first image. Can you export/save images from different registries at the same time? I couldn't find any docs for this.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2661,9 +2670,24 @@ public RegistryAuth registryAuth() {
...
+      if (this.registryAuthSupplier != null) {
+        throw new IllegalStateException(ERROR_MESSAGE);
+      }
       this.registryAuth = registryAuth;
+      this.registryAuthSupplier = new NoOpRegistryAuthSupplier(registryAuth);
+      return this;
+    }
+
+    public Builder registryAuthSupplier(final RegistryAuthSupplier registryAuthSupplier) {
+      if (this.registryAuthSupplier != null) {
```
davidxia: Delete these three lines to allow calling this method twice?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2661,9 +2670,24 @@ public RegistryAuth registryAuth() {
...
+     *
+     * @deprecated in favor of registryAuthSupplier
      */
+    @Deprecated
     public Builder registryAuth(final RegistryAuth registryAuth) {
+      if (this.registryAuthSupplier != null) {
+        throw new IllegalStateException(ERROR_MESSAGE);
+      }
       this.registryAuth = registryAuth;
+      this.registryAuthSupplier = new NoOpRegistryAuthSupplier(registryAuth);
```
davidxia: This line makes calling this method twice throw which might not be obvious to caller if it doesn't know about the implementation here.

src/test/java/com/spotify/docker/client/gcr/GoogleContainerRegistryAuthSupplierTest.java
```diff
@@ -0,0 +1,82 @@
...
+        new GoogleContainerRegistryAuthSupplier(
+            dockerCfgReader, googleContainerRegistryCredRefresher, path);
+
+    googleContainerRegistryAuthSupplier.authFor("us.gcr.io/awesome-project/example-image");
+    inOrder.verify(googleContainerRegistryCredRefresher).refresh();
+    inOrder.verify(dockerCfgReader).fromComfig(any(Path.class), anyString());
+  }
+
+  @Test
+  public void authForReturnsRegisteryAuthThatMatchesRegisteryName() throws Exception {
```
davidxia: type in `Registery`

src/main/java/com/spotify/docker/client/DockerConfigReader.java
davidxia: Let's test this method with a test fixture/file with multiple serverAddresses.

src/test/java/com/spotify/docker/client/gcr/GoogleContainerRegistryCredRefresherTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.rules.ExpectedException;
+
+public class GoogleContainerRegistryCredRefresherTest {
+
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+
+  @Test
+  public void refreshShellsOutToGCloudCli() throws IOException, InterruptedException {
```
davidxia: I'd be ok not having this test :)

src/test/java/com/spotify/docker/client/gcr/GoogleContainerRegistryCredRefresherTest.java
```diff
@@ -0,0 +1,80 @@
...
+    when(failedProc.getErrorStream())
+        .thenReturn(
+            new ByteArrayInputStream("ERROR: (gcloud.docker)".getBytes(StandardCharsets.UTF_8)));
+
+    when(gcloudProcess.runGcloudDocker()).thenReturn(failedProc);
+
+    GoogleContainerRegistryCredRefresher googleContainerRegistryCredRefresher =
+        new GoogleContainerRegistryCredRefresher(gcloudProcess);
+
+    googleContainerRegistryCredRefresher.refresh();;
```
davidxia: extra semicolon

src/main/java/com/spotify/docker/client/gcr/GCloudProcess.java
mattnworb: I think it might be better to run this like `gcloud --account <account name> docker ...` so that the user can specify which account the credentials come from. Otherwise this is forcing that the "active account" in the user's/host's gcloud installation is the one to use for docker credentials. I think it would be worth decoupling those.

src/main/java/com/spotify/docker/client/DockerConfigReader.java
```diff
@@ -0,0 +1,117 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DockerConfigReader {
+  private static final Logger LOG = LoggerFactory.getLogger(DockerConfigReader.class);
+
+  private static final ObjectMapper MAPPER = ObjectMapperProvider.objectMapper();
+
+
+  public RegistryAuth fromComfig(Path configPath, String serverAddress) throws IOException {
```
davidxia: typo in `Comfig`

mattnworb:
I think that my test refactoring in #753 will make it much simpler to test what auth headers are sent to the docker API by this change

johnflavin:
I don't remember seeing the tests from #549 failing on travis. They do fail on my local machine sometimes due to the clock on docker's vm being out of sync with the clock on the host machine. Can you link to a travis build where they failed?

mattnworb:
@johnflavin I think this is an example of a failure, for a separate PR: https://travis-ci.org/spotify/docker-client/jobs/234609588. I don't have any data to support this but it seems to have started failing a lot more often just recently, but then again I've only been trying to get some PRs of my own merged recently so I may just not have noticed.

@mavenraven I think it would make sense to pull the commits about CircleCI and the flaky test into a separate PR(s); that way we could merge them and benefit from them earlier without having to wait for this PR to be finished.

mattnworb:
continued in #759

PR #744, opened 2017-05-16
#446 make default docker client thread safe

unicolet:
Fixes #446 

The problem was that `DEFAULT_CONFIG` in `DefaultDockerClient.java` was not thread safe. The rest is just Intellij *optimizations* and a test to verify the fix.

If you merge this **PLEASE** release a new version of maven-docker-client with the fix, which is the only reason for this PR to exist in the first place.

PR #752, opened 2017-05-19
ConnectionPoolTest: avoid tag operations

mattnworb:
Remove the tag operations from ConnectionPoolTest as they fail
consistently on earlier versions of Docker (1.9 at least) due to the
new-tag-name already existing.

The behavior around `docker tag` was changed in 1.10 and the `-force/-f`
flag was completely removed:
https://docs.docker.com/engine/deprecated/#old-command-line-options


mattnworb:
this was introduced in #744; I merged the PR thinking that the builds failed due to flakey tests but some of the builds (for 1.9 at least) probably failed because of this.

PR #753, opened 2017-05-20
simplify DefaultDockerClientUnitTest

mattnworb:
refactor this test to use a MockWebServer rather than mocking the
Jersey/jax-rs Client interface used by the DefaultDockerClient.

Using a (Mockito) mock Client is problematic for a test since it has
such a rich and fluent interface - the test needs to know exactly which
methods are called on the Client.

Ultimately, this is not what a unit test of DefaultDockerClient should
care about - a good test would only care about what HTTP requests it
sends to the Docker Remote API and how it behaves when it gets certain
HTTP responses back.

The MockWebServer allows us to control the HTTP responses and capture
the HTTP requests much easier than using a Mockito mock of the
particular HTTP client library used by DefaultDockerClient.

With this approach, it would be possible to one day change
DefaultDockerClient to use a different HTTP client library without
having to change one line of this test. The same cannot be said of using
Mockito.

This change was initially driven by a desire to make PR #749 easier to
test, to check that it sends expected authentication headers, but I
think it will be useful and help improve testing in docker-client across
the board.

https://github.com/square/okhttp/tree/master/mockwebserver

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
@@ -20,9 +20,6 @@
 
 package com.spotify.docker.client;
 
-import static com.spotify.hamcrest.jackson.IsJsonArray.jsonArray;
-import static com.spotify.hamcrest.jackson.IsJsonObject.jsonObject;
-import static com.spotify.hamcrest.jackson.IsJsonText.jsonText;
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.equalTo;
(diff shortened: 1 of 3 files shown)


mattnworb:
with this, the unit test becomes much simpler to read, with less code, and is also testing more of what we actually care about - HTTP request/response content - instead of which specific methods are called on the jax-rs Client.

cc @johnflavin

mattnworb:
@davidxia @mavenraven

src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
```diff
@@ -137,62 +106,40 @@ public void testHostForIpHttps() {
...
-    verify(builderMock, times(1)).header(keyArgument.capture(), valueArgument.capture());
+    final RecordedRequest recordedRequest = takeRequestImmediately();
+    assertThat(recordedRequest.getMethod(), is("GET"));
+    assertThat(recordedRequest.getPath(), is("/info"));
 
-    Assert.assertEquals("foo", keyArgument.getValue());
-    Assert.assertEquals(1, valueArgument.getValue());
+    assertThat(recordedRequest.getHeader("int"), is("1"));
+    assertThat(recordedRequest.getHeader("string"), is("2"));
+    // TODO (mbrown): this seems like incorrect behavior - the client should send 3 headers with
```
davidxia: Is this an issue with our code or upstream?

mattnworb: it is an existing bug in this Builder's "Add header" method which I didn't want to fix as a part of this, but I think should be fixed at some point

PR #754, opened 2017-05-21
Swarm info v2

tiagoalves83:
Docker "Info" has a Swarm field with multiple information about a SwarmCluster. 
Could not found that info in https://docs.docker.com/engine/api/v1.26/#operation/SystemInfo but the field exists, believe me! :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,7 @@ pom.xml.versionsBackup
 .DS_Store
 dependency-reduced-pom.xml
 .checkstyle
+/.apt_generated/
+/lib/
+/.factorypath
+/.gitignore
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -9,3 +9,7 @@ pom.xml.versionsBackup
 .DS_Store
 dependency-reduced-pom.xml
 .checkstyle
+/.apt_generated/
```
davidxia: Are these intentional? Maybe it's better not to include in this PR.

davidxia:
closed in favor of https://github.com/spotify/docker-client/pull/814

PR #758, opened 2017-05-22
Testing 1s sleep to clean out event stream

johnflavin:
Attempt to fix flaky test pointed out in #749.


mattnworb:
Seems to help, although just one run might not be enough data - the failure is for a different test.

mattnworb:
on second thought, this can't hurt so 👍

PR #760, opened 2017-05-23
add note pointing contributors to DefaultDockerClientUnitTest

mattnworb:
Document that we would prefer to have more unit tests than integration
tests for new functionality going forward, now that
DefaultDockerClientUnitTest uses a mock web server where we can capture
and stub HTTP requests/responses.


mattnworb:
@davidxia

PR #762, opened 2017-05-24
use google-auth-library-oauth2-http for fetching access tokens

mattnworb:
Instead of invoking `gcloud docker` and parsing a docker config file, this
adds a new ContainerRegistryAuthSupplier which programmatically fetches
access tokens for a Google service or user account loaded for a credentials
file.

The benefit of this implementation is that any arbitrary account can be
used without having to be registered/added to the `gcloud` CLI - the
supplier just needs to be pointed at a file containing the credentials for
the account to be used. There is also no need for gcloud to be available
on the helios agent.

This support should probably be pulled into a new library at some point,
something like `docker-client-gcr`, so that non-GCR users of
docker-client don't have to have this additional logic pulled in. For
now, the Google library needed for GCR support is marked as `optional`
in the pom.xml so users have to make sure to add the dependency
themselves.


mattnworb:
@davidxia @mavenraven @gimaker @caipre 

one other thing I want to add here is another static method like `ContainerRegistryAuthSupport.forApplicationDefault()` that will load the [application default credentials](https://developers.google.com/identity/protocols/application-default-credentials). The current method `fromStream` loads credentials from a given file path to specify what account is used, but this would add a way to use the default credentials, for example on a GCE instance

mattnworb:
also I think once something like `ContainerRegistryAuthSupport.forApplicationDefaultCredentials()` is loaded, then the existing (already merged) implementation that invokes `gcloud docker -a` won't be needed anymore and can be removed.

mattnworb:
it feels so great to be able to get green status on commits without having to rerun individual travis jobs 20 times ✅  😄

mattnworb:
with the two new commits above I believe this is complete and ready for a full review

davidxia:
🎉

src/main/java/com/spotify/docker/client/gcr/ContainerRegistryAuthSupplier.java
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  private final GoogleCredentials credentials;
+  // TODO (mbrown): change to java.time.Clock once on Java 8
+  private final Clock clock;
+  private final long minimumExpiryMillis;
+  private final CredentialRefresher credentialRefresher;
+
+  @VisibleForTesting
+  ContainerRegistryAuthSupplier(
```
mavenraven: Do we need to expose this at all vs. using a Builder in the tests?

src/main/java/com/spotify/docker/client/gcr/ContainerRegistryAuthSupplier.java
```diff
@@ -0,0 +1,251 @@
...
+    this.credentialRefresher = credentialRefresher;
+  }
+
+  /**
+   * Get an accessToken to use, possibly refreshing the token if it expires within the
+   * minimumExpiryMillis.
+   */
+  private AccessToken getAccessToken() throws DockerException {
+    // synchronize attempts to refresh the accessToken
+    synchronized (credentials) {
```
mavenraven: We're locking on an object that's exposed outside of ContainerRegistryAuthSupplier, which could lead to deadlock. Ideally, we don't need to need to mutate the credentials object at, and just get a new one, but that might not possible with the API.

src/main/java/com/spotify/docker/client/gcr/ContainerRegistryAuthSupplier.java
mattnworb: yeah, so the tests can pass in a mock CredentialRefresher - someone using a builder will never need to do that

mavenraven: I guess my question is, we also have this in the builder https://github.com/spotify/docker-client/pull/762/files/46e5fafb70d59ead941a02afb1a39284550116d1#diff-980dfc9639d40441a5814482f27ccba5R126 . Seems like we should either being using only the contructor or only the optional builder methods in the tests.

mattnworb: ah yea, fair point. That withClock method doesn't need to be there

src/main/java/com/spotify/docker/client/gcr/ContainerRegistryAuthSupplier.java
mattnworb: I don't think this will be an issue as I think when we call credentials.createScoped() we get a new instance, and only this class holds that instance. 

I don't like synchronizing on what seems like a shared object either but wasn't sure what else to use here to avoid redundant calls to refresh if there are concurrent calls to authFor - would be open to any suggestions you might have.

pom.xml
```diff
@@ -169,6 +169,19 @@
       <version>3.0.1</version>
       <scope>provided</scope>
     </dependency>
+    <!-- TODO we should pull out the Google Cloud support to a new library -->
```
davidxia: @mattnworb Just curious. What did you have in mind for this?

mattnworb: see these two comments:
https://github.com/spotify/dockerfile-maven/issues/54#issuecomment-319155527
https://github.com/spotify/dockerfile-maven/issues/54#issuecomment-319764523

PR #759, opened 2017-05-22
add RegistryAuthSupplier interface

mattnworb:
This completes #749 - David and I worked on this together in-person.

The new `RegistryAuthSupplier` interface replaces the existing `RegistryAuth` support in DefaultDockerClient so that users can customize the authentication behavior much more closely. On each operation that requires authentication, DefaultDockerClient will invoke the appropriate method on the RegistryAuthSupplier instance, allowing the implementation to supply up-to-date authentication information/tokens.

This change is inspired by our attempts to integrate with Google Container Registry, where `~/.docker/config.json` file contains short-lived access tokens that expire shortly after creation. This means that a long-lived DefaultDockerClient instance cannot store static RegistryAuth information from construction-time.

Also included is an implementation of RegistryAuthSupplier that refreshes the access tokens supplied by `gcloud docker -a` prior to each `authFor(..)` operation, as well as an implementation of RegistryAuthSupplier that returns static data.

Closes #740.

PR #765, opened 2017-05-25
RegistryAuth: handle case where password contains colon

### LINKED ISSUE #764: Auth fails when .docker/config.json base64 decoded password contains a ':' character
## Description
RegistryAuth.forAuthToken() fails to return a correctly populated auth object if the password contains a ':' (colon) character.

## How to reproduce
Use a registry auth file that has a ':' in the password.

## What do you expect
Should be able to authenticate with a registry correctly.

## What happened instead
Authentication fails.

## Software:

- `docker version`: Docker version 17.03.1-ce, build c6d412e
- Spotify's docker-client version: current master 

## Full backtrace
N/A

mattnworb:
Fixes #764. This now matches [how docker-cli parses this field][0] from
~/.docker/config.json.

[0]: https://github.com/docker/cli/blob/db5620026dda9b8f5da519e382cf40ed22775e6d/cli/config/configfile/file.go#L184-L189

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/RegistryAuth.java
+++ b/src/main/java/com/spotify/docker/client/messages/RegistryAuth.java
@@ -161,11 +161,13 @@ public abstract class RegistryAuth {
 
   /** Construct a Builder based upon the "auth" field of the docker client config file. */
   public static Builder forAuth(String auth) {
-    final String[] authParams = Base64.decodeAsString(auth).split(":");
+    // split with limit=2 to catch case where password contains a colon
+    final String[] authParams = Base64.decodeAsString(auth).split(":", 2);
 
     if (authParams.length != 2) {
       return builder();
     }
+
     return builder()
         .username(authParams[0].trim())
         .password(authParams[1].trim());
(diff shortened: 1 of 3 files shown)


mavenraven:
@mattnworb what if the password contains 2 colons?

mattnworb:
@mavenraven it will be taken care of by [String.split(String, int)](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#split-java.lang.String-int-):

> The limit parameter controls the number of times the pattern is applied and therefore affects the length of the resulting array. If the limit n is greater than zero then the pattern will be applied at most n - 1 times, the array's length will be no greater than n, and the array's last entry will contain all input beyond the last matched delimiter. If n is non-positive then the pattern will be applied as many times as possible and the array can have any length. If n is zero then the pattern will be applied as many times as possible, the array can have any length, and trailing empty strings will be discarded.

mattnworb:
btw seems like this bug existed before #759: https://github.com/spotify/docker-client/blob/55b993dae556bd59bfa433d9ec9a6030e4123d83/src/main/java/com/spotify/docker/client/messages/RegistryAuth.java#L189-L194

PR #767, opened 2017-05-26
GCR: default to read_write scope instead of read_only

mattnworb:
`read_only` seems like too restrictive of a default for docker-client to use. Also add more javadocs about how users can change this, and how to restrict applications to pull access only.

Also, add forCredentials() factory method to ContainerRegistryAuthSupplier as requested in https://github.com/spotify/helios/pull/1120

Finally, remove the withClock method from the Builder, as a user will never need to change this.


mattnworb:
@gimaker @davidxia

gimaker:
:+1:

PR #771, opened 2017-05-31
Issue 742: Enable http proxy support for docker-client

### LINKED ISSUE #742: Enable Proxy support
Currently `com.spotify.docker.client.DefaultDockerClient` does not provide a way to configure http proxy.
We could add something on the following lines so that the internal jaxws.rs client has proxy configured.
```
final ClientConfig config = DEFAULT_CONFIG
        .connectorProvider(new ApacheConnectorProvider())
        .property(ApacheClientProperties.CONNECTION_MANAGER, cm)
        .property(ApacheClientProperties.REQUEST_CONFIG, requestConfig)
        .property(ClientProperties.PROXY_URI, proxy);
        .property(ClientProperties.PROXY_USERNAME,user);
        .property(ClientProperties.PROXY_PASSWORD,pass);
```

shrids:
**Change log description**
Enable http proxy support for docker-client. 

**Purpose of the change**
Fixes #742 .

**What the code does**
- ensure `org.glassfish.jersey.client.ClientProperties#PROXY_URI` is updated with the proxy URI.
- ensure `org.glassfish.jersey.client.ClientProperties#REQUEST_ENTITY_PROCESSING` is set to buffered to ensure that the request's Content-Length is populated while sending the request via a proxy. 

**How to verify it**
Tested it by using https://github.com/tinyproxy/tinyproxy as  http proxy. Version of tinyproxy: 1.8.3


src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -426,6 +427,18 @@ protected DefaultDockerClient(final Builder builder) {
     this.headers = new HashMap<>(builder.headers());
   }
 
+  private ClientConfig updateProxy(ClientConfig config) {
+    final String proxyHost = System.getProperty("http.proxyHost");
+    if (proxyHost != null) {
+      config.property(ClientProperties.PROXY_URI, proxyHost + ":"
+              + System.getProperty("http.proxyPort", "8080"));
```
mattnworb: if you have already checked that this property exists, is there any point in passing the second argument?

mattnworb: ah sorry I misread that this is accessing a different property. I don't think it is safe to default to 8080 though, that doesn't seem to be a good idea to me.

shrids: Thanks for the review comment, removed the default port and used a` Preconditions.checkNotNull ()`to validate `http.proxyPort`.

davidxia:
👍

PR #774, opened 2017-06-02
FEATURE: Added an inspectNode method for the docker swarm API

ssullivan:
I use this library in another project and have need for inspectNode functionality.

I'm still working on tests and cleaning up checkstyle issues for this new feature but wanted to solicit some feedback per the contribution guidelines.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -72,6 +72,7 @@ import com.spotify.docker.client.exceptions.ExecNotFoundException;
 import com.spotify.docker.client.exceptions.ExecStartConflictException;
 import com.spotify.docker.client.exceptions.ImageNotFoundException;
 import com.spotify.docker.client.exceptions.NetworkNotFoundException;
+import com.spotify.docker.client.exceptions.NodeNotFoundException;
 import com.spotify.docker.client.exceptions.NonSwarmNodeException;
 import com.spotify.docker.client.exceptions.NotFoundException;
 import com.spotify.docker.client.exceptions.PermissionException;
@@ -109,6 +110,7 @@ import com.spotify.docker.client.messages.Version;
 import com.spotify.docker.client.messages.Volume;
 import com.spotify.docker.client.messages.VolumeList;
 import com.spotify.docker.client.messages.swarm.Node;
+import com.spotify.docker.client.messages.swarm.NodeInfo;
 import com.spotify.docker.client.messages.swarm.Secret;
 import com.spotify.docker.client.messages.swarm.SecretCreateResponse;
 import com.spotify.docker.client.messages.swarm.SecretSpec;
@@ -1969,6 +1971,27 @@ public class DefaultDockerClient implements DockerClient, Closeable {
 
     WebTarget resource = resource().path("nodes");
     return request(GET, NODE_LIST, resource, resource.request(APPLICATION_JSON_TYPE));
+  }
+
+  @Override
+  public NodeInfo inspectNode(final String nodeId) throws DockerException, InterruptedException {
+    assertApiVersionIsAbove("1.24");
+
+    WebTarget resource = resource().path("nodes")
+        .path(nodeId);
+
+    try {
+      return request(GET, NodeInfo.class, resource, resource.request(APPLICATION_JSON_TYPE));
+    } catch (DockerRequestException e) {
+      switch (e.status()) {
+        case 404:
+          throw new NodeNotFoundException(nodeId);
+        case 503:
+          throw new NonSwarmNodeException("Node " + nodeId + " is not in a swarm", e);
+        default:
+          throw e;
+      }
+    }
   }
 
   @Override
--- a/src/main/java/com/spotify/docker/client/DockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DockerClient.java
@@ -33,6 +33,7 @@ import com.spotify.docker.client.exceptions.DockerException;
 import com.spotify.docker.client.exceptions.ExecNotFoundException;
 import com.spotify.docker.client.exceptions.ImageNotFoundException;
 import com.spotify.docker.client.exceptions.NetworkNotFoundException;
+import com.spotify.docker.client.exceptions.NonSwarmNodeException;
 import com.spotify.docker.client.exceptions.NotFoundException;
 import com.spotify.docker.client.exceptions.PermissionException;
 import com.spotify.docker.client.exceptions.UnsupportedApiVersionException;
@@ -65,6 +66,7 @@ import com.spotify.docker.client.messages.Version;
 import com.spotify.docker.client.messages.Volume;
 import com.spotify.docker.client.messages.VolumeList;
 import com.spotify.docker.client.messages.swarm.Node;
+import com.spotify.docker.client.messages.swarm.NodeInfo;
 import com.spotify.docker.client.messages.swarm.Secret;
 import com.spotify.docker.client.messages.swarm.SecretCreateResponse;
 import com.spotify.docker.client.messages.swarm.SecretSpec;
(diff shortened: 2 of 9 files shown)


src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -2845,4 +2847,16 @@ public ListVolumesFilterParam(String name, String value) {
...
+  /**
+   * Inspects a swarm node. Only available in Docker API &gt;= 1.24.s
+   *
+   * @param nodeId The id of the swarm node to inspect
+   * @return info about the node
+   *
+   * @throws NonSwarmNodeException if the node is not part of a swarm (503)
+   * @throws DockerException      if a server error occurred (500)
+   * @throws InterruptedException if the thread is interrupted
+   */
```
johnflavin: Can you add `@since API 1.24`? I know you have it in the text above, but it is nice to keep the format consistent.

src/main/java/com/spotify/docker/client/messages/swarm/Node.java
```diff
@@ -28,6 +28,7 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 
+
```
johnflavin: You don't need this whitespace.

src/main/java/com/spotify/docker/client/messages/swarm/Node.java
```diff
@@ -52,6 +53,7 @@
   @JsonProperty("Description")
   public abstract NodeDescription description();
 
+
```
johnflavin: You don't need this whitespace.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -867,6 +868,7 @@ public void testInspectImage() throws Exception {
     assertThat(info.virtualSize(), notNullValue());
   }
 
+
```
johnflavin: You don't need this whitespace.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -5489,7 +5491,61 @@ public void testListNodes() throws Exception {
...
+    Node nut = nodes.get(0);
+
+
+    assertThat(nut.id(), allOf(notNullValue(), not("")));
+
+    Date now = new Date();
+    NodeInfo nodeInfo = sut.inspectNode(nut.id());
+
+    assertThat(nodeInfo.id(), allOf(notNullValue(), not("")));
```
johnflavin: For this and the next few lines, you don't need to assert that they are non-null. They are guaranteed to be non-null since the methods in the `NodeInfo` class were not annotated `@Nullable`.

src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
```diff
@@ -20,11 +20,14 @@
 
 package com.spotify.docker.client;
 
+
```
johnflavin: You don't need this whitespace.

src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
```diff
@@ -82,6 +88,7 @@
  * @see <a href="https://github.com/square/okhttp/tree/master/mockwebserver">
  * https://github.com/square/okhttp/tree/master/mockwebserver</a>
  */
+
```
johnflavin: You don't need this whitespace.

ssullivan:
Thanks for taking the time to review. I will update the PR with the requested changes.

src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
```diff
@@ -301,4 +306,103 @@ public void testBuildPassesMultipleRegistryConfigs() throws Exception {
     final JsonNode headerJsonNode = toJson(BaseEncoding.base64().decode(registryConfigHeader));
     assertThat(headerJsonNode, is(toJson(registryConfigs.configs())));
   }
+
+  @Test
+  public void testInspectNode() throws Exception {
+    final DefaultDockerClient dockerClient = new DefaultDockerClient(builder);
+
+    // build() calls /version to check what format of header to send
+    server.enqueue(new MockResponse()
```
davidxia: Since this is repeated many times, maybe you can abstract into separate method like `enqueueServerApiVersion(final String apiVersion)`?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1957,6 +1959,27 @@ public Task inspectTask(final String taskId) throws DockerException, Interrupted
...
+    WebTarget resource = resource().path("nodes")
+        .path(nodeId);
+
+    try {
+      return request(GET, NodeInfo.class, resource, resource.request(APPLICATION_JSON_TYPE));
+    } catch (DockerRequestException e) {
+      switch (e.status()) {
+        case 404:
+          throw new NodeNotFoundException(nodeId);
+        case 503:
```
davidxia: [Docs for API 1.24](https://docs.docker.com/engine/api/v1.24/#inspect-a-node) say this is 406 while those for [1.28](https://docs.docker.com/engine/api/v1.29/#operation/NodeInspect) say 503. If 1.24 docs are correct, is it worth handling for that old API version?

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -5489,7 +5490,63 @@ public void testListNodes() throws Exception {
       assertThat(plugin.name(), allOf(notNullValue(), not("")));
     }
   }
-  
+
+  @Test
+  public void testInspectNode() throws Exception {
```
davidxia: @mattnworb Do we want to add integration tests like these at all anymore? Seems like if we have a unit test we can skip this.

davidxia:
@ssullivan Thanks so much for your PR.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
ssullivan: I wasn't sure which way to go on this initially so I made both.

src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
ssullivan: Yeah, I was thinking the same thing. I'll update it

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
ssullivan: I need to look over the API for newer versions but I think they don't use that 406 status code so it could probably safely handled.

src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
```diff
@@ -301,4 +306,80 @@ public void testBuildPassesMultipleRegistryConfigs() throws Exception {
     final JsonNode headerJsonNode = toJson(BaseEncoding.base64().decode(registryConfigHeader));
     assertThat(headerJsonNode, is(toJson(registryConfigs.configs())));
   }
+
+  @Test
+  public void testInspectNode() throws Exception {
+    final DefaultDockerClient dockerClient = new DefaultDockerClient(builder);
+
+    // build() calls /version to check what format of header to send
+    enqueueServerApiVersion("1.28");
```
mattnworb: awesome 👏  although you don't have to repeat the comment each time you call this method

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
mattnworb: +1 to what david said - with the unit test, you can test that each field has a certain expected value, whereas with this test you can only really test really broad things about the response.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
ssullivan: Looks like its not being used in the newer versions for some other purpose.
Would you like me to modify this to throw the NonSwarmNodeException for the 406?

davidxia: Meh, you can leave this as is. I was just curious.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
davidxia: Let just remove this integration test. Builds are slow enough. :)

ssullivan: Just updated this test class and removed this test method in favor of the unit tests.

PR #775, opened 2017-06-02
default to pulling auth info from docker config file

mattnworb:
This commit changes the behavior of DefaultDockerClient to default to using
authentication info from the docker-cli configuration file. This will allow
users to get the same authentication support they would normally have from
the docker-cli out-of-the-box.

Additionally, MultiRegistryAuthSupplier can be used to combine multiple
RegistryAuthSuppliers - for instance, to be able to combine the dynamic
"fetch AccessTokens" support for Google Container Registry with the ability
to read authentication info from the config file for non-GCR registries.

As a part of this, also shuffled the package names around a bit.

Still TODO:
- [x] more tests
- [x] update CHANGELOG


mattnworb:
@davidxia @mavenraven

src/main/java/com/spotify/docker/client/ImageRef.java
```diff
@@ -44,18 +49,39 @@ public ImageRef(final String image) {
         this.tag = null;
       }
     }
+    final String[] parts = image.split("/", 2);
+    if (parts.length > 1 && isRegistry(parts[0])) {
+      this.registry = parts[0];
+    } else {
+      this.registry = DEFAULT_REGISTRY;
+    }
```
mattnworb: this was added here to ImageRef to make the job of `ConfigFileRegistryAuthSupplier.authFor(imagename)`  easier

mattnworb:
cc @caipre

src/test/java/com/spotify/docker/client/auth/MultiRegistryAuthSupplierTest.java
```diff
@@ -0,0 +1,127 @@
...
+      new MultiRegistryAuthSupplier(ImmutableList.of(supplier1, supplier2));
+
+  @Test
+  public void testAuthFor() throws Exception {
+    final String image1 = "foobar:latest";
+    final RegistryAuth auth1 = RegistryAuth.builder().build();
+    when(supplier1.authFor(image1)).thenReturn(auth1);
+
+    assertThat(multiSupplier.authFor(image1), is(auth1));
+    verify(supplier2, never()).authFor(image1);
```
davidxia: Is this important to test? ie that the sub-suppliers are in a list that's traversed in order? Test needs to know the implementation of MultiRegistryAuthSupplier rather than just checking the correct auth is returned.

src/test/java/com/spotify/docker/client/auth/MultiRegistryAuthSupplierTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Test
+  public void testAuthForSwarm() throws Exception {
+    final RegistryAuth auth1 = RegistryAuth.builder()
+        .email("a@b.com")
+        .build();
+    when(supplier1.authForSwarm()).thenReturn(auth1);
+
+    assertThat(multiSupplier.authForSwarm(), is(auth1));
+    verify(supplier2, never()).authForSwarm();
```
davidxia: See comment above.

src/main/java/com/spotify/docker/client/auth/MultiRegistryAuthSupplier.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+    return null;
+
+  }
+
+  @Override
+  public RegistryConfigs authForBuild() throws DockerException {
+    final Map<String, RegistryAuth> allConfigs = new HashMap<>();
+    // iterate through suppliers in reverse so that the earlier suppliers in the list
+    // have precedence
```
davidxia: Just curious why earlier ones should have precedence.

src/test/java/com/spotify/docker/client/NoOpRegistryAuthSupplierTest.java
```diff
@@ -24,6 +24,7 @@
 import static org.junit.Assert.assertNull;
 import static org.mockito.Mockito.mock;
 
+import com.spotify.docker.client.auth.NoOpRegistryAuthSupplier;
```
mattnworb: thanks, fixed

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## 8.6.3 (not yet released)
+
+### Expaned RegistryAuthSupplier support
+Add RegistryAuthSuppliers for:
+- reading from docker config file (ConfigFileRegistryAuthSupplier)
+- combining multiple suppliers (MultiRegistryAuthSupplier)
+
+Also change the default behavior of DefaultDockerClient.Builder to use the
```
mattnworb: I'm not really sure, I'm not good on knowing when to change this. What do you think?

src/test/java/com/spotify/docker/client/auth/MultiRegistryAuthSupplierTest.java
mattnworb: yea I intend for the order of the list to have meaning. I'll add something to the javadoc to make that more specific.

src/main/java/com/spotify/docker/client/auth/MultiRegistryAuthSupplier.java
mattnworb: I think if we were to leave the behavior as undefined here it would lead to confusion, so I rather be specific.

src/test/java/com/spotify/docker/client/auth/MultiRegistryAuthSupplierTest.java
mattnworb: added a commit to clarify this, PTAL

CHANGELOG.md
mattnworb: we decided to bump the minor version

caipre:
👍

PR #776, opened 2017-06-02
Make unauthed image pull error clearer

davidxia:
Updated corresponding test


mattnworb:
does the Remote API actually return 401? I don't see it [in the docs](https://docs.docker.com/engine/api/v1.28/#operation/ImageCreate)

davidxia:
A 401 is returned (at least on recent versions of docker). It's not in the docs. I opened a PR upstream to update docs. https://github.com/moby/moby/pull/33499

mattnworb:
I have some work in 775 and 777 that would be nice to release with this.

davidxia:
Oops, the 401 I saw was coming from the docker registry.

PR #779, opened 2017-06-04
ISSUE-778: Quick fix for path matching error

ssullivan:
I've been running into the issue reported in ISSUE #778 and this is what I've been using as a quick fix  for the problem (it may not be the best way to do it but thought I would share)

Issue #778, opened 2017-06-03
CompressedDirectory code fails in path match with * on windows


lucastheisen:
## Description

Seems that * breaks your CompressedDirectory.  Discovered when trying to use the [maven plugin](https://github.com/spotify/dockerfile-maven/issues/15).

## How to reproduce

`mvn test` on a windows machine

## What do you expect

All tests to pass

## What happened instead

Exception was thrown

## Software:

- `docker version`: Docker version 17.03.1-ce, build c6d412e (for both client and server, its the same machine)
- Spotify's docker-client version: 8.6.3-SNAPSHOT

## Full backtrace

```
Running com.spotify.docker.client.CompressedDirectoryTest
Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.099 sec <<< FAILURE!
testFileWithIgnore(com.spotify.docker.client.CompressedDirectoryTest)  Time elapsed: 0.022 sec  <<< ERROR!
java.nio.file.InvalidPathException: Illegal char <*> at index 10: !subdir3\\*.keep
        at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
        at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
        at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
        at sun.nio.fs.AbstractPath.startsWith(AbstractPath.java:43)
        at com.spotify.docker.client.CompressedDirectory$DockerIgnorePathMatcher.matches(CompressedDirectory.java:427)
        at com.spotify.docker.client.CompressedDirectory$Visitor.exclude(CompressedDirectory.java:316)
        at com.spotify.docker.client.CompressedDirectory$Visitor.visitFile(CompressedDirectory.java:293)
        at com.spotify.docker.client.CompressedDirectory$Visitor.visitFile(CompressedDirectory.java:252)
        at java.nio.file.Files.walkFileTree(Files.java:2670)
        at com.spotify.docker.client.CompressedDirectory.create(CompressedDirectory.java:114)
        at com.spotify.docker.client.CompressedDirectoryTest.testFileWithIgnore(CompressedDirectoryTest.java:71)
[...]

PR #780, opened 2017-06-04
Fix InvalidPathException when match string includes * #778

lucastheisen:
This is an alternative to #779 which is a little safer in that it captures any invalid path (not just "*", and not just windows).  It also [fixed some java url wonkiness](https://stackoverflow.com/a/17870390/516433) that caused other unit tests to fail for illegal char in path.

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -455,7 +455,8 @@ public class DefaultDockerClientTest {
   public void testBuildImageIdWithBuildargs() throws Exception {
     requireDockerApiVersionAtLeast("1.21", "build args");
 
-    final String dockerDirectory = Resources.getResource("dockerDirectoryWithBuildargs").getPath();
+    final String dockerDirectory = 
+        Paths.get(Resources.getResource("dockerDirectoryWithBuildargs").toURI()).toString();
     final String buildargs = "{\"testargument\":\"22-12-2015\"}";
     final BuildParam buildParam =
         BuildParam.create("buildargs", URLEncoder.encode(buildargs, "UTF-8"));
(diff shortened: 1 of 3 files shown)


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -455,7 +455,8 @@ public void testPullPrivateRepoWithoutAuth() throws Exception {
   public void testBuildImageIdWithBuildargs() throws Exception {
     requireDockerApiVersionAtLeast("1.21", "build args");
 
-    final String dockerDirectory = Resources.getResource("dockerDirectoryWithBuildargs").getPath();
```
davidxia: Were these tests failing on Windows machines?

lucastheisen:
@davidxia , its not :
```
final String dockerDirectory = Resources.getResource("dockerDirectoryWithBuildargs").getPath();
```
that fails, its later when `dockerDirectory` is used that it throws an exception (see [this comment](https://github.com/spotify/docker-client/pull/779#issuecomment-306052712) for details).  Specifically, on windows, `dockerDirectory` would have a value like `/C:/...` instead of `C:/...` which throws an `InvalidPathException` when you try to use it as a `Path` (and `Path.startsWith(string)` converts the string to a `Path` object before comparing).

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
lucastheisen: See reply in [comment below](https://github.com/spotify/docker-client/pull/780#issuecomment-306270015) (i didn't notice this reply box :) )

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -455,7 +455,8 @@ public void testPullPrivateRepoWithoutAuth() throws Exception {
...
+    final String dockerDirectory = 
+        Paths.get(Resources.getResource("dockerDirectoryWithBuildargs").toURI()).toString();
```
mattnworb: this line is repeated many times but with different directory names - can you wrap it in a method and also have a comment in the method explaining why the URL -> URI -> String -> Path dance is needed?

lucastheisen: Yeah, i did that on purpose...  I find it easier to get my pull requests accepted (across many different projects) if i minimize my changes and adopt the approach of the project.  So i changed each line, but definitely agree that creating a helper method makes more sense.  Let me do so real quick.  Should have an update in a little while.

lucastheisen:
Argh!  Cut/paste error.  My bad... new version should be pushed in a second...

mattnworb:
thanks!

PR #789, opened 2017-06-09
788 - Added NanoCpus property to HostConfig

mateuszzawadzki:


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1882,6 +1882,31 @@ public void testContainerWithMoreCpuOptions() throws Exception {
   }
 
   @Test
+  public void testContainerWithNanoCpusOptions() throws Exception {
+    requireDockerApiVersionAtLeast("1.19", "Container creation with NanoCpus option");
```
johnflavin: Looking at the test failures, for instance, job [#1530.4](https://travis-ci.org/spotify/docker-client/jobs/241179322), I am guessing that you should require at least API version `1.25` for this test.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -1882,6 +1882,31 @@ public void testContainerWithMoreCpuOptions() throws Exception {
   }
 
   @Test
+  public void testContainerWithNanoCpusOptions() throws Exception {
```
davidxia: Could you replace this integration test with a unit test instead? See https://github.com/spotify/docker-client/blob/master/CONTRIBUTING.md#unit-tests-and-integration-tests

mateuszzawadzki: Can you please explain why it should be done in this case? If I understand this correctly, I'd have to mock json response and then parse it to HostConfig, which seems kinda confusing and I'm not really sure how to do it correctly. Testing it against docker daemon, like the rest of HostConfig properties, seems like a better choice.

ssullivan: There are some good examples in DefaultDockerClientUnitTest but the idea is that you can have some sample JSON files that represent the different kinds of responses the actual docker daemon might return.  I think its harder to test certain scenarios if you are using a live docker daemon (also the tests run slower).

davidxia:
I cleaned up the git commits here https://github.com/spotify/docker-client/pull/806

Issue #737, opened 2017-05-07
Add ManagerStatus into Node


olegtikhonov:
## Description
Node is not contains ManagerStatus
[Add feature/bug description here]

## How to reproduce
dockerClient.listNodes() or look at Node.java
[Add steps on how to reproduce this issue]

## What do you expect
    @JsonProperty("Leader")
    private Boolean leader;
    @JsonProperty("Reachability")
    private String reachability;
    @JsonProperty("Addr")
    private String addr;
[Describe what do you expect to happen]

## What happened instead
Is not there
[Describe the actual results]

## Software:

- `8.4.0`: [Add the output of `docker version` here, both client and server]
- Spotify's docker-client version: [Add docker-client version here]

## Full backtrace

```text
[Paste full backtrace here]
```

davidxia:
added in https://github.com/spotify/docker-client/pull/774

ssullivan:
The ManagerStatus got added onto a new class called NodeInfo which the inspectNode method returns. I could do a quick PR now that #774 has been merged and add ManagerStatus into the Node class.

hatunet:
It's did.

Why not solve it ?

PR #790, opened 2017-06-10
ISSUE-737: Add ManagerStatus into Node

ssullivan:
I saw that there was a closed issue #737  that my PR was referenced but my additions didn't add ManagerStatus into node.

This PR adds MangerStatus and NodeStatus into Node which should satisfy #737

### DIFF
--- a/src/main/java/com/spotify/docker/client/DockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DockerClient.java
@@ -2851,6 +2851,18 @@ public interface DockerClient extends Closeable {
   List<Node> listNodes() throws DockerException, InterruptedException;
 
   /**
+   * List swarm nodes that match the given criteria. Only available in Docker API &gt;= 1.24.
+   *
+   * @param criteria Node listing and filtering options.
+   * @return A list of nodes.
+   *
+   * @throws DockerException      if a server error occurred (500)
+   * @throws InterruptedException If the thread is interrupted
+   * @since Docker 1.12, API version 1.24
+   */
+  List<Node> listNodes(final Node.Criteria criteria) throws DockerException, InterruptedException;
+
+  /**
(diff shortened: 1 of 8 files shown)


src/main/java/com/spotify/docker/client/DockerClient.java
```diff
@@ -2850,6 +2849,18 @@ public ListVolumesFilterParam(String name, String value) {
   List<Node> listNodes() throws DockerException, InterruptedException;
 
   /**
+   * List swarm nodes that match the given criteriag. Only available in Docker API &gt;= 1.24.
```
davidxia: typo in criteria

src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
```diff
@@ -364,6 +416,184 @@ public void testInspectNonSwarmNode() throws Exception {
...
+
+    server.enqueue(new MockResponse()
+        .setResponseCode(200)
+        .addHeader("Content-Type", "application/json")
+        .setBody(
+            fixture("fixtures/1.28/listNodes.json")
+        )
+    );
+
+    final List<Node> nodes = dockerClient.listNodes(Node.find().nodeId("24ifsmvkjbyhk").build());
```
davidxia: `listNodes()` will return the list of one node provided by `fixtures/1.28/listNodes.json` all the time, regardless of the nodeId used here. So this test only seems to check that `listNodes()` can take `Node.Criteria` as a parameter. I would lean towards removing this and the tests below (keep `testListNodesWithServerError()`) since they don't add much.

ssullivan: I'll update and remove the tests, thanks for taking a look.

ssullivan:
I've updated the PR with the requested changes.

PR #796, opened 2017-06-13
Update SwarmJoin non/nullable fields

fdonze:
SwarmJoin.AdvertiseAddr field is Nullable but RemoteAddrs and JoinToken are not


davidxia:
Is there an easy way to write a test for this?

fdonze:
@davidxia An integration test would need a second node, I do not know enough about the test infrastructure to know how feasible/easy it is. I would need guidance on this.

davidxia:
Yea I thought an integration test would be hard. How about a unit test then? See DefaultDockerClientUnitTest. The test failure is caused by https://github.com/spotify/docker-client/pulls

fdonze:
@davidxia I added a unit test for joinSwarm. Not sure if it makes sense to add more tests for each nullable fields as it would just test the Nullable annotation handling. Wdyt?

PR #798, opened 2017-06-14
Add ContainerConfig.Healthcheck.startPeriod

fdonze:
Add the ContainerConfig.Healthcheck.startPeriod field introduced in API v1.29

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
@@ -374,6 +374,7 @@ public abstract class ContainerConfig {
 
     /**
      * In nanoseconds.
+     * @since API 1.29
      */
     @Nullable
     @JsonProperty("StartPeriod")
--- a/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
@@ -384,6 +384,14 @@ public abstract class ContainerConfig {
             @JsonProperty("Test") final List<String> test,
             @JsonProperty("Interval") final Long interval,
             @JsonProperty("Timeout") final Long timeout,
+            @JsonProperty("Retries") final Integer retries) {
+      return create(test, interval, timeout, retries, null);
+    }
+
+    public static Healthcheck create(
+            @JsonProperty("Test") final List<String> test,
+            @JsonProperty("Interval") final Long interval,
+            @JsonProperty("Timeout") final Long timeout,
             @JsonProperty("Retries") final Integer retries,
             @JsonProperty("StartPeriod") final Long startPeriod) {
       final Builder builder = builder();
--- a/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
@@ -379,7 +379,6 @@ public abstract class ContainerConfig {
     @JsonProperty("StartPeriod")
     public abstract Long startPeriod();
 
-    @JsonCreator
     public static Healthcheck create(
             @JsonProperty("Test") final List<String> test,
             @JsonProperty("Interval") final Long interval,
@@ -388,6 +387,7 @@ public abstract class ContainerConfig {
       return create(test, interval, timeout, retries, null);
     }
 
+    @JsonCreator
     public static Healthcheck create(
             @JsonProperty("Test") final List<String> test,
             @JsonProperty("Interval") final Long interval,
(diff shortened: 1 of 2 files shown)


src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
```diff
@@ -372,12 +372,20 @@ public Builder volumes(final String... volumes) {
...
+    public abstract Long startPeriod();
+
     @JsonCreator
     public static Healthcheck create(
             @JsonProperty("Test") final List<String> test,
             @JsonProperty("Interval") final Long interval,
             @JsonProperty("Timeout") final Long timeout,
-            @JsonProperty("Retries") final Integer retries) {
+            @JsonProperty("Retries") final Integer retries,
+            @JsonProperty("StartPeriod") final Long startPeriod) {
```
johnflavin: I don't know if this is a great idea, so maybe @davidxia or @mattnworb could weigh in, but changing this method signature seems too disruptive. Should we add another convenience `create` method with the same signature as the old one (`create(List<String, Long, Long, Integer)`) that just calls this one with a final `(Long) null`? It could even be `@Deprecated`. But that would give users some time to migrate their code before it just breaks on a version bump.

I'm sure this situation has come up in the past, but I don't know what we have done.

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
```diff
@@ -392,6 +400,9 @@ public static Healthcheck create(
       if (retries != null) {
         builder.retries(retries);
       }
+      if (startPeriod != null) {
```
johnflavin: You should not need this null check since you marked it `@Nullable` above. If someone sends in a null value, it's no big deal.

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
mattnworb: I like what @johnflavin said; we can have multiple `create()` factory methods, and adding overloads is a nice way to deal with the addition of new (optional) parameters. There can only be one `@JsonCreator` annotated method but that doesn't mean you can't have other `public static` factories.

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
fdonze: @johnflavin It seems to be the pattern regardless if the field is nullable (see retries for example).  I can update if necessary.

fdonze:
@johnflavin @mattnworb.  I updated the PR, thx for your help.

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
```diff
@@ -372,12 +372,28 @@ public Builder volumes(final String... volumes) {
...
     @JsonCreator
     public static Healthcheck create(
             @JsonProperty("Test") final List<String> test,
             @JsonProperty("Interval") final Long interval,
             @JsonProperty("Timeout") final Long timeout,
             @JsonProperty("Retries") final Integer retries) {
+      return create(test, interval, timeout, retries, null);
+    }
+
+    public static Healthcheck create(
```
johnflavin: This `create` method, the one with all the parameters, needs to be the `@JsonCreator`. Otherwise `"StartPeriod"` will never be set when creating a `Heathcheck` from json, even if it is set in the json.

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
davidxia: haha, @johnflavin looks like you added those null check originally.

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
```diff
@@ -372,12 +372,28 @@ public Builder volumes(final String... volumes) {
     @JsonProperty("Retries")
     public abstract Integer retries();
 
-    @JsonCreator
+    /**
+     * In nanoseconds.
+     */
+    @Nullable
+    @JsonProperty("StartPeriod")
+    public abstract Long startPeriod();
```
davidxia: Can you add a javadoc saying this is only available for docker api versions 1.29+?

davidxia:
Thanks for your PR. Thanks @johnflavin for reviewing.

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
johnflavin: Yeah, that sounds about right. To look at it charitably, if I can identify other people's mistakes that I have also made myself, it means I am growing.

davidxia: I added a commit to this PR here to remove these checks. https://github.com/spotify/docker-client/pull/807

davidxia:
Closed in favor of #807.

PR #799, opened 2017-06-14
Add ServiceCreateResponse.warnings

fdonze:
Add ServiceCreateResponse.warnings nullable field.

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/ServiceCreateResponse.java
+++ b/src/main/java/com/spotify/docker/client/messages/ServiceCreateResponse.java
@@ -27,6 +27,10 @@ import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+import javax.annotation.Nullable;
 
 
 @AutoValue
@@ -36,8 +40,24 @@ public abstract class ServiceCreateResponse {
   @JsonProperty("ID")
   public abstract String id();
 
+  /**
+   * @since API 1.25
+   */
+  @Nullable
+  @JsonProperty("Warnings")
+  public abstract ImmutableList<String> warnings();
+
+  static ServiceCreateResponse create(
+          @JsonProperty("ID") final String id) {
+    return create(id, null);
+  }
+
   @JsonCreator
-  static ServiceCreateResponse create(@JsonProperty("ID") final String id) {
-    return new AutoValue_ServiceCreateResponse(id);
+  static ServiceCreateResponse create(
+          @JsonProperty("ID") final String id,
+          @JsonProperty("Warnings") final List<String> warnings) {
+    final ImmutableList<String> warningsT = warnings == null
+            ? null : ImmutableList.copyOf(warnings);
+    return new AutoValue_ServiceCreateResponse(id, warningsT);
   }
 }
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -4870,6 +4870,28 @@ public class DefaultDockerClientTest {
   }
 
   @Test
+  public void testCreateServiceWithWarnings() throws Exception {
+    requireDockerApiVersionAtLeast("1.25", "swarm support");
+
+    final TaskSpec taskSpec = TaskSpec.builder()
+        .containerSpec(ContainerSpec.builder()
+            .image("this_image_is_not_found_in_the_registry")
+            .build())
+        .build();
+
+    final ServiceSpec spec = ServiceSpec.builder()
+        .name(randomName())
+        .taskTemplate(taskSpec)
+        .build();
+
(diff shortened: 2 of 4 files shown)


src/main/java/com/spotify/docker/client/messages/ServiceCreateResponse.java
```diff
@@ -36,8 +40,21 @@
...
+
   @JsonCreator
-  static ServiceCreateResponse create(@JsonProperty("ID") final String id) {
-    return new AutoValue_ServiceCreateResponse(id);
+  static ServiceCreateResponse create(
+          @JsonProperty("ID") final String id) {
+    return create(id, null);
+  }
+  
+  static ServiceCreateResponse create(
```
johnflavin: Same comment as in [798](https://github.com/spotify/docker-client/pull/798#discussion_r122107311). This `create` needs to be the `@JsonCreator`, or `"Warnings"` would never be deserialized from json.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4870,6 +4870,19 @@ public void testCreateService() throws Exception {
...
 
   @Test
+  public void testCreateServiceWithWarnings() throws Exception {
+    requireDockerApiVersionAtLeast("1.25", "swarm support");
+
+    final ServiceSpec spec = createBadServiceSpec(randomName());
+
+    final ServiceCreateResponse response = sut.createService(spec);
+    assertThat(response.id(), is(notNullValue()));
+    assertThat(response.warnings(), is(hasSize(1)));
```
johnflavin: It might be nice to also have an assertion about the contents of the warning. Like, if the warning message contains the name that wasn't found, something like `assertThat(warning.get(0), contains(whatever the random name was))`. But if you don't actually know what the warning message would be, then feel free to ignore this request.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -5552,6 +5565,21 @@ private ServiceSpec createServiceSpec(final String serviceName,
         .build();
   }
 
+  private ServiceSpec createBadServiceSpec(final String serviceName) {
```
johnflavin: Do you need this bit of code to be in a method rather than just part of the test? Do you plan to re-use it?

src/main/java/com/spotify/docker/client/messages/ServiceCreateResponse.java
fdonze: @johnflavin That's interesting, the test "testCreateServiceWithWarnings" does seem to return the "warnings" content even though the JsonCreator is not on the create with all parameters.

davidxia: hm, that's strange...

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4870,6 +4870,28 @@ public void testCreateService() throws Exception {
   }
 
   @Test
+  public void testCreateServiceWithWarnings() throws Exception {
```
davidxia: Could you replace this integration test with a unit test instead? See `DefaultDockerClientUnitTest` and https://github.com/spotify/docker-client/blob/master/CONTRIBUTING.md#unit-tests-and-integration-tests

fdonze: @davidxia I can add a unit test. It might still be useful to keep the current test. Let me explain: I noticed testing the changes that the docker documentation is incorrect (it states that the field in "warning" and a string instead of "warnings" as an array). A unit test would not have caught this. This might not be the intent of the docker-client tests and the developer should be responsible to test the right contract but let me know if you think that the current test as any value.

davidxia: Oh good catch. In that case, yes this has value. Let's keep it.

davidxia: @fdonze I see the [docs here](https://docs.docker.com/engine/api/v1.25/#operation/ServiceCreate) are as you said. I wanted to create a PR for them. Where's the source code for the docs? Doesn't seem to be [here](https://github.com/moby/moby/tree/master/docs/api).

src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
```diff
@@ -364,6 +369,39 @@ public void testInspectNonSwarmNode() throws Exception {
...
 
+  @Test
+  public void testCreateServiceWithWarnings() throws Exception {
+    final DefaultDockerClient dockerClient = new DefaultDockerClient(builder);
+
+    // build() calls /version to check what format of header to send
+    enqueueServerApiVersion("1.25");
+
+    server.enqueue(new MockResponse()
+            .setResponseCode(201)
```
davidxia: Indentation here should be four instead of eight spaces.

fdonze:
@davidxia I fixed the indentation and squashed the commits. 
All tests are failing due to some strange issue: 'The command "./.travis.sh install_docker" failed and exited with 100 during ."

davidxia:
Don't worry about the test failures. It's caused by https://github.com/spotify/docker-client/pulls

PR #801, opened 2017-06-16
Do not refresh token if no expiration

davidxia:
Make `ContainerRegistryAuthSupplier` not refresh token
if no expiration has been provided.

A `GoogleCredentials` can be created from an `AccessToken`
that has no expiration. If that's passed to
`ContainerRegistryAuthSupplier` there'll be an NPE.


davidxia:
Thanks for the speedy review. cc @ravenmaven

PR #803, opened 2017-06-17
ISSUE-793: Make ManagerStatus and ManagerStatus.Leader nullable

fdonze:
Fix NodeInfo.ManagerStatus and ManagerStatus.Leader: both fields can be null.


man4j:
+1

davidxia:
This will be fixed by https://github.com/spotify/docker-client/pull/790.

PR #809, opened 2017-06-26
Replace deprecated registryAuth() methods in IT

davidxia:


mavenraven:
looks good except for lack of tests around the new impl of RegistryAuthSupplier

davidxia:
I didn't think `FixedAuthRegistrySupplier` needed tests since it's only in the test sources because it's only used for tests.

davidxia:
Oops, I overlooked that class probably because of the name. See https://github.com/spotify/docker-client/pull/818

PR #812, opened 2017-06-29
Fix bug in pushing hub images

### LINKED ISSUE #786: can't push images to docker hub if image name is like "user/reponame"
```java
docker.push("mattnworb/foobar");
```

results in the DefaultDockerClient calling `RegistryAuthSupplier.imageFor("mattnworb/foobar")`. If the DefaultDockerClient is using the ConfigFileRegistryAuthSupplier, this implementation doesn't know to expand this image name and won't retrieve the config for `https://index.docker.io/v1` from ~/.docker/config.json.

davidxia:
when image is named "user/repo." Fixes #786

### DIFF
--- a/src/main/java/com/spotify/docker/client/ImageRef.java
+++ b/src/main/java/com/spotify/docker/client/ImageRef.java
@@ -20,12 +20,15 @@
 
 package com.spotify.docker.client;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 
 public class ImageRef {
 
   private static final String DEFAULT_REGISTRY = "docker.io";
+  private static final String DEFAULT_REGISTRY_URL = "https://index.docker.io/v1/";
 
+  private final String registryUrl;
   private final String registry;
   private final String image;
   private final String tag;
@@ -52,8 +55,10 @@ public class ImageRef {
     final String[] parts = image.split("/", 2);
     if (parts.length > 1 && isRegistry(parts[0])) {
       this.registry = parts[0];
+      this.registryUrl = parseRegistryUrl(parts[0]);
     } else {
       this.registry = DEFAULT_REGISTRY;
+      this.registryUrl = DEFAULT_REGISTRY_URL;
     }
   }
 
@@ -75,6 +80,11 @@ public class ImageRef {
     return registry;
   }
 
+  /** Registry URL. */
+  public String getRegistryUrl() {
+    return registryUrl;
+  }
+
   @Override
   public String toString() {
     return MoreObjects.toStringHelper(this)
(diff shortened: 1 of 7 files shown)


src/main/java/com/spotify/docker/client/ImageRef.java
```diff
@@ -20,12 +20,17 @@
...
 import com.google.common.base.MoreObjects;
+import java.net.MalformedURLException;
+import java.net.URL;
 
 public class ImageRef {
 
   private static final String DEFAULT_REGISTRY = "docker.io";
+  private static final URL DEFAULT_REGISTRY_URL = parseRegistryUrl("https://index.docker.io/v1/");
 
+  private final URL registryUrl;
```
davidxia: Decided to make this a URL.

src/main/java/com/spotify/docker/client/ImageRef.java
```diff
@@ -75,6 +82,11 @@ public String getRegistryName() {
     return registry;
   }
 
+  /** Registry URL. */
+  public URL getRegistryUrl() {
```
davidxia: New method to not break backwards compat on `getRegistryName()`.

src/main/java/com/spotify/docker/client/ImageRef.java
```diff
@@ -83,4 +93,23 @@ public String toString() {
         .add("tag", tag)
         .toString();
   }
+
+  /** Return registry server address given first part of image. */
```
davidxia: Is there a better way?

src/main/java/com/spotify/docker/client/ImageRef.java
```diff
@@ -75,6 +80,11 @@ public String getRegistryName() {
     return registry;
   }
 
+  /** Registry URL. */
+  public String getRegistryUrl() {
```
davidxia: New method to not break backwards compat on `getRegistryName()`.

src/main/java/com/spotify/docker/client/ImageRef.java
```diff
@@ -83,4 +93,23 @@ public String toString() {
...
+      return DEFAULT_REGISTRY_URL;
+    } else if (url.matches("(^|(\\w+)\\.)gcr\\.io$") || url.equals("gcr.kubernetes.io")) {
+      // GCR uses https
+      return "https://" + url;
+    } else if (url.equals("quay.io")) {
+      // Quay doesn't use any protocol in credentials file
+      return "quay.io";
+    } else if (!url.contains("http://") && !url.contains("https://")) {
+      // Assume http
+      return "http://" + url;
```
mattnworb: why assume http? I think https is probably a smarter assumption to make.

mattnworb: Docker will interpret a "registry URL" like `example.com/foo/bar` as a registry that uses `https` and will only pull over `http` if specially configured to do so: https://docs.docker.com/engine/reference/commandline/pull/#pull-from-a-different-registry

mattnworb:
I think the root bug fix - knowing when to use `index.docker.io` for docker hub images - is correct, but I think the default of http is dangerous

davidxia:
Updated to use https.

mattnworb:
:+1:

davidxia:
hopefully we won't need to add more custom registry names to `ImageRef`.

PR #815, opened 2017-07-02
Clarify why X-Registry-Auth header can't be empty

### LINKED ISSUE #763: DefaultDockerClient sends "null" as header value when there is no authentication configured
https://github.com/spotify/docker-client/blob/1b513bf225b87795b75f3d730ebdf8c66ae0e161/src/main/java/com/spotify/docker/client/DefaultDockerClient.java#L2413-L2415

Not a big deal (since it is working obviously), but it would be better to not send a header at all in these cases.

davidxia:
Move comment to more noticeable place.
Make `DefaultDockerClient.authRegistryHeader()` return null
if auth is empty.

If `DefaultDockerClient.authHeader()` returns null,
`PushPullIT`'s `testPushImageToPrivateAuthedRegistryWithoutAuth()`
and `testPushImageToPrivateUnauthedRegistryWithoutAuth()` would fail.

resolves #763

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -1281,8 +1281,6 @@ public class DefaultDockerClient implements DockerClient, Closeable {
       resource = resource.queryParam("tag", imageRef.getTag());
     }
 
-    // the docker daemon requires that the X-Registry-Auth header is specified
-    // with a non-empty string even if your registry doesn't use authentication
     try (ProgressStream push =
              request(POST, ProgressStream.class, resource,
                      resource.request(APPLICATION_JSON_TYPE)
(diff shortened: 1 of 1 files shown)


src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2528,6 +2526,8 @@ private String message(final Response response) {
   }
 
   private String authHeader(final RegistryAuth registryAuth) throws DockerException {
+    // the docker daemon requires that the X-Registry-Auth header is specified
+    // with a non-empty string even if your registry doesn't use authentication
```
mattnworb: is there any reference to this need? It is not clear in https://docs.docker.com/engine/api/v1.30/#section/Authentication

davidxia:
I couldn't find any docs on this. Ryan left the original comment.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2528,6 +2526,8 @@ private String message(final Response response) {
...
     if (registryAuth == null) {
       return "null";
```
mattnworb: @davidxia did you mean to keep this line?

davidxia: Yes, otherwise it'll break. See commit message above.

PR #816, opened 2017-07-02
Fix RegistryAuth JSON properties

### LINKED ISSUE #648: identityToken RegistryAuth should look for identitytoken
## Description

My previous feature add #581 missed the mark slightly. I apparently goofed the actual identitytoken being entirely lowercase in my unit tests and code.

## How to reproduce

`RegistryAuth.fromDockerConfig()` with a real identitytoken based config.json file.

## What do you expect

RegistryAuth should be valid and correctly parsed.

## What happened instead

"Failed to parse auth string for {}"

## Software:

- `docker version`: 
```
$ docker version
Client:
 Version:      1.12.3
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   6b644ec
 Built:        Wed Oct 26 23:26:11 2016
 OS/Arch:      darwin/amd64

Server:
 Version:      1.12.3
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   6b644ec
 Built:        Wed Oct 26 23:26:11 2016
 OS/Arch:      linux/amd64
```
- Spotify's docker-client version: 8.0.0

davidxia:
`RegistryAuth` keys should all be lowercase.
See https://docs.docker.com/engine/api/v1.25/#section/Authentication.

Fixes #648

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/RegistryAuth.java
+++ b/src/main/java/com/spotify/docker/client/messages/RegistryAuth.java
@@ -40,26 +40,26 @@ import org.glassfish.jersey.internal.util.Base64;
 public abstract class RegistryAuth {
 
   @Nullable
-  @JsonProperty("Username")
+  @JsonProperty("username")
   public abstract String username();
 
   @Nullable
-  @JsonProperty("Password")
+  @JsonProperty("password")
   public abstract String password();
 
   /**
    * Unused but must be a well-formed email address (e.g. 1234@5678.com).
    */
   @Nullable
-  @JsonProperty("Email")
+  @JsonProperty("email")
   public abstract String email();
 
   @Nullable
-  @JsonProperty("ServerAddress")
+  @JsonProperty("serveraddress")
   public abstract String serverAddress();
 
   @Nullable
-  @JsonProperty("IdentityToken")
+  @JsonProperty("identitytoken")
   public abstract String identityToken();
 
   @Override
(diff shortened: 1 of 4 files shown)


scheerer:
Good find. Looks like it might possibly fix it. Sadly I am no longer with my previous employer where I have access to test such s setup.

mattnworb:
It is also possible to configure Jackson to treat property names as case insensitive, in case there was ever a previous docker version that had a different casing strategy: https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.8.9/src/main/java/com/fasterxml/jackson/databind/MapperFeature.java#L310-L324

mattnworb:
Should note that the `json` package in golang that docker-cli uses [will fallback to case-insensitive behavior](https://golang.org/pkg/encoding/json/#Unmarshal), so even though docker defines these fields [in all lowercase](https://github.com/moby/moby/blob/91e197d614547f0202e6ae9b8a24d88ee131d950/api/types/auth.go#L4-L7) it will support deserializing a config file with mixed-case.

(good find though)

davidxia:
Should I make any further changes?

davidxia:
More forgiving deserialization of incoming JSON might be helpful. I can't think how it'd break anything, but the javadocs does say there's a [performance hit](https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/MapperFeature.html#ACCEPT_CASE_INSENSITIVE_PROPERTIES).

PR #817, opened 2017-07-03
Adapt to API changes in latest jnr-unixsocket.

mbooth101:
In the latest version of jnr-unixsocket, some getter/setter methods in
the UnixSocketChannel object went away. This change replaces usage of
those getters/setters with calls to getOption/setOption instead.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -139,7 +139,7 @@
     <dependency>
       <groupId>com.github.jnr</groupId>
       <artifactId>jnr-unixsocket</artifactId>
-      <version>0.8</version>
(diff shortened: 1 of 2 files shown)


mbooth101:
FWIW, it looks like only the "DOCKER_VERSION=1.12.6" build failed -- all others succeeded -- but it's not immediately obvious to me why that one test failed only failed for that version of docker.

davidxia:
See https://github.com/spotify/docker-client/pull/918

davidxia:
closed in favor of #918

Issue #822, opened 2017-07-09
Problem with event.next()


olegtikhonov:
## Description
[During the lookup the event I get InvalidFormatException]
## How to reproduce
[event.next()]
## What do you expect
[Will get event.action and event.actor]
## What happened instead
[Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Can not deserialize value of type com.spotify.docker.client.messages.Event$Type from String "service": value not one of declared Enum instance names: [container, volume, image, daemon, network]]
## Software:
- `docker 17.06.0-ce`: [Add the output of `docker version` here, both client and server]
- Spotify's docker-client version: [Add docker-client version here]
## Full backtrace

```text
[Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Can not deserialize value of type com.spotify.docker.client.messages.Event$Type from String "service": value not one of declared Enum instance names: [container, volume, image, daemon, network]
 at [Source: org.apache.http.conn.EofSensorInputStream@2b355cec; line: 1, column: 9] (through reference chain: com.spotify.docker.client.messages.Event["Type"])
	at com.fasterxml.jackson.databind.exc.InvalidFormatException.from(InvalidFormatException.java:74)
	at com.fasterxml.jackson.databind.DeserializationContext.weirdStringException(DeserializationContext.java:1410)
	at com.fasterxml.jackson.databind.DeserializationContext.handleWeirdStringValue(DeserializationContext.java:926)
	at com.fasterxml.jackson.databind.deser.std.EnumDeserializer._deserializeAltString(EnumDeserializer.java:189)
	at com.fasterxml.jackson.databind.deser.std.EnumDeserializer.deserialize(EnumDeserializer.java:126)
	at com.fasterxml.jackson.databind.deser.SettableBeanProperty.deserialize(SettableBeanProperty.java:504)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeWithErrorWrapping(BeanDeserializer.java:511)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeUsingPropertyBased(BeanDeserializer.java:400)
	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.deserializeFromObjectUsingNonDefault(BeanDeserializerBase.java:1191)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeFromObject(BeanDeserializer.java:314)
[...]

PR #825, opened 2017-07-11
Fixes for docker 17.06.0

johnflavin:
In fixing #822, I saw that a lot of tests were failing on docker `17.06.0`. I fixed all the errors in both tests and code so that the tests run cleanly on my machine. 

However, we can't test `17.06.0` on travis yet because it isn't in the [docker apt repo](https://apt.dockerproject.org/repo/).

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -4894,6 +4894,8 @@ public class DefaultDockerClientTest {
   @Test
   public void testCreateServiceWithWarnings() throws Exception {
     requireDockerApiVersionAtLeast("1.25", "swarm support");
+    requireDockerApiVersionLessThan("1.30",
+            "warning on create service with bad image");
 
     final TaskSpec taskSpec = TaskSpec.builder()
         .containerSpec(ContainerSpec.builder()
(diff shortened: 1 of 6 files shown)


olegtikhonov:
Is there a way to get the "fixed" jar somehow? with tag .... "alpha/beta"?

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -5099,13 +5117,22 @@ public void testCreateServiceWithSecretHostnameHostsAndHealthcheck() throws Exce
             .build();
 
     final String[] commandLine = {"ping", "-c4", "localhost"};
+    final long interval = dockerApiVersionLessThan("1.30") ? 30L : 30000000L;
```
mattnworb: did they change the unit in between api versions? 😬

johnflavin: I think they did, yes. Or maybe they were lenient about the units before, and started being more strict. I think these were always intended to be in nanoseconds (at least [our javadoc](https://github.com/spotify/docker-client/blob/master/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java#L358) says they are in nanoseconds).

johnflavin:
I can add something to the change log, but I want to be careful about the wording. I didn't actually bring us up to full compatibility with the new API. I mostly just made sure the tests we run are asserting accurate stuff, and in a couple places I made small modifications to the code where the tests exposed things that were broken.

mattnworb:
@johnflavin sounds good to me, I think any extra info in the changelog is helpful and better than none. Thanks!

PR #836, opened 2017-07-21
Fix https json authFor.  Reader fromConfig throws IllegalArgumentAcce…

efenglu:
docker login creates a json auth config file without a scheme on the host.  There was code to handle this but it assumed the method would return null on failure.  The method fromConfig throws and IllegalArgumentException if it is unable to find the configuration.  I've update the method to handle the fallback case in the catch block.

### DIFF
--- a/src/main/java/com/spotify/docker/client/auth/ConfigFileRegistryAuthSupplier.java
+++ b/src/main/java/com/spotify/docker/client/auth/ConfigFileRegistryAuthSupplier.java
@@ -77,8 +77,12 @@ public class ConfigFileRegistryAuthSupplier implements RegistryAuthSupplier {
       }
       return reader.fromConfig(path, ref.getRegistryName());
     } catch (IllegalArgumentException e) {
-      // no configuration for registry
-      return null;
+      log.debug("Failed first attempt to find auth for {}", ref.getRegistryUrl(), e);
+      try {
+        return reader.fromConfig(path, ref.getRegistryName());
+      } catch (IllegalArgumentException | IOException e2) {
+        throw new DockerException(e2);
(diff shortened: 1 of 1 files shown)


src/main/java/com/spotify/docker/client/auth/ConfigFileRegistryAuthSupplier.java
```diff
@@ -77,8 +77,12 @@ public RegistryAuth authFor(final String imageName) throws DockerException {
       }
       return reader.fromConfig(path, ref.getRegistryName());
     } catch (IllegalArgumentException e) {
-      // no configuration for registry
-      return null;
+      log.debug("Failed first attempt to find auth for {}", ref.getRegistryUrl(), e);
+      try {
+        return reader.fromConfig(path, ref.getRegistryName());
+      } catch (IllegalArgumentException | IOException e2) {
+        throw new DockerException(e2);
```
mattnworb: this shouldn't throw an exception, as there are other code paths that call `authFor(image)` and expect a `null` response if the config file / supplier has no authentication info for the given image name. It's not necessarily an error to not have RegistryAuth info for a given image.

efenglu: I don't disagree, just matching the existing behavior

efenglu: oh oops, I see it now

efenglu: I'll change it to return null

PR #839, opened 2017-07-24
Add the boolean parameter 'Force' to DockerClient#disconnectFromNetwork and fix issue on exceptions being swalled

hgontijo:
Add the boolean parameter 'Force' to DockerClient#disconnectFromNetwork at [Docker Networks](https://docs.docker.com/engine/api/v1.30/#operation/NetworkDisconnect).

Refactored accordingly due to the introduction of the new parameter.

Fix issue at DockerClient#disconnectFromNetwork and DockerClient#connectToNetwork that led to an request error (HTTP 4xx, 5xx) to be swalled and no exception has been thrown.
This was happening since the request was using Response.class and JerseyInvocation#translate would just report the error via Response.class whenever this class is used to carry the response data. If a different class other then Response.class is given, then Jersey will thrown an exception.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -2230,7 +2230,7 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   @Override
   public void connectToNetwork(String containerId, String networkId)
       throws DockerException, InterruptedException {
-    manageNetworkConnection(containerId, "connect", networkId);
+    connectToNetwork(networkId, NetworkConnection.builder().containerId(containerId).build());
   }
 
   @Override
@@ -2239,7 +2239,7 @@ public class DefaultDockerClient implements DockerClient, Closeable {
     final WebTarget resource = resource().path("networks").path(networkId).path("connect");
 
     try {
-      request(POST, Response.class, resource, resource.request(APPLICATION_JSON_TYPE),
+      request(POST, String.class, resource, resource.request(APPLICATION_JSON_TYPE),
               Entity.json(networkConnection));
     } catch (DockerRequestException e) {
       switch (e.status()) {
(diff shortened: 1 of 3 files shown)


hgontijo:
Here's the JerseyInvocation#translate (jersey-client-2.22.2) code I mentioned above: 
```java
    private <T> T translate(final ClientResponse response, final RequestScope scope, final Class<T> responseType)
            throws ProcessingException {
        if (responseType == Response.class) {
            return responseType.cast(new InboundJaxrsResponse(response, scope));
        }

        if (response.getStatusInfo().getFamily() == Response.Status.Family.SUCCESSFUL) {
            try {
                return response.readEntity(responseType);
            } catch (final ProcessingException ex) {
                if (ex.getClass() == ProcessingException.class) {
                    throw new ResponseProcessingException(new InboundJaxrsResponse(response, scope), ex.getCause());
                }
                throw new ResponseProcessingException(new InboundJaxrsResponse(response, scope), ex);
            } catch (final WebApplicationException ex) {
                throw new ResponseProcessingException(new InboundJaxrsResponse(response, scope), ex);
            } catch (final Exception ex) {
                throw new ResponseProcessingException(new InboundJaxrsResponse(response, scope),
                        LocalizationMessages.UNEXPECTED_ERROR_RESPONSE_PROCESSING(), ex);
            }
        } else {
            throw convertToException(new InboundJaxrsResponse(response, scope));
        }
    }
```

hgontijo:
Thoughts on this PR?

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4062,6 +4062,62 @@ public void testNetworksConnectContainer() throws Exception {
   }
 
   @Test
+  public void testNetworksConnectContainerWithoutForcefulDisconnectNetwork() throws Exception {
```
davidxia: I think you can leave out this test as it's almost the same as the one above and we're not testing the difference between forceful and not.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2256,18 +2256,20 @@ public void connectToNetwork(String networkId, NetworkConnection networkConnecti
   @Override
   public void disconnectFromNetwork(String containerId, String networkId)
       throws DockerException, InterruptedException {
-    manageNetworkConnection(containerId, "disconnect", networkId);
+    disconnectFromNetwork(containerId, networkId, false);
   }
 
-  private void manageNetworkConnection(String containerId, String methodname, String networkId)
```
davidxia: Thanks for cleaning this up.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
hgontijo: It makes sense to me, I left the test out. Thanks.

PR #840, opened 2017-07-27
Implemented docker config api methods: list, inspect, create, update and delete

ssullivan:
The latest Docker release has added Config functionality that is similar to the existing Secrets functionality.
This code adds new model objects for the configs API and an implementation of the listConfigs API call.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -38,6 +38,7 @@ import static javax.ws.rs.HttpMethod.DELETE;
 import static javax.ws.rs.HttpMethod.GET;
 import static javax.ws.rs.HttpMethod.POST;
 import static javax.ws.rs.HttpMethod.PUT;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON_TYPE;
 import static javax.ws.rs.core.MediaType.APPLICATION_OCTET_STREAM;
 import static javax.ws.rs.core.MediaType.APPLICATION_OCTET_STREAM_TYPE;
@@ -109,6 +110,9 @@ import com.spotify.docker.client.messages.TopResults;
 import com.spotify.docker.client.messages.Version;
 import com.spotify.docker.client.messages.Volume;
 import com.spotify.docker.client.messages.VolumeList;
+import com.spotify.docker.client.messages.swarm.Config;
+import com.spotify.docker.client.messages.swarm.ConfigCreateResponse;
+import com.spotify.docker.client.messages.swarm.ConfigSpec;
 import com.spotify.docker.client.messages.swarm.Node;
 import com.spotify.docker.client.messages.swarm.NodeInfo;
 import com.spotify.docker.client.messages.swarm.NodeSpec;
@@ -325,6 +329,8 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   private static final GenericType<List<Task>> TASK_LIST = new GenericType<List<Task>>() { };
 
   private static final GenericType<List<Node>> NODE_LIST = new GenericType<List<Node>>() { };
+
+  private static final GenericType<List<Config>> CONFIG_LIST = new GenericType<List<Config>>() { };
 
   private static final GenericType<List<Secret>> SECRET_LIST = new GenericType<List<Secret>>() { };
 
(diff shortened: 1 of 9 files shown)


davidxia:
We need to make travis test on 17.06 (see https://github.com/spotify/docker-client/pull/848/files#diff-354f30a63fb0907d4ad57269548329e3R25), but it looks like that package isn't available yet? https://travis-ci.org/spotify/docker-client/jobs/258952809

ssullivan:
I'm not sure on the availability yet on Travis but I think the install process has changed in 17.0.6 so the .travis.sh may need to be updated to handle the newer versions

ssullivan:
I've submitted #853 with some changes to get 17.06 working

davidxia:
@ssullivan I just merged #853. Try rebasing now.

ssullivan:
Thanks! I will re base and push later this weekend

ssullivan:
@davidxia I've rebased the latest into this branch now.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1968,6 +1974,143 @@ public Task inspectTask(final String taskId) throws DockerException, Interrupted
...
+        case 503:
+          throw new NonSwarmNodeException("node is not part of a swarm", e);
+        default:
+          throw e;
+      }
+    }
+  }
+
+  @Override
+  public List<Config> listConfigs(Config.Criteria criteria)
```
davidxia: Make this final for consistency.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1968,6 +1974,143 @@ public Task inspectTask(final String taskId) throws DockerException, Interrupted
...
+      filters.put("id", Collections.singletonList(criteria.configId()));
+    }
+    if (criteria.label() != null) {
+      filters.put("label", Collections.singletonList(criteria.label()));
+    }
+    if (criteria.name() != null) {
+      filters.put("name", Collections.singletonList(criteria.name()));
+    }
+
+    WebTarget resource = resource().path("configs");
```
davidxia: You can just inline these two lines.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1968,6 +1974,143 @@ public Task inspectTask(final String taskId) throws DockerException, Interrupted
...
+    final Map<String, List<String>> filters = new HashMap<>();
+
+    if (criteria.configId() != null) {
+      filters.put("id", Collections.singletonList(criteria.configId()));
+    }
+    if (criteria.label() != null) {
+      filters.put("label", Collections.singletonList(criteria.label()));
+    }
+    if (criteria.name() != null) {
+      filters.put("name", Collections.singletonList(criteria.name()));
```
davidxia: [Docker API docs](https://docs.docker.com/engine/api/v1.30/#operation/ConfigList) say there's also `names`. What's the difference between that and this one? Should we add both?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -1968,6 +1974,143 @@ public Task inspectTask(final String taskId) throws DockerException, Interrupted
...
+        case 503:
+          throw new NonSwarmNodeException("Config not part of a swarm.", ex);
+        default:
+          throw ex;
+      }
+    }
+  }
+
+  @Override
+  public void updateConfig(String configId, Long version, ConfigSpec nodeSpec)
```
davidxia: add `final`s

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
ssullivan: @davidxia I've tried using the 'names' filter but docker 17.06 throws an error whenever I try and specify it so I only implemented the 'name' option because that appears to work. I'll try and dig into it some more to figure out what's going on.

davidxia: No worries if you can't figure it out. It's better to have some new functionality than to block on getting 100% of it.

ssullivan: I was curious so I've been doing a little digging around the moby repo and I haven't seen anything that references the 'names' filter yet in the api or tests.

ssullivan:
I've updated all of the variables, and parameters with the final keyword and rebased with master again.

davidxia:
Weird, maybe their docs are wrong.

PR #848, opened 2017-07-29
Test on latest docker versions

davidxia:
stop testing on 1.11.2


davidxia:
just testing don't merge

ssullivan:
I think to install 17.0.6 on ubuntu the package has changed from docker-engine to docker-ce so the install script in .travis.sh may need to be updated.

davidxia:
Thanks. I removed `ENABLE_SWARM`. I just like having `UPLOAD_SONATYPE` on the latest even though it doesn't make a difference. :)

davidxia:
yay, tests pass

davidxia:
bump

PR #850, opened 2017-07-29
Relocate guava classes with maven shade plugin

### LINKED ISSUE #826: Old Guava used, causes NoSuchMethodError with Guava 22.0
## Description

Currently the project uses Guava version 18.0

from pom:

```
<dependency>
   <groupId>com.google.guava</groupId>
   <artifactId>guava</artifactId>
   <version>18.0</version>
</dependency>
```

## How to reproduce

Set Guava version to 22.0
Try to compile it

or use the project in environment with new guava

## What do you expect
The application is compiled/works.

## What happened instead
java.lang.NoSuchMethodError: com.google.common.net.HostAndPort.getHostText()Ljava/lang/String;

	at com.spotify.docker.client.DefaultDockerClient.fromEnv(DefaultDockerClient.java:1697)

## JavaDoc of Guava 21

@Deprecated
public String getHostText()
Deprecated. Use getHost() instead. This method is scheduled for removal in Guava 22.0.
Old name of getHost().

https://google.github.io/guava/releases/21.0/api/docs/com/google/common/net/HostAndPort.html#getHostText--

---
Please update Guava version in this project. The currently used version 18 is more than 3 years old and it causes problems with recent Guava versions.

Thanks!

davidxia:
to avoid class conflicts with other guava versions.

Fixes #826

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -319,6 +319,7 @@
               <include>org.glassfish.**</include>
               <include>com.github.jnr:*</include>
               <include>org.ow2.asm:*</include>
+              <include>com.google.guava:**</include>
             </includes>
           </artifactSet>
           <relocations>
@@ -337,6 +338,10 @@
             <relocation>
               <pattern>com.fasterxml.jackson</pattern>
               <shadedPattern>com.spotify.docker.client.shaded.com.fasterxml.jackson</shadedPattern>
+            </relocation>
+            <relocation>
+              <pattern>com.google.guava</pattern>
+              <shadedPattern>com.spotify.docker.client.shaded.com.google.guava</shadedPattern>
             </relocation>
           </relocations>
           <shadedArtifactAttached>true</shadedArtifactAttached>
--- a/pom.xml
+++ b/pom.xml
@@ -340,8 +340,8 @@
               <shadedPattern>com.spotify.docker.client.shaded.com.fasterxml.jackson</shadedPattern>
             </relocation>
             <relocation>
-              <pattern>com.google.guava</pattern>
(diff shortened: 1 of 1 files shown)


davidxia:
@mattnworb What do you think?

pom.xml
```diff
@@ -338,6 +339,10 @@
               <pattern>com.fasterxml.jackson</pattern>
               <shadedPattern>com.spotify.docker.client.shaded.com.fasterxml.jackson</shadedPattern>
             </relocation>
+            <relocation>
+              <pattern>com.google.guava</pattern>
```
mattnworb: I think this should be `com.google.common`

see https://github.com/google/guava/tree/master/guava/src/com/google

pom.xml
```diff
@@ -338,6 +339,10 @@
...
+              <shadedPattern>com.spotify.docker.client.shaded.com.google.guava</shadedPattern>
```
mattnworb: same update needed on this line

pom.xml
```diff
@@ -319,6 +319,7 @@
               <include>org.glassfish.**</include>
               <include>com.github.jnr:*</include>
               <include>org.ow2.asm:*</include>
+              <include>com.google.guava:**</include>
```
mattnworb: I believe this will include the original Guava classes as-is in the shaded jar.

I tested it out and:

```
dxia/relocate-guava!docker-client $> unzip -l target/docker-client-8.8.3-SNAPSHOT-shaded.jar | grep GwtCompatible
      640  08-05-2017 12:48   com/google/common/annotations/GwtCompatible.class
```

so I think this line should be removed (although maybe this behavior changes if the relocation/pattern below is corrected?)

PR #852, opened 2017-07-29
Test with java 8; still build for java 7 

davidxia:


davidxia:
cc @mattnworb

davidxia:
@mattnworb Makes sense. I agree.

PR #853, opened 2017-07-30
Updates to .travis.sh to install docker-ce 17.06

ssullivan:
I was able to get docker-ce 17.06 working in Travis with the following modifications.
I chose to add an additional environment variable to make it easy to switch between the correct package to install because I wanted to avoid having to check each version in the .trasvis.sh.


davidxia:
@ssullivan Thanks so much.

PR #859, opened 2017-08-07
Add AWS Elastic Container Service authentication support

toelen:
Hi,

before further testing this implementation, I would first like to know if you would be willing to add ECR authentication to docker-client. There is a TODO to split up the GCR authentication, and ECR authentication would be a driver for this split I suppose.

Best regards,
Leen Toelen

### DIFF
--- a/src/main/java/com/spotify/docker/client/auth/ecr/ContainerRegistryAuthSupplier.java
+++ b/src/main/java/com/spotify/docker/client/auth/ecr/ContainerRegistryAuthSupplier.java
@@ -69,7 +69,7 @@ public class ContainerRegistryAuthSupplier implements RegistryAuthSupplier {
    * <p>
    * The default value for the minimum expiry time of an access token is one minute. When the
    * ContainerRegistryAuthSupplier is asked for a RegistryAuth, it will check if the existing
-   * AccessToken for the AWS AuthorizationData expires within this amount of time. If it does, then
+   * authorization token for the AWS AuthorizationData expires within this amount of time. If it does, then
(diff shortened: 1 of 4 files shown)


src/main/java/com/spotify/docker/client/auth/ecr/ContainerRegistryAuthSupplier.java
```diff
@@ -0,0 +1,230 @@
...
+
+  @Override
+  public RegistryAuth authForSwarm() throws DockerException {
+    final AuthorizationData accessToken;
+
+    try {
+      accessToken = getAccessToken();
+    } catch (IOException e) {
+      // ignore the exception, as the user may not care if swarm is authenticated to use GCR
+      log.warn("unable to get access token for Google Container Registry due to exception, "
```
johnflavin: You should change the error message here to be about AWS rather than GCR.

src/main/java/com/spotify/docker/client/auth/ecr/ContainerRegistryAuthSupplier.java
```diff
@@ -0,0 +1,230 @@
...
+  public RegistryConfigs authForBuild() throws DockerException {
+    final AuthorizationData accessToken;
+
+    try {
+      accessToken = getAccessToken();
+    } catch (IOException e) {
+      // do not fail as the GCR access token may not be necessary for building the image
+      // currently
+      // being built
+      log.warn("unable to get access token for Google Container Registry, "
```
johnflavin: Change this warning as well.

PR #864, opened 2017-08-08
travis: change how the docker gpg key is downloaded

mattnworb:
requests to p80.pool.sks-keyservers.net are timing out in recent travis
builds


davidxia:
Let's see if this one works. https://github.com/spotify/docker-client/pull/865

PR #865, opened 2017-08-09
travis: change how docker gpg key is downloaded

davidxia:
Requests to p80.pool.sks-keyservers.net are timing out in recent travis
builds. Use the ubuntu key server.


davidxia:
Yea, there must be more than one.

PR #866, opened 2017-08-09
added registry auth to update function

soulan:
if you need to update the password or credentials you have to send the new credentials with update, also important if ecr is your registry of choice because you need to update them every 12 hours


soulan:
thanks maven...

davidxia:
thanks

PR #869, opened 2017-08-16
Updated the ContainerSpec to have support for Config bindings

ssullivan:
This PR updates the existing ContainerSpec value object to have support for the specification of config bindings for docker API >= 1.30

### DIFF
--- a/src/main/java/com/spotify/docker/client/messages/swarm/ContainerSpec.java
+++ b/src/main/java/com/spotify/docker/client/messages/swarm/ContainerSpec.java
@@ -111,6 +111,13 @@ public abstract class ContainerSpec {
   @JsonProperty("Secrets")
   public abstract ImmutableList<SecretBind> secrets();
 
+  /**
+   * @since API 1.30
+   */
+  @Nullable
+  @JsonProperty("Configs")
+  public abstract ImmutableList<ConfigBind> configs();
+
   @Nullable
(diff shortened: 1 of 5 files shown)


src/main/java/com/spotify/docker/client/messages/swarm/ContainerSpec.java
```diff
@@ -1,396 +1,410 @@
-/*-
```
davidxia: Why is there such a large diff here?

ssullivan: I think this was an IDE error on my part, I will correct this.

ssullivan:
@davidxia My apologies for changing the line endings and causing the large diff. This should be corrected now.  I think this may be an inconsistency in the docker API docs. The sample JSON they have on the right does not include 'configs' and related objects in the sample but if you drill down into the Request Body for
https://docs.docker.com/engine/api/v1.30/#operation/ServiceCreate under TaskTemplate > ContainerSpec they have Configs listed. This what I based these changes off .

ssullivan:
docker released v1.31 of their API but the sample JSON for the create service API still only list secrets for ContainerSpec. Their [swagger.yaml](https://docs.docker.com/engine/api/v1.31/swagger.yaml) does document the ContainerSpec as having config options though.

PR #884, opened 2017-09-08
added relocation for the org.glassfish dependency

PinkSheep:
I ran into an issue with some clashing glassfish dependencies when using docker-client on a payara (glassfish) server. This change fixed it for me. Any reason why you added the org.glassfish pattern in the artifactSet of the shade plugin, but chose to not relocate it? If there is one feel free to ignore this pull request.


mydeadlyvenoms:
@davidxia do you have any comments on this PR?

PR #885, opened 2017-09-09
include asm and apache httpcomponents jar files for shading to avoid interop issues while using docker-client library

sanjaypujare:
@mattnworb could you review and merge if appropriate. The rationale for this is as follows: I am trying to use docker-client in my application which uses different versions of httpclient and asm jar files and I just could not make my app work with any common versions. So shading the 2 seems to be the best option. I have tested the change in my application

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -321,6 +321,9 @@
               <include>com.github.jnr:*</include>
               <include>org.ow2.asm:*</include>
               <include>com.google.guava:**</include>
+              <include>org.apache.httpcomponents:httpclient</include>
+              <include>org.apache.httpcomponents:httpcore</include>
+              <include>asm:asm</include>
             </includes>
           </artifactSet>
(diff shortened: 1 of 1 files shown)


sanjaypujare:
@davidxia @mattnworb any comments on this PR?

sanjaypujare:
@davidxia  thx. Will it merge in a day or two?

mattnworb:
The config itself looks good to me but I can't really comment on whether or not httpclient is still functional after being shaded like this. I would be ok with merging if it is understood that this might be a bit experimental

sanjaypujare:
@mattnworb I did test this after shading with a couple of tests so I believe the httpclient is functional. But I do understand your concern and I will fix any issues if they are found (I will be using this extensively myself)

mattnworb:
👍

PR #891, opened 2017-09-14
Service name can be null

johnflavin:
If you create a service with a `null` value for `name` in `ServiceSpec`, docker just assigns a name for you. As such, I've marked `ServiceSpec.name()` `@Nullable`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 8.9.1
+
+Not yet released.
+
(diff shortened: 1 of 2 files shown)


johnflavin:
Test failure appears to be a travis problem. Tons of tests reporting errors that indicate cgroups are not mounted.

johnflavin:
Yes, I can. The changelog for this project keeps slipping my mind. You'd think I could remember, since I have one in my project that I have to nag people to add their notes to.

PR #898, opened 2017-10-02
Fix serialization of sets

### LINKED ISSUE #893: Empty exposedPorts: ImmutableSet serialization not working as configured by docker-client, with Jackson 2.9.x
## Description

I am creating a container with the exposedPorts property filled. However, it's not being used:
```
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "POST /containers/create?name=test-kafka-1392566563 HTTP/1.1[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "Accept: application/json[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "Content-Type: application/json[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "User-Agent: Jersey/2.25.1 (Apache HttpClient 4.5.3)[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "Transfer-Encoding: chunked[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "Host: 192.168.99.100:2376[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "Connection: Keep-Alive[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "Accept-Encoding: gzip,deflate[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "12[\r][\n]"
10:59:20.247 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "{"ExposedPorts":{}[\r][\n]"
10:59:20.294 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "215[\r][\n]"
10:59:20.294 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> ","Env":["SSL_TRUSTSTORE_LOCATION=/var/private/ssl/server.truststore.jks","ADVERTISED_HOST=192.168.99.100","SSL_CLIENT_AUTH=required","SSL_KEYSTORE_LOCATION=/var/private/ssl/server.keystore.jks","ADVERTISED_SSL_PORT=22480","ADVERTISED_PORT=17186"],"Cmd":[],"Image":"jolivares/kafka:latest","Volumes":{"/tmp/some-strange-volume":{}},"HostConfig":{"Binds":["/Users/flozano/Projects/java/platform-common/test-container/bin/certs:/var/private/ssl"],"PortBindings":{},"PublishAllPorts":true,"NetworkMode":"docker-container-provider-test"}}[\r][\n]"
10:59:20.294 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "0[\r][\n]"
10:59:20.294 [jersey-client-async-executor-1] DEBUG org.apache.http.wire - http-outgoing-3 >> "[\r][\n]"
```

I am checking thru the debugger however that the ContainerConfig contains the right set of values.

I have Jackson 2.9.1 and Jersey 2.25.1 in classpath, and I can't use lower versions because of other dependencies. My guess is that it's related to this...  with older versions (docker-client 8.8.0, jackson 2.8.9, hk2-2.5.0-b38) it didn't happen.

johnflavin:
All `Set`s are now serialized as maps of keys to empty maps, rather than maps of keys to `null`.

This specific motivating case is `ContainerConfig.exposedPorts`, but every other `Set` in the model should have the same behavior. As far as I am aware we only use `Set`s to model docker's "map of keys to empty objects" API pattern. I infer they use this pattern as a workaround for a lack of a native set type in Go.

I also refactored `ContainerConfig.volumes` to be a `Set` rather than being a hard-coded `Map<String, Map>`. I had implemented `volumes` as the latter because I was unaware we were using a custom serializer for `Set` to reproduce docker's API pattern. Refactoring `volumes` to also be a `Set` simplifies the code but leaves the serialization the same.

While I was in `ContainerConfig`, I cleaned up all the `if (foo != null) { builder.foo(foo); }` statements for properties that are `@Nullable`. AutoValue doesn't care if you feed a `null` into those builder methods, so we don't need to check.

Fixes #893.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 Not yet released.
 
 - `ContainerConfig.exposedPorts` is serialized as a map of keys to empty map placeholders. `exposedPorts` had been serialized as keys -> `null`. (Fixes [893][])
-- Breaking API change: The method `ContainerConfig.volumes()` has changed signature from `ImmutableMap<String, Map> volmues()` (where the second `Map` was always empty) to `Set<String> volumes()`. This change simplifies the code internally. ([898][])
+- Breaking API change: The method `ContainerConfig.volumes()` has changed signature from `ImmutableMap<String, Map> volumes()` (where the second `Map` was always empty) to `ImmutableSet<String> volumes()`. This change simplifies the code internally. ([898][])
 
 [893]: https://github.com/spotify/docker-client/issues/893
 [898]: https://github.com/spotify/docker-client/issues/898
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 8.10.0
+
+Not yet released.
+
+- `ContainerConfig.exposedPorts` is serialized as a map of keys to empty map placeholders. `exposedPorts` had been serialized as keys -> `null`. (Fixes [893][])
+- Breaking API change: The method `ContainerConfig.volumes()` has changed signature from `ImmutableMap<String, Map> volmues()` (where the second `Map` was always empty) to `Set<String> volumes()`. This change simplifies the code internally. ([898][])
+
+[893]: https://github.com/spotify/docker-client/issues/893
+[898]: https://github.com/spotify/docker-client/issues/898
+
 ## 8.9.2
 
 Released November 15, 2017
--- a/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
@@ -267,8 +267,25 @@ public abstract class ContainerConfig {
 
     public abstract Builder volumes(final String... volumes);
 
+    /**
+     * @deprecated As of 8.10.0, use {@link #volumes(Set)} or
+     *             {@link #volumes(String...)}.
+     */
+    @Deprecated
+    public Builder volumes(final Map<String, Map> volumes) {
+      this.volumes(volumes.keySet());
+      return this;
+    }
+
     public Builder addVolume(final String volume) {
       volumesBuilder().add(volume);
+      return this;
+    }
+
+    public Builder addVolumes(final String... volumes) {
+      for (final String volume : volumes) {
+        volumesBuilder().add(volume);
+      }
       return this;
     }
 
--- a/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
@@ -265,14 +265,7 @@ public abstract class ContainerConfig {
 
     public abstract Builder volumes(final Set<String> volumes);
 
-    public Builder volumes(final String... volumes) {
-      if (volumes != null) {
-        for (final String volume : volumes) {
-          addVolume(volume);
-        }
-      }
-      return this;
-    }
+    public abstract Builder volumes(final String... volumes);
 
     public Builder addVolume(final String volume) {
       volumesBuilder().add(volume);
--- a/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
@@ -101,7 +101,7 @@ public abstract class ContainerConfig {
   public abstract String image();
 
   /**
-   * @deprecated As of 8.9.1, use {@link #volumes()}.
+   * @deprecated As of 8.10.0, use {@link #volumes()}.
    */
   @Deprecated
   public Set<String> volumeNames() {
--- a/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
+++ b/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
@@ -263,20 +263,6 @@ public abstract class ContainerConfig {
 
     abstract ImmutableSet.Builder<String> volumesBuilder();
 
-    public abstract Builder volumes(final Set<String> volumes);
-
-    public abstract Builder volumes(final String... volumes);
-
-    /**
-     * @deprecated As of 8.10.0, use {@link #volumes(Set)} or
-     *             {@link #volumes(String...)}.
-     */
-    @Deprecated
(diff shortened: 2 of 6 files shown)


flozano:
One Q, when is this going to be merged and released?

CHANGELOG.md
```diff
@@ -1,9 +1,17 @@
 # Changelog
 
-## 8.9.1
+## 8.9.2
```
davidxia: Would this need to be a major version bump because we're changing Map to Set [here](https://github.com/spotify/docker-client/pull/898/files#diff-30a967cd2986f001106d3df2023eb161R113)?

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
```diff
@@ -201,37 +203,17 @@ static ContainerConfig create(
         .macAddress(macAddress)
         .hostConfig(hostConfig)
         .stopSignal(stopSignal)
-        .networkingConfig(networkingConfig);
-
-    if (portSpecs != null) {
```
davidxia: Thanks. Don't know why I did this.

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
```diff
@@ -265,14 +265,7 @@ public static Builder builder() {
...
 
-    public Builder volumes(final String... volumes) {
-      if (volumes != null) {
-        for (final String volume : volumes) {
-          addVolume(volume);
-        }
-      }
-      return this;
-    }
+    public abstract Builder volumes(final String... volumes);
```
davidxia: Just wanted to point out that the behavior is different here. This replaces whereas the previous version was cumulative. I prefer this one, but it would be backwards incompatible. We should point this out in the changelog if we merge this.

davidxia:
@johnflavin Thanks so much. See comments above.

davidxia:
Can you also add [this commit for testing](https://github.com/spotify/docker-client/commit/580313c9157c48fa98252c33adee82f0fc5121ad) right before the commit where you remove the null checks?

I removed some null checks in other autovalue classes and found the deserialization sometimes changes from the attribute being null to being an empty map. See https://github.com/spotify/docker-client/pull/925/files#r150894093

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
johnflavin: No, this is the same. The old code had a method with the same signature which also replaced the volumes.

This method has been removed, but had the same signature as the new abstract method. I can show that it did replace and not accumulate. It was explicitly implemented to turned the varargs into a `Set` and pass it to another method:
```java
/**		 
 * @deprecated  As of release 7.0.0, replaced by {@link #volumes(Map)}.
 */		
@Deprecated		
public Builder volumes(final String... volumes) {		
  if (volumes != null && volumes.length > 0) {		
    volumes(ImmutableSet.copyOf(volumes));		
  }		
  return this;
}
```
Just from that we can't tell if it replaced or accumulated. But we can look at that second Builder method, the one that took a `Set` and turned it into a `Map<String, Map>`:
```java
/**
 * @deprecated  As of release 7.0.0, replaced by {@link #volumes(Map)}.		
 */		
@Deprecated		
public Builder volumes(final Set<String> volumes) {		
  if (volumes != null && !volumes.isEmpty()) {		
    final ImmutableMap.Builder<String, Map> volumesBuilder = ImmutableMap.builder();		
    for (final String volume : volumes) {		
      volumesBuilder.put(volume, new HashMap());		
    }		
    volumes(volumesBuilder.build());		
  }		
  return this;		
}
```

This passes the `Map<String, Map>` into the abstract Builder method that replaces the volumes. So the whole chain, from `volumes(String... volumes)`, all were replacing any existing contents.

The new method does the same.

CHANGELOG.md
johnflavin: I think you’re right. I had thought that I kept the API the same and just shuffled things around internally, but this is an API change.

Should I bump this line and the version in `pom.xml` up to `8.10.0`?

johnflavin: I guess, regardless of what I change the version to, I cannot make it `8.9.2`. That version has already been released.

src/main/java/com/spotify/docker/client/messages/ContainerConfig.java
johnflavin: There was a method that accumulated: `addVolumes(String...)`. That one got removed. I'll add it back.

I'll also add back what had been the primary `Builder` method for `volumes`: `volumes(Map<String, Map>)`. I removed that, but I should have deprecated it.

johnflavin:
Summary of recent changes:

* Rebased on master.
* Bumped version (in `CHANGELOG`, not yet in `pom`) to `8.10.0` in response to [comment](https://github.com/spotify/docker-client/pull/898#discussion_r150459627).
* Noted breaking API change in `CHANGELOG`. [ibid.](https://github.com/spotify/docker-client/pull/898#discussion_r150459627)
* Added back two `ContainerConfig.Builder` methods that had been removed: `addVolumes(String...)` which accumulates volumes, and a deprecated `volumes(Map<String, Map>)`. See [comment](https://github.com/spotify/docker-client/pull/898#discussion_r152007309).

davidxia:
Thanks for the fixes.

PR #904, opened 2017-10-17
improve e.g. code in README by replace command with one that works

harschware:
README example for running command inside running container yields:
```
2017-10-17 10:00:04.545 DEBUG main:TestDockerApi:107 - Output of 'ls' in container: oci runtime error: exec failed: container_linux.go:265: starting container process caused "exec: \"bash\": executable file not found in $PATH"
```
replace it with "du - h -s" which is a more interesting output and works..

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ final ContainerInfo info = docker.inspectContainer(id);
 docker.startContainer(id);
 
 // Exec command inside running container with attached STDOUT and STDERR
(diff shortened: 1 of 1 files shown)


davidxia:
Thanks. How about just replacing `bash` with `sh`? That works.

harschware:
@davidxia  'sh' works too...  I've pushed a change

PR #906, opened 2017-10-21
Added spread over support in task template placement preference

jeanlouistournay:
Support of placement preference and spread over in Task template to provide a way to make the scheduler aware of factors such as topology and node label (since Docker API 1.28)
 
For retro compatibility with previous version, public static create(constraints) method has been kept. A unit test with spread over has been added.

### DIFF
--- a/src/test/java/com/spotify/docker/client/messages/swarm/PlacementPrefTest.java
+++ b/src/test/java/com/spotify/docker/client/messages/swarm/PlacementPrefTest.java
@@ -1,0 +1,43 @@
+/*-
+ * -\-\-
+ * docker-client
+ * --
+ * Copyright (C) 2016 - 2017 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.docker.client.messages.swarm;
+
+import static com.spotify.docker.FixtureUtil.fixture;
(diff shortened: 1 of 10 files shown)


src/main/java/com/spotify/docker/client/messages/swarm/Placement.java
```diff
@@ -41,10 +41,46 @@
   @JsonProperty("Constraints")
   public abstract ImmutableList<String> constraints();
 
+  @Nullable
+  @JsonProperty("Preferences")
+  public abstract ImmutableList<PlacementPreference> preferences();
```
davidxia: Can you add a javadoc stating when this is available? See https://github.com/spotify/docker-client/blob/master/src/main/java/com/spotify/docker/client/messages/HostConfig.java#L222-L226

src/test/resources/fixtures/1.28/serviceSpec.json
```diff
@@ -0,0 +1,32 @@
...
+    "ContainerSpec" : {
+      "Image" : "image:latest",
+      "Labels" : {},
+      "Env" : [],
+      "Secrets" : [ ]
+    },
+    "Placement" : {
+      "Preferences":[
+        { "Spread":
+        { "SpreadDescriptor" :"node.labels.test" }
```
davidxia: indent one more level

src/test/java/com/spotify/docker/client/messages/swarm/PlacementPrefTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static com.spotify.docker.FixtureUtil.fixture;
+import static org.junit.Assert.assertNotNull;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.spotify.docker.client.ObjectMapperProvider;
+import org.junit.Test;
+
+
+public class PlacementPrefTest {
```
davidxia: Name this ServiceSpecTets instead?

ssullivan:
I would find the additions of placement preferences very useful. I have a closed PR at the moment that only makes similar modifications without the additional image, and volume modification methods. Since this hasn’t been changed in several months should I re-open my closed PR?

mattnworb:
I merged #981 which seems to implement the same thing (although I am not familiar enough with the underlying feature to say 100% for sure). Please feel free to re-open if I am wrong.

PR #910, opened 2017-10-30
Support loading auth config for repos with no scheme in their URL

### LINKED ISSUE #804: Pushing to private registry fails with 'No basic auth' for older style config.json
## Description

Pushing to private registry fails with 'No basic auth', even if there are credentials for this registry stored in ~/.docker/config.json

Example:

    {
        "auths": {
		"https://some.custom.repository.tld": {
			"auth": "**secret**"
		},
		"https://index.docker.io/v1/": {
			"auth": "**secret**"
		}
        }
    }

Configuration of dockerfile-maven:

    <configuration>
        <repository>some.custom.repository.tld/${project.artifactId}</repository>
    </configuration>

It seems this is because previous versions of docker used to store auth credentials together with repository URL containing transport part (https://some.custom.repository.tld instead of some.custom.repository.tld)

So this currently breaks builds using recent versions of dockerfile-maven, docker-maven plugins ( (>=1.0.0),  i.e. on Shippable, or whenever an older docker was used to issue `docker login`.

In case this is useful to someone, a work-around for Shippable seems to be:

      ci:
       - sed -ie 's/https:\/\///g' ~/.docker/config.json

## What do you expect

Could you just strip transport part out of ~/.docker/config.json, if present?

56quarters:
Support loading auth configuration for repos when their `serverAddress`
is specified with no scheme component but their entry in the configuration
file includes a scheme component.

Older versions of Docker write authentication tokens to the configuration
file including their scheme ("https://") but newer onces do not. Thus
in order to allow people to use the newer format (looking up auth for
a repo that isn't prefixed with scheme) with an older configuration file
we try prefixing the `serverAddress` with "https://" or "http://" in some cases.

Fixes #804

### DIFF
--- a/src/main/java/com/spotify/docker/client/DockerConfigReader.java
+++ b/src/main/java/com/spotify/docker/client/DockerConfigReader.java
@@ -36,6 +36,7 @@ import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


src/main/java/com/spotify/docker/client/DockerConfigReader.java
```diff
@@ -86,6 +85,25 @@ private RegistryAuth parseDockerConfig(final Path configPath, final String serve
...
+      return configs.get(serverAddress);
+    }
+
+    // If the given server address didn't have a protocol try adding the 'https' protocol.
+    // This handles cases where older versions of Docker included the protocol when writing
+    // auth tokens to config.json.
+    try {
+      final URI serverAddressUri = new URI(serverAddress);
+      if (serverAddressUri.getScheme() == null) {
+        final String addrWithProto = "https://" + serverAddress;
```
davidxia: Let's also check with `http://`.

56quarters: Sure, will do.

davidxia:
@mattnworb @caipre How does this look?

demaniak:
Hi!

Has this been released yet?

davidxia:
Yes, should be in latest docker-client.

Issue #756, opened 2017-05-22
start requiring Java 8?


mattnworb:
Java 7 has been [EOL since April 2015](http://www.oracle.com/technetwork/java/eol-135779.html). Do we still need to support it in this library, i.e. compile for target=1.7 source=1.7?

The motivation for this question is that I started to use [our hamcrest-jackson](https://github.com/spotify/java-hamcrest) library in some unit tests in #753 but forgot that this library is built with Java 7 and hamcrest-jackson is built with 8.

Is there anyone using this that is stuck on 7 for some reason, or some other argument to be made against moving to compiling with 8?

johnflavin:
My application is still on java 7. We plan to move to java 8 at some point, but it is never the top priority.

However, I wouldn't want you to hold up progress on my account. My needs are well served by the existing functionality in `docker-client`, so if I had to depend on an older version that supported java 7 it wouldn't bother me.

mattnworb:
We could also explore building for both 7 and 8, using some maven profile magic, and upload both to Maven Central (with a classifier for one of jdk8 or jdk7). 

At this point all I am really after is being able to use some 8-only libraries *in tests*, and not using 8 language features or standard library classes, so perhaps it would be enough to run tests with jdk8 (to allow classes compiled for 8 to load) but build for 7 only (source/target=7 should catch use of 8 language features, but not catch use of 8 classes like Stream etc AFAIK).

davidxia:
I'm fine with using 8 for tests only. It'd be nice to just build new versions for 8 as well, but I'm not sure how many people will complain.

mattnworb:
Another bonus of using Java 8 is I think we could convert all usages of ListenableFuture to CompletableFuture and possibly ditch the Guava dependency altogether.

hekonsek:
Keep in mind that some of your dependencies (like Jersey #915) has already moved to Java8. In practice that would mean that you can't upgrade those dependencies as long as you stick to Java 7. Also some other libraries which happen to use those latest, Java8-based versions of dependencies will be conflicting with docker-client. Please do consider upgrading to Java8. :)

mattnworb:
I happened to notice today that Oracle will end public updates to Java SE 8 in September 2018, so with less than one year left of real "life", I think it is overdue to set this as a minimum version.

diemol:
Can we help with this? I am running often to something similar to #877, and it would be nice to start updating the dependencies.
[...]

PR #915, opened 2017-11-03
Upgrade Jersey version

hekonsek:
Hi,

Would it be possible to upgrade Jersey/Glassfish version? Right now docker-client conflicts with applications using the most recent version of Jersey/Glassfish.

Upgrading Jersey to the most recent version requires aligning jackson-annotations to the same version as jackson-core (which is good practice anyway) and adding hk2 dependencies explicitly to your POM.

Thanks!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,19 +102,34 @@
       <version>2.8.8</version>
     </dependency>
     <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
+      <version>2.8.8</version>
+    </dependency>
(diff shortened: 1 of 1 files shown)


hekonsek:
Apparently the latest Jersey version requires Java >= 8. Have you considered upgrading minimal Java version to 8? You will have more and more libraries conflicting with old Jersey version. For example right now Spring Boot 2.0 is conflicting with your library already :( .

davidxia:
@hekonsek We've considered using [java 8 here](https://github.com/spotify/docker-client/issues/756), but we haven't decided yet. If we do, it'll require some bit of work.

PR #916, opened 2017-11-07
execStart should use the 'noTimeoutClient'.

rgrunber:
Like attachContainer, the execution created by execStart could take an
arbitrary amount of time, and if the inactivity execeeds the socket read
timeout (default 30s), data could be overlooked, or the stream could be
closed.

For example, if one has a running container, and creates a separate execution instance (running /bin/sh) using execCreate/execStart, after 30s of inactivity on the returned stream, it will time out and be unusable.

The test in this commit reproduces this by setting the timeout to 5s and creating a separate execution that sleeps for 10s and then prints "Finished". When execStart uses the regular resource() call, this fails due to a timeout after 5s, but with noTimeoutResource() this succeeds.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -1616,7 +1616,7 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   @Override
   public LogStream execStart(final String execId, final ExecStartParameter... params)
       throws DockerException, InterruptedException {
-    final WebTarget resource = resource().path("exec").path(execId).path("start");
+    final WebTarget resource = noTimeoutResource().path("exec").path(execId).path("start");
 
     final StringWriter writer = new StringWriter();
     try {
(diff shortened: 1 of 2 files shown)


caipre:
We've seen problems potentially related to this when using `exec` based commands: if the command tries to examine its environment with [`isatty(3)`](http://man7.org/linux/man-pages/man3/isatty.3.html), the request may block indefinitely.

@rgrunber Is it possible that the timeout and subsequent request closure used by the previous client allowed the `exec` command's output and exit code to be captured? I'll have to do some pseudo-tty spelunking.

rgrunber:
Is there a reason to use something that does 'isatty' as a command in the environment as opposed to using execInspect(execId).processConfig().tty() ?

Is there a test case that could reproduce this ? I've tried just taking [testExecNoTimeout](https://github.com/spotify/docker-client/blob/master/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java#L3130) and then changing the "echo Finished" part to just calling tty but it always seems to finish. the result of 'tty' always seems to be 'not a tty' even when I've attempted to pass ExecCreateParam.tty() to execCreate.

PR #918, opened 2017-11-12
Upgrade to jnr-unixsocket 0.18

davidxia:
Use upstream UnixSocket instead of our own ApacheUnixSocket wrapper.


caipre:
I guess the comment is no longer valid re: delaying the socket options?

https://github.com/spotify/docker-client/pull/918/files#diff-a479355e73762bc07b466a285d105088L41

I'm fine with this given that tests pass.

davidxia:
@caipre Yea, I think `jnr.unixsocket.UnixSocket` does the right thing compared to `jnr.unixsocket.UnixSocketChannel`, but I haven't dug into it deeply. I just [noticed this as well](https://github.com/spotify/docker-client/pull/103/files#r151032803).

src/test/java/com/spotify/docker/client/UnixConnectionSocketFactoryTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class UnixConnectionSocketFactoryTest {
+
+  @Rule
+  public ExpectedException exception = ExpectedException.none();
+
+  private static UnixConnectionSocketFactory sut;
```
mattnworb: I don't think this needs to be `static` when it is reused in all `@Test`s

davidxia: ah yes, changed

PR #919, opened 2017-11-12
Fix test for recent docker versions

davidxia:
Fix DefaultDockerClientTest.testSecretOperations()

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -4949,8 +4949,9 @@ public class DefaultDockerClientTest {
     try {
       sut.createSecret(secretSpecConflict);
       fail("Should fail due to secret name conflict");
-    } catch (DockerRequestException ex) {
-      // Ignored; Docker should return status code 409, but it doesn't for some reason.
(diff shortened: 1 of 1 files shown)


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4949,8 +4949,9 @@ public void testSecretOperations() throws Exception {
     try {
       sut.createSecret(secretSpecConflict);
       fail("Should fail due to secret name conflict");
-    } catch (DockerRequestException ex) {
-      // Ignored; Docker should return status code 409, but it doesn't for some reason.
+    } catch (ConflictException | DockerRequestException ex) {
+      // Ignored; Docker should return status code 409 but doesn't in some earlier versions.
```
caipre: Do we know at which version this exception started being thrown?

davidxia: Creating secrets was introduced in API v1.25 but was returning HTTP status code 400 (becomes DockerRequestException) instead of the 409 (becomes ConflictException) as specified in the docs.

I don't know which API version correctly starts returning 409. All I know is that on my local docker instance with API 1.33, it's 409.

PR #920, opened 2017-11-12
travis: test on more recent 17.09

davidxia:
instead of 17.04 and 17.05. Testing on quarterly releases
17.03 and .06 should be good enough for previous versions.
Testing on latest two releases will help keep this library
up to date with docker changes.

### DIFF
--- a/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
+++ b/src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
@@ -4220,10 +4220,14 @@ public class DefaultDockerClientTest {
       assertThat(e.getContainerId(), equalTo(id2));
       assertThat(e.getNewName(), equalToIgnoreLeadingSlash(newName));
     } catch (DockerRequestException ignored) {
-      // This is a docker bug. It responds with HTTP 500 when it should be HTTP 409.
+      // This is a docker bug. Docker responds with HTTP 500 when it should be HTTP 409.
       // See https://github.com/docker/docker/issues/21016.
-      // Fixed in version 1.11.0, so we should see a lower version.
-      assertThat(compareVersion(sut.version().version(), "1.11.0"), lessThan(0));
+      // Fixed in version 1.11.0 API version 1.23. So we should see a lower API version here.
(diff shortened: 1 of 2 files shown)


src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
```diff
@@ -4662,6 +4666,8 @@ public void testReadonlyRootfs() throws Exception {
   public void testStorageOpt() throws Exception {
     requireDockerApiVersionAtLeast("1.24", "StorageOpt");
     requireStorageDriverNotAufs();
+    // Doesn't work on Travis because storage driver doesn't have pquota mount option enabled.
+    assumeFalse(dockerApiVersionEquals("1.32") && TRAVIS);
```
caipre: A boolean OR seems more logical here, to me. If either are true, we want to skip the test.

davidxia: Oh it’s meant to skip if both on Travis and that version.

davidxia: I've clarified this comment.

caipre: _Only_ this version is broken? Is that documented on https://travis-ci.org/ anywhere?

davidxia:
So this version requires pquota to be set regardless of Travis or not. We
can set it on a host. But to be activated I think it requires a reboot of
the host. This obviously doesn’t work on Travis. That’s why the cobination
of the version and Travis don’t work.
-- 
David Xia
Software Engineer
dxia@spotify.com

davidxia:
@caipre Let me know if this is good now.

src/test/java/com/spotify/docker/client/DefaultDockerClientTest.java
davidxia: So this version requires pquota to be set regardless of Travis or not. We
can set it on a host. But to be activated I think it requires a reboot of
the host. This obviously doesn’t work on Travis. That’s why the cobination
of the version and Travis don’t work.

PR #925, opened 2017-11-14
Remove unnecessary null checks in AutoValue builders

davidxia:


src/main/java/com/spotify/docker/client/messages/swarm/ContainerSpec.java
```diff
@@ -387,24 +387,15 @@ static ContainerSpec create(
         .stopGracePeriod(stopGracePeriod)
         .healthcheck(healthcheck)
         .hosts(hosts)
-        .dnsConfig(dnsConfig);
+        .dnsConfig(dnsConfig)
+        .command(command)
+        .secrets(secrets)
+        .configs(configs);
 
     if (labels != null) {
```
davidxia: The default value set by autovalue is an empty map. If I remove this null check, labels changes from being an empty map to being null. This would be a backwards breaking change. See [test here](https://github.com/spotify/docker-client/pull/925/files#diff-61d18acd266c4b5bdd95e6297dd30d39R41) which passes before any changes are made to this class. The test fails if you remove the null check here.

davidxia: Turns out this only happens if you have a [builder property that accumulates values](https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#-let-my-builder-accumulate-values-for-a-collection-valued-property-not-require-them-all-at-once).

davidxia: The default value set by autovalue is an empty map. If I remove this null check, labels changes from being an empty map to being null. This would be a backwards breaking change. See [test here](https://github.com/spotify/docker-client/pull/925/files#diff-61d18acd266c4b5bdd95e6297dd30d39R41) which passes before any changes are made to this class. The test fails if you remove the null check here.

Turns out the default is an empty collection only if you have a [builder property that accumulates values](https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#-let-my-builder-accumulate-values-for-a-collection-valued-property-not-require-them-all-at-once).

PR #944, opened 2017-12-04
Fix containerAction. It was not considering queryParams

sahadevanrakesh:
While testing killContainer, ran into issue that "signal=SIGHUP" is not getting POST'ed in the URL queryParams. The below change fixes it.


davidxia:
@sahadevanrakesh Thanks for the fix. I've added a unit test as well here https://github.com/spotify/docker-client/pull/961

Issue #657, opened 2017-03-02
support for credsStore or credential-helper in new docker versions


johnflavin:
The current implementation of `RegistryAuth` has several static builder methods that parse the credentials out of the supplied or default `config.json` file. That strategy does not work in the newest docker version ([v17.03.0-ce](https://github.com/docker/docker/releases/tag/v17.03.0-ce)), as the structure of that file has changed and the credentials are no longer be stored there. They are stored in a more secure system-native way (i.e. the keychain on macOS) using the [docker credential helpers](https://github.com/docker/docker-credential-helpers). 

These changes have already rolled out on the beta channel of Docker for Mac. When you use `docker login` to store your credentials, they are put into the keychain. The `config.json` file looks like this:
```bash
$ cat ~/.docker/config.json
{
	"auths": {
		"https://index.docker.io/v1/": {}
	},
	"credsStore": "osxkeychain"
}
```

So that presents a problem for Remote API users like us. We can no longer read the credentials from a file. I don't know exactly how to interact with the credential helpers, whether they have their own API or if we need to use system calls to execute the binary. It is possible that there is already some easy solution and I just don't know what it is.

I want to open this ticket to discuss the coming changes and brainstorm solutions.

mattnworb:
Good catch, and thanks for opening the issue up.

mattnworb:
seems like the API for the credential helpers is defined at the bottom of https://github.com/docker/docker-credential-helpers#development

NicoTexas:
This is true also on Windows as Docker login is storing in wincred.
`{
	"auths": {
		"docker-dev.domain.com": {}
	},
	"credsStore": "wincred"
}`

mattnworb:
@NicoTexas yea there is no support in docker-client for credsStore or helpers yet.

I think it would be possible to add support for the credential helpers to this library, since the API seems pretty well established, although it would feel weird that this library might end up execing random processes on the user's host.

beihaifeiwu:
I got the fllow exception on windows after docker upgrade:
```bash
[INFO] --- dockerfile-maven-plugin:1.3.0:build (default) @ nagrand-ws ---
[INFO] Building Docker context F:\Source\nagrand\nagrand-ws
[INFO]
[INFO] Image will be built as 10.0.25.21:5000/nagrand-ws:2.1-SNAPSHOT
[INFO]
[WARNING] An attempt failed, will retry 1 more times
org.apache.maven.plugin.MojoExecutionException: Could not build image
[...]

PR #945, opened 2017-12-08
Add support for credsStore in new docker versions

smythie86:
This is a fix for this issue: https://github.com/spotify/docker-client/issues/657


gavinbunney:
Any update on this being merged in? `credsStore` support would be 💯

smythie86:
As far as I know, I am still waiting for this to be reviewed.

ssozonoff:
It would be amazing if this could get pushed through ....

lhanson:
Is there anything else that can be done to get this reviewed? This would unblock a lot of peoples' workflows.

mattnworb:
version 8.11.2 should be on the way to maven central now

yuelipku:
Does it mean we could now use `dockerfile-maven` plugin with `credHelpers`?
I used the aws ecr credential helper but the dockerfile-maven could not recognize that.

PR #954, opened 2017-12-27
Map 'Options' json property for volume inspection

jendib:
The 'Options' JSON property is not mapped for GET /volumes/{name}, this PR adds this mapping.


davidxia:
Thanks for the PR. I added a test here https://github.com/spotify/docker-client/pull/962

PR #903, opened 2017-10-17
#890 fix for proxy usage

deki:
fixes #890
http.proxyPort is supposed to be specified without http://, nevertheless I've added a check for it. Another option would be to set URI object instead.

Issue #890, opened 2017-09-13
build failed: Host name may not be null 


gr33npr:
## Description

While building a docker-image, which included the dockerfile-maven-plugin > 1.3.1 and an activated proxy in mavens setting.xml, the build will fail with the fhe following message:

org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.spotify:dockerfile-maven-plugin:1.3.4:build (default) on project XXXXX: Execution default of goal com.spotify:dockerfile-maven-plugin:1.3.4:build failed: Host name may not be null

If i remove the proxy-part or use an older version of dockerfile-maven-plugin, everything is fine. I`ve read that dockerfile-maven-plugin is using docker-client 8.8.X instead 8.7.X which added proxy-support, so i thought it could be 

## How to reproduce

Define a pom which includes dockerfile-maven-plugin/docker-client, define "build" as goal and set a proxy in mavens settings.xml

## What do you expect

BUILD SUCCESS

## What happened instead

BUILD FAILED

## Software:

- `docker version`: [Add the output of `docker version` here, both client and server]
- Spotify's docker-client version: [Add docker-client version here]

## Full backtrace

```text
com.spotify:dockerfile-maven-plugin:1.3.4:build failed: Host name may not be null
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:225)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:84)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:59)
	at org.apache.maven.lifecycle.internal.LifecycleStarter.singleThreadedBuild(LifecycleStarter.java:183)
	at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:161)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:320)
	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:156)
	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:537)
	at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:196)
[...]

PR #970, opened 2018-01-25
#890 fix for proxy usage ("Expected authority at index 7" error)

deki:
Unfortunately the last optimization in my PR #903 causes "Expected authority at index 7" error due to missing brackets. I've fixed that and also added a unittest for it.


deki:
@davidxia why this has not been merged before the 8.11.0 release? :-(

davidxia:
Sorry, I must've missed this. Thanks for fixing.

PR #971, opened 2018-01-25
add support for http.proxyUser and http.proxyPassword

deki:
@Tcharl requested in #903 to add user and password to authenticate.


davidxia:
@deki Thanks!

PR #981, opened 2018-02-09
Add support for Service placement preferences

ssullivan:
Placement preferences have been available in the Docker API since [1.28.](https://docs.docker.com/engine/api/v1.28/#operation/ServiceList)

This PR modifies the existing Placement object and introduces two new objects Preference, and Spread to support this feature in the docker client.

I added an additional method to the Placement object to not break compatibility for previous versions of this library.


ssullivan:
I didn’t see the existing PR that does the same thing so I am going to close this.

alokhan:
I am really interested by this feature as well and would like to avoid creating a fork. Can you please review this PR and merge it ?

PR #985, opened 2018-02-18
Update ServiceSpec.java

man4j:
add ServiceSpec.Builder builder(ServiceSpec source) constructor


src/main/java/com/spotify/docker/client/messages/swarm/ServiceSpec.java
```diff
@@ -178,6 +178,10 @@ public Builder withEndpointSpec(final EndpointSpec endpointSpec) {
   public static ServiceSpec.Builder builder() {
     return new AutoValue_ServiceSpec.Builder();
   }
+  
+  public static ServiceSpec.Builder builder(ServiceSpec source) {
+      return new AutoValue_ServiceSpec.Builder(source);
```
davidxia: Tests failed because this should be indented 4 not 6.

Issue #978, opened 2018-02-02
Docker client does not honour the -Dhttp.nonProxyHosts setting when it builds an image.


ytheohar:
## Description

The docker client does not honour the -Dhttp.nonProxyHosts setting, when it builds an image.

If the http.proxyHost is set, the docker client will go to the proxy even if DOCKER_HOST is in the http.nonProxyHosts.

## How to reproduce
Provide the following settings for maven or java. One way to do this is to set MAVEN_OPTS or _JAVA_OPTIONS environmental variable.

-Dhttp.proxyHost=myProxyUrl -Dhttp.proxyPort=myProxyPort -Dhttps.proxyHost=myProxyUrl -Dhttps.proxyPort=myProxyPort -Dhttp.nonProxyHosts="localhost|127.0.0.1|192.168.*"

and run:
mvn dockerfile:build

The problem lies in the DefaultDockerClient.updateProxy() method. If the "http.proxyHost" is set it uses the proxy without looking at if the "http.nonProxyHosts" setting includes the docker host.

## What do you expect

Successful build.

## What happened instead

The HTTP response fails with 403:

ClientResponse{method=GET, uri=https://192.168.99.100:2376/version, status=403, reason=Forbidden}
Headers in the response:
{Server=[squid], Mime-Version=[1.0], Date=[Fri, 02 Feb 2018 15:21:59 GMT], Content-Type=[text/html;charset=utf-8], Content-Length=[2062], X-Squid-Error=[ERR_ACCESS_DENIED 0], Vary=[Accept-Language], Content-Language=[en], X-Cache=[MISS from proxy-1_6], Via=[1.1 proxy-1_6:8006], Connection=[keep-alive]}

## Software:

- `docker version`: [18.01.0-ce, build 03596f51b1] 
- Spotify's docker-client version: [8.8.1] This is the version used by dockerfile-maven-plugin:1.3.7

## Full backtrace

```text
[WARNING] An attempt failed, will retry 1 more times
org.apache.maven.plugin.MojoExecutionException: Could not build image
        at com.spotify.plugin.dockerfile.BuildMojo.buildImage(BuildMojo.java:185)
        at com.spotify.plugin.dockerfile.BuildMojo.execute(BuildMojo.java:105)
        at com.spotify.plugin.dockerfile.AbstractDockerMojo.tryExecute(AbstractDockerMojo.java:246)
[...]

PR #986, opened 2018-02-28
Fix for #978: Docker client does not honour the -Dhttp.nonProxyHosts …

beskow:
…setting when it builds an image.


beskow:
Any news regarding this pull request?

Issue #887, opened 2017-09-10
Invalid RSA private key when using Docker client with maven dockerfile:build with Docker Data Center 17.06


blueelephants:
## Description
I have created my client bundle in Docker Data Center which consists of these files:
```
ca.pem
cert.pem
cert.pub
env.cmd
env.ps1
env.sh
key.pem
```

Where `key.pem` is an "EC PRIVATE KEY".

When I load client bundle (via `source ./env.sh`) the environment variables are set correctly:
```
DOCKER_HOST=tcp://dckdftest01.cloud.mycompany.com:443
DOCKER_TLS_VERIFY=1
DOCKER_CERT_PATH=/home/myUsera/ucp/ucp-bundle-myUser-dckdftest01
```
This means, I can run commands like `docker info, docker ps, docker build,` etc. without any problems on the remote Docker cluster.

However when I use dockerfile-maven-plugin to build a Docker image on the remote Docker cluster, I get an "Invalid RAS private key" error message.

## How to reproduce

1. `source ./env.sh`
2. change to your Java Maven project with dockerfile pluging configured
    My plugin configuration in Maven is straight-forward:
   ```
			<plugin>
				<groupId>com.spotify</groupId>
				<artifactId>dockerfile-maven-plugin</artifactId>
				<version>1.3.5</version>
				<configuration>
					<repository>myRepository:10002/${project.artifactId}</repository>
					<tag>${project.version}</tag>
					<buildArgs>
						<APP_NAME>${project.artifactId}</APP_NAME>
						<APP_VERSION>${project.version}</APP_VERSION>
					</buildArgs>
				</configuration>
			</plugin>
      ```
3. `mvn dockerfile:build`

## What do you expect
Docker images should be built on the remote Docker cluster successfully.

## What happened instead
I get these error messages from conc. "Invalid RSA private key" (see full-strack trace below):

It somehow cannot find my Docker certificates and/or it seems to be that there is an invalid RSA private key having a version not being 0?

I also read [this post](https://security.stackexchange.com/questions/84327/converting-ecc-private-key-to-pkcs1-format) and I converted the EC key to a PKCS#8 private key using these commands
```
mv key.pem key.pem.ec   
openssl pkcs8 -topk8 -nocrypt -in key.pem.ec -out key.pem
```

But unfortunately, that still doesn't help - I get the very same error messages.

## Software:

- `docker version`: 
Docker Data Center 17.06:
```
Client:
 Version:      1.13.1-cs1
[...]

PR #1007, opened 2018-04-04
Support for EC algorithm used in Docker EE client bundles (#887)

chrsoo:
The PR will first try RSA and if this fails fall back to EC which is used in Docker EE client bundles, cf. #887.

Admitedly trying and catching exceptions is not the most elegant way to detect the algorithm to be used for the key spec but there does not seem to be an easy way to dynamically detect the correct algorithm. I guess this is why RSA was hard-coded in the first place...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
   </parent>
 
   <artifactId>docker-client</artifactId>
-  <version>8.11.5</version>
+  <version>8.11.6-SNAPSHOT</version>
   <packaging>jar</packaging>
   <name>docker-client</name>
   <description>A docker client</description>
@@ -26,7 +26,7 @@
     <connection>scm:git:https://github.com/spotify/docker-client</connection>
     <developerConnection>scm:git:git@github.com:spotify/docker-client</developerConnection>
     <url>https://github.com/spotify/docker-client</url>
-    <tag>v8.11.5</tag>
+    <tag>HEAD</tag>
   </scm>
 
   <developers>


chrsoo:
Hi @mattnworb, 

The CONTRIBUTING.md document says to poke you guys if there's no answer so here goes...  would you mind looking at my PR?

src/main/java/com/spotify/docker/client/DockerCertificates.java
```diff
@@ -145,8 +145,26 @@ private PrivateKey readPrivateKey(Path file) throws IOException, InvalidKeySpecE
...
+        kf = KeyFactory.getInstance(algorithm);
+        key = kf.generatePrivate(spec);
+        log.debug("Generated private key from spec using the '{}' algorithm", algorithm);
+        return key;
+      } catch (InvalidKeySpecException | NoSuchAlgorithmException e) {
+        log.debug("Tried generating private key from spec using the '{}' algorithm", algorithm, e);
+      }
+    }
+
+    throw new InvalidKeySpecException("Could not generate private key from spec");
```
davidxia: Can you also include the attempted algorithms in this exception message?

src/main/java/com/spotify/docker/client/DockerCertificates.java
```diff
@@ -145,8 +145,26 @@ private PrivateKey readPrivateKey(Path file) throws IOException, InvalidKeySpecE
   private static PrivateKey generatePrivateKey(PrivateKeyInfo privateKeyInfo) throws IOException,
           InvalidKeySpecException, NoSuchAlgorithmException {
```
davidxia: You can remove `NoSuchAlgorithmException` now.

src/main/java/com/spotify/docker/client/DockerCertificates.java
```diff
@@ -145,8 +145,26 @@ private PrivateKey readPrivateKey(Path file) throws IOException, InvalidKeySpecE
...
     final PKCS8EncodedKeySpec spec = new PKCS8EncodedKeySpec(privateKeyInfo.getEncoded());
-    final KeyFactory kf = KeyFactory.getInstance("RSA");
-    return kf.generatePrivate(spec);
+    return tryGeneratePrivateKey(spec, "RSA", "EC");
```
davidxia: nit: can you put "RSA" and "EC" at the top of the class as a static final ImmutableSet?

like `private static final Set<String> PRIVATE_KEY_ALGS = ImmutableSet.of("RSA", "EC");`

src/main/java/com/spotify/docker/client/DockerCertificates.java
chrsoo: Done.

src/main/java/com/spotify/docker/client/DockerCertificates.java
chrsoo: Done.

chrsoo:
Regarding the unit test I added a Docker EE Client Bundle that uses the EC algorithm and copy-pasted the `testReadPrivateKeyPkcs8` test. It passes but I admit to not understanding how it should work. 

As it seems to work even if I change the `key.pem`  key spec that uses the EC algorithm I highly suspect something is off.  Please have a look at the test and help by suggesting any necessary changes!

davidxia:
@chrsoo Thanks for following up. Yea, it looks like the merge has changed some stuff. Could you cherry-pick the relevant commits onto a new branch and PR that's based on the latest upstream master branch? Once you do that it'll be easier for me to review. Thanks again for contributing.

PR #876, opened 2017-08-29
Add AWS Elastic Container Service authentication support

toelen:

PR #1013, opened 2018-04-19
Add authentication support for Amazon ECR

sigpwned:
Hello,

I'm interested in using the (wonderful!) https://github.com/spotify/dockerfile-maven project to manage my docker images during the build process, but I noticed that it doesn't support ECR. Adding ECR support to this project seems like the proper first step; once it's included here, support can more easily be added to that project next.

I checked for open and closed issues and pull requests about ECR and didn't find any on this project. I apologize if they're there, but I missed them! If that's the case, please point me in the right direction.

The ECR implementation is heavily based off the GCR implementation. I've added what feels like the natural analog to the GCR features for ECR, including unit tests. Again, please let me know if I've misunderstood or overlooked some important aspect of the authorization piece!

I hope this patch looks OK and makes sense! I'm happy to make any changes you need.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,12 +129,12 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
-      <version>4.5</version>
+      <version>4.5.5</version>
     </dependency>
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpcore</artifactId>
-      <version>4.4.5</version>
+      <version>4.4.9</version>
     </dependency>
     <dependency>
       <groupId>com.github.jnr</groupId>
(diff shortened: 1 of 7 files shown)


sigpwned:
As long as ECR authentication makes its way into the plugin, it doesn't make much difference to me which patch we use one way or another.

Regarding some choices I made in my implementation, to minimize friction, I've tried to model this implementation against the GCR implementation, which is (ostensibly) already known-good and trusted code since it's integrated in the codebase. If you compare this ECR implementation to the GCR implementation, you should find them to be essentially identical. In my experience, this reduces technical risk and maintenance cost, but obviously you may feel differently!

sigpwned:
Just checking in here. Is there anything I can do to help, either on this PR or on #876?

davidxia:
@sigpwned Sorry for the delay. I was asking about whether we need the extra `Authenticator`, `EcrAuthenticator`, and `EcrAuthenticatorFactory` classes? I like how there's less code in https://github.com/spotify/docker-client/pull/876 since it tries to just implement `RegistryAuthSupplier`.

But I like how your PR is more flexible in the various ways that the user can provide ECR credentials to `ElasticContainerRegistryAuthSupplier`. #876 only allows [`AmazonECRClientBuilder.defaultClient()`](https://github.com/spotify/docker-client/pull/876/files#diff-a1a80e23b8fc23b03ece11855c2ca66aR58) and with [`AmazonECR`](https://github.com/spotify/docker-client/pull/876/files#diff-a1a80e23b8fc23b03ece11855c2ca66aR67).

If you could combine my two suggestions or make a case for the extra classes here, I'd be happy to review more closely before merging. Since I haven't used ECR at all, let me know if you've tested and the results.

tiernanscully:
@mattnworb Has #1051 made the need for this obsolete or does the underlying docker client still require code changes to enable ECR Auth support?

mattnworb:
I haven't used AWS ECR at all, so it would be great if someone else could verify this, but I think supporting other registries (like ECR) via the credential-helpers is the best long-term approach. If that works for AWS ECR then I would consider this obsolete, yeah.

acmcelwee:
@tiernanscully @mattnworb I used ECR, and I just did a quick test to wire up https://github.com/spotify/dockerfile-maven, which uses this project, and it does, indeed, seem to be solved by #1051, once I setup the appropriate ecr credHelpers entry.

davidxia:
@acmcelwee Thanks for trying it out. Closing as this issue seems to be solved.

PR #1021, opened 2018-05-15
Allow setup DockerClientBuilder with RequestEntityProcessing.

rnacher:


rnacher:
Hi!
I am sorry but right now, I don't know the full list of docker end-points
that need the Content-Length header.  I only tested the container
creational endpoint but accordingly, to docker 1.24 API documentation
<https://docs.docker.com/engine/api/v1.24/>, it seems that
Content-Length header
should be present to all POST requests with a body.  I know that it is an
old version but I think that it has not changed yet.



I am using bitbucket pipelines for my projects and I was unable to create
any docker-container through the docker API, so I perform some
investigation and at the end, I realized that the problem was related to
missing HTTP header: You can find out more details here:
https://bitbucket.org/rnacher/docker-test-pipelines/addon/pipelines/home#!/results/1

So, my first design intention was to be able to set up the client to check
if that could fix my issues without doing any major change inside the
library.

If everything is right, next step will be that one that you pointed out:
use BUFFERED for every POST request. Anyway, before doing that, we will
need to do a further investigation because I am concerned about those
endpoints that use a great amount of data like image creation. I don't like
to set BUFFERED as default and then have a performance issue.

best regards,

2018-05-20 16:49 GMT+02:00 David Xia <notifications@github.com>:

PR #1024, opened 2018-05-20
Update ISSUE_TEMPLATE.md to point questions to SO and user manual

davidxia:
We get a lot of issues that are just questions on usage. Let's direct these
to StackOverflow and the user manual so issues can be reserved for bugs and
feature requests.


davidxia:
cc @myott @caipre @danielhan @vbhavsar

mattnworb:
Might not change behavior but worth trying 👍

PR #1034, opened 2018-06-03
Add `--cpus` options to a container

GabrielSVinha:
Changes the class `HostConfig` to receive the property `cpus` when creating a  new container.

Add unit test to the specified parameters.


rwmajor2:
Can someone comment on this?  There is a CpuCount, but is it really Windows only?  It's located under the HostConfig.

PR #1041, opened 2018-06-22
Use secure password for dxia docker Hub account in tests

davidxia:
Use Travis' secure env vars.


johnflavin:
Does this mean we won't be able to run tests on local machines any more?

johnflavin:
Yeah, I suppose that would be fine.

PR #1043, opened 2018-06-26
Skip unit tests that require Docker Hub passwords

davidxia:
that are only stored encrypted in Travis.


davidxia:
@johnflavin

johnflavin:
I am getting test failures when running locally, but they aren't with the tests you changed. So I guess this looks good. 👍 

I can try to track down the test failures separately from this.

Issue #1044, opened 2018-07-02
Docker info does not return the networks under the Plugin path


alexsapran:
## Description
When using the `client.info().plugings().networks()` its always returning `[ ]`

## How to reproduce

In a host with a docker daemon (I used a file socket, but I am guessing its the same) run 
`dockerClient.info().plugins().networks()` will return `[]`

But when using the info endpoint I get this response 
```
$ curl -s --unix-socket /var/run/docker.sock "http:/v1.23/info" |jq .
.....
 "SystemStatus": null,
  "Plugins": {
    "Volume": [
      "local"
    ],
    "Network": [
      "host",
      "bridge",
      "null"
    ],
    "Authorization": null
  },
  "MemoryLimit": true,
  "SwapLimit": true,
  "KernelMemory": true,
....
```

I know I am using an older version of a docker, but the same still exists on [newer version of docker](https://docs.docker.com/engine/api/v1.37/#operation/SystemInfo).

Digging a bit more into the code I see that the marshaling is happening with the wrong JSON field. Its mapped as `Networks` were the result output is `Network` [docs ref]() 

https://github.com/spotify/docker-client/blob/14b4464fb7acec25073c8c9c564136188f94562c/src/main/java/com/spotify/docker/client/messages/Info.java#L278

## What do you expect

When calling the `dockerClient.info().plugins().networks()` I should be able to get whatever is listed in the json path `Plugins.Network`

## What happened instead

I get always `[ ]`

## Software:

- `docker version`: 
```
$ docker version
docker version
Client:
 Version:      1.11.2
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   b9f10c9
 Built:        Wed Jun  1 22:00:43 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.11.2
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   b9f10c9
 Built:        Wed Jun  1 22:00:43 2016
 OS/Arch:      linux/amd64
```

- Spotify's docker-client version: `com.spotify:docker-client:8.11.7`

johnflavin:
Looks like this is just a typo. It would be a quick and easy PR, if you want to take care of it.

I dug into the history here. Looks like I introduced this property in #426. I named the property correctly, but that doesn't matter because I didn't have the correct API key for the `JsonProperty`.
```java
@JsonProperty("Networks") private ImmutableList<String> network;
```

[...]

PR #1045, opened 2018-07-02
Fix docker info object to deserialise the networks path

alexsapran:
- Add a new fixture for the docker info response
- Add tests for the path of the network of docker info 
- Fix typo in the JSON field name for the network

Fixes: #1044


johnflavin:
The CI is failing because the indentation of your test code is four spaces per level, but this project uses two spaces per level.

Other than that, LGTM. 👍

alexsapran:
@johnflavin fixed the styling and seems to pass the CI

davidxia:
Thanks, @alexsapran.

PR #1051, opened 2018-08-04
Refactor Config-file-based auth

### LINKED ISSUE #1037: Add support for `credHelpers ` in new docker versions
Hi, I am using the `dockerfile-maven` together with `amazon-ecr-credential-helper`
[https://github.com/awslabs/amazon-ecr-credential-helper](url)
`amazon-ecr-credential-helper` support `credHelpers`. 
`With Docker 1.13.0 or greater, you can configure Docker to use different credential helpers for different registries.`

I am wondering whether docker-client could support `credHelpers`?

### LINKED ISSUE #1042: Compatibility between credsStore and passphrase.
My situation is we have two different registries. One uses the traditional passphrase, the other uses the `credsStore` which is Amazon `ecr-login`.
It seems that the `docker-client` would try to use `credsStore` to login to both of the registries, which will lead to an error.
Currently, I get around with this by creating a separate `config.json` file for ecr. But this lead to a lot of hassles.

Could you consider to change the configuration of docker-client? For example, it will use the passphrase if the passphrase part if not NULL. Otherwise, it will try to use the `credsStore`.

johnflavin:
I wanted to support the `credsHelper` values in the `.docker/config.json` file. Along the way I kind of refactored all the code around that file and using it for credentials, because it was all kind of a mess. 

* Add support for `credsHelper` values in `.docker/config.json`. Closes #1037.
* You can have different registries in the `"auth"` and `"credsHelpers"` sections, and their values will get used correctly. And if you want to auth with a registry that is not explicitly mentioned in either of those, the credential helper from `"credsStore"` will be used as a fallback. Closes #1042.
* Added auto value classes `DockerConfig` and `DockerCredentialHelperAuth` to bind JSON into objects and avoid parsing out JSON in code. The former reflects the structure of the `.docker/config.json` file. The latter is the auth response received from a credential helper on a `get`, and sent on a `store`.
* Abstracted all communication with credential helpers into its own class: `DockerCredentialHelper`. I implemented all the methods that are supported for these processes: `get`, `list`, `store`, and `erase`. (Right now only the `get` is used in any of the code, but the rest are available for users to use as they like. This might be something we want to expose through the main client? Or maybe not.) I followed the `delegate` pattern established in `DockerHost` (see #441 and code comments therein) to enable mocking out those system calls for tests.
* Related to above, I removed some existing tests that were interacting with live credential helpers. I'm not comfortable storing and erasing values in secure locations on servers, let alone on my own machine, during tests. I didn't add back any tests of the credential helpers explicitly because... it feels like I would have just been mocking something and verifying that mock. I'd be testing "did Mockito work? yep!"
* The credential helper `get` operation *is* mocked and used in some other tests that I added, since it is the one credential helper operation involved in the whole authentication chain.
* There's probably other stuff I did in here. 

Lastly, in my mind this fills some of the same role of the ECR support PRs #876 and #1013. Amazon has a docker credential helper ([awslabs/amazon-ecr-credential-helper](https://github.com/awslabs/amazon-ecr-credential-helper)), so having that installed and referenced as a `credsStore` or `credsHelper` in your docker config file should allow `docker-client` to support it with minimal code changes. I don't know the degree to which this level of support is better or worse than a full ECR `RegistryAuth` implementation, because I don't know the intricacies of their auth API, so take that for what you will. But I expect (and hope) this will be sufficient to fulfill my needs.


johnflavin:
Ping @davidxia @mattnworb for review.

src/main/java/com/spotify/docker/client/DockerConfig.java
```diff
@@ -0,0 +1,86 @@
...
+  @Nullable
+  @JsonProperty("credsHelpers")
+  public abstract ImmutableMap<String, String> credsHelpers();
+
+  @Nullable
+  @JsonProperty("auths")
+  public abstract ImmutableMap<String, RegistryAuth> auths();
+
+  @Nullable
+  @JsonProperty("HttpHeaders")
```
mattnworb: confusing that this field is capitialized differently than all the others 😕  I double-checked the format in https://github.com/docker/cli/blob/08cf36daa65e22771cc47365ff1507c156c4a459/man/docker-config-json.5.md to make sure this seems ok though 👍

src/main/java/com/spotify/docker/client/DockerConfigReader.java
```diff
@@ -56,63 +44,174 @@
...
+   * @param configPath Path to config file.
+   * @return All registry auths that can be generated from the config file
+   * @throws IOException If the file cannot be read, or its JSON cannot be parsed
+   */
+  public RegistryConfigs authForAllRegistries(final Path configPath) throws IOException {
     checkNotNull(configPath);
 
-    final Map<String, RegistryAuth> configs = parseDockerConfig(configPath).configs();
+    final DockerConfig config = MAPPER.readValue(configPath.toFile(), DockerConfig.class);
+    if (config == null) {
```
mattnworb: not very important, but can `ObjectMapper.readValue()` actually ever return `null`?

src/main/java/com/spotify/docker/client/messages/DockerCredentialHelperAuth.java
```diff
@@ -0,0 +1,55 @@
...
+
+package com.spotify.docker.client.messages;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class DockerCredentialHelperAuth {
```
mattnworb: would you mind adding a brief javadoc summary of what this class, and the other new AutoValue ones, are intended to represent? I have a hard time remembering the difference between all of these that have very similar names (`RegistryAuth`, `RegistryAuths`) etc

mattnworb:
> I didn't add back any tests of the credential helpers explicitly because... it feels like I would have just been mocking something and verifying that mock. I'd be testing "did Mockito work? yep!"

IMHO tests like what you can describe can still be useful to test that the class-under-test is behaving the way you expect based on the return value of the mocked method, but if all the class-under-test does is return the same value from the mocked method, then yeah I would agree it's not a very useful test.

src/main/java/com/spotify/docker/client/DockerConfigReader.java
johnflavin: I wrote this null check purely defensively. But, given your question, I dug into the source to figure this out. Yes, it is possible.

`readValue(File , Class<T>) → _readMapAndClose(JsonParser, JavaType)`. The latter does under some conditions set `Object result = null` and returns `result`.

johnflavin:
I wrote some javadoc on the new classes `DockerConfig`, `DockerCredentialHelper`, and `DockerCredentialHelperAuth`. I also wrote some on `RegistryAuth`, even though I didn't write that, because why not?

I also rebased on master, cleaned up a couple things, and logged the changes.

src/main/java/com/spotify/docker/client/DockerConfig.java
johnflavin: I hadn't seen that page. All the searches I did for `docker config` got swamped by the new feature: https://docs.docker.com/engine/reference/commandline/config/.

I should add several of these properties to the object. Not that I think there is any real chance someone will need them, but if we have the thing, it may as well be able to read what we know can be in there.

mattnworb:
thanks for also updating the changelog ❤️

mattnworb:
I'll try to release `8.12.0` with this in it next

mattnworb:
my gpg is screwed up so I did not get to finish releasing this today, will take another crack on Monday.

mattnworb:
`8.12.0` has been released!

PR #1055, opened 2018-08-20
Fix Info.plugins.volume deserialization

davidxia:
JSON property should be `volume` instead of `volumes`.


davidxia:
cc @spotify/docker-admins

PR #1064, opened 2018-08-30
Fix credential helpers config name

### LINKED ISSUE #1063: Typo in credential helpers config reader
Hi,

There's a typo in the config mapper for the property referencing the credential helpers.
It should be `credHelpers` instead of `credsHelper`.

The correct spelling is referenced in the documentation : https://docs.docker.com/engine/reference/commandline/cli/

As a result, I cannot use cred helpers to authenticate to google cloud registries.

A PR will follow.

ophelie:
This PR fixes #1063


ophelie:
@johnflavin np 😉

PR #1076, opened 2018-09-23
Fix flaky testInspectServiceEndpoint()

davidxia:
Sometimes DockerClient.inspectService() returns a service with
no Service.Endpoint.Spec.Ports for some reason.

Might be a race condition. Retry inspecting a service
until we get a reply with ports for up to three seconds.


davidxia:
cc @spotify/docker-admins

davidxia:
replace with a unit test instead.

PR #1081, opened 2018-10-02
Change maven-compiler-plugin source and target from 1.7 to 1.8

### LINKED ISSUE #756: start requiring Java 8?

davidxia:
Change maven-compiler-plugin source and target from 1.7 to 1.8

to prepare for Java 8.
JAR should still be compatible with JRE 1.7.

Set Travis CI JDK to `oraclejdk8`.

fixes #756

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ services:
 language: java
 
 jdk:
-  - openjdk7
+  - oraclejdk8
 
 env:
   global:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 8.14.0
+
+Not yet released
+
+- Change maven-compiler-plugin source and target from 1.7 to 1.8 to prepare for Java 8.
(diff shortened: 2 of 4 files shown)


mattnworb:
could you also add something to the changelog? I think it would be good to bump the version too, like  to `8.14` or whatever.

davidxia:
Ah that's true. I'm not using any java 8 language features yet. I'll change to 8.14 then.

diemol:
Thanks for moving it to Java 8 @davidxia!

PR #1068, opened 2018-09-07
build on jdk10, fix java 9+ incompatibilities 

mattnworb:

PR #1082, opened 2018-10-02
Debug testInspectTask

### LINKED ISSUE #756: start requiring Java 8?

davidxia:


mattnworb:
this test fails a lot for me in #1068 too.

mattnworb:
is there much of a point to asserting the data that the docker daemon is returning here anyway? I’d rather see a test that checks that the data returned by this call is correct given some mock response from the daemon, but not something that presumes some correctness or behavior inside the daemon.

davidxia:
agreed. I'll replace with a unit test.

davidxia:
replaced by https://github.com/spotify/docker-client/pull/1090

PR #1093, opened 2018-10-08
Fix SystemCredentialHelperDelegate

davidxia:
* fix error when parsing multi-line JSON string
* allow DockerCredentialHelperAuth.ServerURL to be nullable

### DIFF
--- a/src/main/java/com/spotify/docker/client/SystemCredentialHelperDelegate.java
+++ b/src/main/java/com/spotify/docker/client/SystemCredentialHelperDelegate.java
@@ -1,0 +1,137 @@
+/*-
+ * -\-\-
+ * docker-client
+ * --
+ * Copyright (C) 2016 - 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.docker.client;
+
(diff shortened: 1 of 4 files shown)


davidxia:
cc @spotify/docker-admins

src/main/java/com/spotify/docker/client/SystemCredentialHelperDelegate.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The default credential helper delegate.
+ * Executes each credential helper operation on the system.
+ */
+@SuppressFBWarnings("SIC_INNER_SHOULD_BE_STATIC_ANON")
```
mattnworb: is this still needed?

src/main/java/com/spotify/docker/client/SystemCredentialHelperDelegate.java
```diff
@@ -0,0 +1,139 @@
...
+
+  @Override
+  public int store(final String credsStore, final DockerCredentialHelperAuth auth)
+      throws IOException, InterruptedException {
+    final Process process = exec("store", credsStore);
+
+    try (final Writer outStreamWriter =
+             new OutputStreamWriter(process.getOutputStream(), StandardCharsets.UTF_8)) {
+      try (final BufferedWriter writer = new BufferedWriter(outStreamWriter)) {
+        writer.write(mapper.writeValueAsString(auth) + "\n");
```
mattnworb: I didn't notice this in the prior version of the code, but I wonder if all of these `"\n"` strings should instead be `System.lineSeparator()` to avoid assuming the code only runs on *nux systems.

src/main/java/com/spotify/docker/client/SystemCredentialHelperDelegate.java
davidxia: It is. Otherwise findbugs errors on this class saying it could be a static inner class. :(

src/main/java/com/spotify/docker/client/SystemCredentialHelperDelegate.java
davidxia: good point. I'll change.

PR #1097, opened 2018-10-18
RFC: Add operation for streaming container stats

dsboger:
Hello!

I have a use-case where it would be convenient to have access to the stream of container stats as provided by the Docker daemon, so I've sketched a new `DockerClient.stats` operation. It is a rip-off of the `pull` operation, where a stream of objects is read from the response and passed to a user-provided handler. I've duplicated some code inside `DefaultDockerClient`, but if there is interest, I can try to refactor it.

Would you have a look at it?

Thanks for your efforts!


src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2736,14 +2754,63 @@ public Void call() throws DockerException, InterruptedException, IOException {
...
+      try {
+        response.close();
+      } catch (ProcessingException e) {
+        // ignore, thrown by jnr-unixsocket when httpcomponent try to read after close
+        // the socket is closed before this exception
+      }
+    }
+  }
+
+  private static class StatsResponseTailReader implements Callable<Void> {
```
mattnworb: this class looks really similar to ProgressResponseTailHeader, do you think a common class could be extracted between the two to avoid the duplication?

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2383,6 +2385,22 @@ public ContainerStats stats(final String containerId)
...
 
+  public void stats(final String containerId, final StatsHandler handler)
+      throws DockerException, InterruptedException {
+    final WebTarget resource = resource().path("containers").path(containerId).path("stats");
+
+    try {
+      requestAndTail(GET, handler, resource, resource.request(APPLICATION_JSON_TYPE));
+    } catch (DockerRequestException e) {
+      switch (e.status()) {
+        case 404:
```
mattnworb: could you move this check into the `requestAndTail` method itself? That would then seem to remove the need to throw an exception only to catch it and re-throw it a single method up the stack

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2383,6 +2385,22 @@ public ContainerStats stats(final String containerId)
     }
   }
 
+  public void stats(final String containerId, final StatsHandler handler)
```
mattnworb: I think this is missing an `@Overrides` annotation (I am surprised the build does not fail on that)

dsboger: Oops! Fixed.

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
dsboger: I have been thinking about it and saw that there is something I don't completely understand: why is a new thread created in `tailResponse`? Is it because calls to `images/create` API cannot be cancelled/interrupted? I've just copied that method but I'm not sure it applies to `stats`, as calls to the `containers/{id}/stats` API must be cancellable/interruptible.

If my thought is correct, maybe I could just call `stream.tail` directly inside `tailResponse`, instead of creating a future. What do you think? Am I missing something?

Thanks again!

src/main/java/com/spotify/docker/client/DefaultDockerClient.java
dsboger: It is doable, but then I would need to pass the `containerId` to `requestAndTail` in order to create the `ContainerNotFoundException`. Personally, I think it would smell a bit, as `requestAndTail` seems more "HTTP-focused", while `stats` seems "API-focused". That's not a strong opinion though. What do you think?

PR #1104, opened 2018-11-06
Add ability to query secrets using filter criteria

llowrey:
Added `listSecrets(Criteria criteria)` to `DockerClient `and `DefaultDockerClient`. Essentially copied the code used by `listConfig` and `Config.Criteria`. I looked for unit test code for Secret and Config but didn't find any so I didn't include any tests. I have verified this functionality with my local swarm.

### DIFF
--- a/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
+++ b/src/main/java/com/spotify/docker/client/DefaultDockerClient.java
@@ -2602,7 +2602,7 @@ public class DefaultDockerClient implements DockerClient, Closeable {
   @Override
   public List<Secret> listSecrets(final Secret.Criteria criteria)
       throws DockerException, InterruptedException {
-    assertApiVersionIsAbove("1.24");
+    assertApiVersionIsAbove("1.25");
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/spotify/docker/client/DefaultDockerClient.java
```diff
@@ -2599,6 +2599,38 @@ public VolumeList listVolumes(ListVolumesParam... params)
     return request(GET, SECRET_LIST, resource, resource.request(APPLICATION_JSON_TYPE));
   }
 
+  @Override
+  public List<Secret> listSecrets(final Secret.Criteria criteria)
```
davidxia: Does the API accept more than one filter at a time? If so, maybe this should take var args Criteria.

llowrey: The API takes only one filter, just like for Config.

llowrey:
When I try to run tests locally, "testBuildInterruption" never completes. Any idea why?

PR #1108, opened 2018-11-11
maven-shade-plugin 3.2.0

sullis:


sullis:
reference: https://maven.apache.org/plugins/maven-shade-plugin/

PR #1111, opened 2018-11-12
* relocate remaining packages that are included in shaded jar

DavidHaunsch:
As mentioned in https://github.com/spotify/docker-client/pull/1096#issuecomment-436764107, I'm trying here to relocate all remaining packages that are in the shaded jar but not relocated yet.
This will solves problems where a docker-client doesn't get the version from inside the shaded jar but the one that appears first on classpath.


DavidHaunsch:
Any chance this PR get's merged?
@davidxia (just to mention one who approved a PR lately) or someone else?

Issue #875, opened 2017-08-26
Support windows named pipe


philippe-granet:
## Feature request

Support Windows named pipe like this :
```java
final DockerClient docker = new DefaultDockerClient("npipe:////./pipe/docker_engine");
```

Example in fabric8io docker maven plugin :
https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/access/DockerConnectionDetector.java#L123

Extract from [Docker for Windows Faq](https://docs.docker.com/docker-for-windows/faqs/)
```
How do I connect to the remote Docker Engine API?
You might need to provide the location of the Engine API for Docker clients and development tools.

On Docker for Windows, clients can connect to the Docker Engine through a named pipe: npipe:////./pipe/docker_engine, or TCP socket at this URL: tcp://localhost:2375.

This sets DOCKER_HOST and DOCKER_CERT_PATH environment variables to the given values (for the named pipe or TCP socket, whichever you use).
```

mattnworb:
Thanks for providing a link on how the named pipe can be accessed as a File, that is useful to know 👍

davidxia:
So would things just work if DefaultDockerClient just stripped "npipe:////./pipe/docker_engine" down to "//./pipe/docker_engine"?

hazsetata:
I wanted to give this a try, here is what I was able to do so far: https://github.com/hazsetata/docker-client/commit/446ef75c9ae9ad43c840f4de07cd4f3a30e9e7f7

It uses a RandomAccessFile and wraps it in a Socket (supporting only the bare minimal things that are needed to make this work). I then created a SocketFactory very similar to the Unix socket implementation's and modified DockerHost / DefaultDockerClient to support it. 

It compiles, runs but some of the tests hang. I'm not sure yet what causes that, one possibility that I'll have to check is that because the code uses a file, the read() methods do not timeout (with a real Socket they would). If supporting that is a requirement, I'm not sure yet how to proceed.

Any comments / ideas would be very welcome!

php-coder:
As far I understand, this issue still hasn't been fixed. Could someone reopen it, then?

jan-zajic:
[...]

PR #1114, opened 2018-11-20
Fix Issue #875 - windows named pipes

jan-zajic:
This PR added support for Docker for Windows - https://docs.docker.com/docker-for-windows/install/ that supports only Windows 10 1607 Anniversary Update or later. This Docker in default configuration used Windows duplex named pipe.

### DIFF
--- a/src/main/java/com/spotify/docker/client/npipe/NamedPipeSocket.java
+++ b/src/main/java/com/spotify/docker/client/npipe/NamedPipeSocket.java
@@ -1,0 +1,308 @@
+/*-
+ * -\-\-
+ * docker-client
+ * --
+ * Copyright (C) 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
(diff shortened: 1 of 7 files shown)


jan-zajic:
It would be nice to add Travis CI Job on Windows 10 and run integration tests that are in sources, but i do not know if it is possible.

src/main/java/com/spotify/docker/client/npipe/NamedPipeSocket.java
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  @Override
+  public void bind(SocketAddress bindpoint) throws IOException {
+    throw new SocketException("Bind is not supported");
+  }
+
+  @Override
+  public InetAddress getInetAddress() {
+    return null;
```
davidxia: Would throwing `UnsupportedOperationException` in these methods be clearer instead?

davidxia:
Could you also update the CHANGELOG?

mohitseth-nitrr:
@jan-zajic nicely done !!!

jan-zajic:
I batlle tested some advanced operations, like coping files to container and execute. it gets stuck during copy tar archive upload. So i disable pooling for this type of connection. And when disabled everything workes fine. We have same problems and same solution at https://github.com/fabric8io/docker-maven-plugin, i forgot it first time.

PR #1119, opened 2018-11-30
Fix #1118 System property http.nonProxyHosts is processed improperly

### LINKED ISSUE #1118: System property http.nonProxyHosts is not processed properly
**BUG REPORT**:

## Description

In case system property *http.nonProxyHosts* specified like *-Dhttp.nonProxyHosts=site1|site2|ipmask1|ipmask2|etc* it's processed properly. However in case it's specified using quoted string like *-Dhttp.nonProxyHosts="site1|site2|ipmask1|ipmask2|etc"*, method com.spotify.docker.client.DefaultDockerClient#updateProxy parses the property into the following collection internally:
- "site1
- site2
- ipmask1
- ipmask2
- etc"
So *site1* and *etc* are not matched as sites with non-proxy connection

## How to reproduce

java -Dhttp.proxyHost=<ipaddress> -Dhttp.proxyPort=<portnumber> -Dhttp.nonProxyHosts="127.0.0.1|localhost" -jar java-application-that-uses-docker-client-which-connects-to-localhost-via-http.jar ...

## What do you expect

ClientConfig is set up to use direct connection

## What happened instead

ClientConfig is set up to use proxy, because *localhost* doesn't match *localhost"*

## Software:

- `docker version`: N/A
- Spotify's docker-client version: 8.11.7

## Full backtrace

```text
N/A
```

alrega:


src/test/java/com/spotify/docker/client/DefaultDockerClientUnitTest.java
```diff
@@ -209,6 +209,16 @@ public void testHostWithNonProxyHost() {
       assertThat((String) client.getClient().getConfiguration()
                       .getProperty("jersey.config.client.proxy.uri"),
               isEmptyOrNullString());
+      final DefaultDockerClient client1 = DefaultDockerClient.builder()
```
davidxia: Just curious what these additional lines are testing?

Is it easy to add a test for the quote-stripping logic above?

alrega: Hello David Xia,
Before making changes in docker-client I encountered the problem that HTTP requests to localhost / 127.0.0.1 were sent via proxy. It was quite unexpected for me to learn so.
So, these test ensures that requests to 127.0.0.1 and localhost never go through proxy server..

davidxia:
@alrega 8.15.2 should be available here soon https://search.maven.org/artifact/com.spotify/docker-client/8.15.2/jar

alrega:
Great. Thank you so much.

PR #1145, opened 2019-03-05
Read configuration file from directory supplied in DOCKER_HOST

albanf:
This is related to an issue when using dockerfile-maven and Jenkins (https://github.com/spotify/dockerfile-maven/issues/191)
If the environment variable DOCKER_HOST is defined, docker.config and certificates should be loaded from this directory.


janario:
Any update here? 

I'm facing the same issue, when I try to build it on jenkins

It would be great to have this :)

albanf:
@davidxia, you think you could check this PR and include it in the next release?

PR #1148, opened 2019-03-20
Update README.md

cciollaro:
just a humble readme fix as I pass through

the `~`'s were rendering incorrectly as a strikethrough in the readme. and travis has migrated their `.org`s to `.com`s.


cciollaro:
will be here if ya want it :)

PR #1157, opened 2019-05-13
Add Hardlink Support

hohle:
Adds support for identifying hardlinks when building a `CompressedDirectory` to match the functionality of docker-go.

I have a use case where there are several hundred MBs of hard linked files that are added to a docker image. The official CLI identifies the files which share the same inode and send the build context appropriately resulting in a layer where links to the same file match what was in the context.

While integrating `dockerfile-maven` I found the hard links were not being respected resulting in layers which were significantly larger that those created with the CLI.

I've added a new unit test for the use case as well as confirmed locally through the maven plugin.


hohle:
I'll need to refactor the unit test to account for the differences in the order of directory entries on different file systems. In the mean time, feel free to leave any feedback while I update the PR.

jvanzyl:
@hohle I needed to fix hardlink support and noticed you did already. So I've forked the docker-client and dockerfile-maven over [here](https://github.com/jvanzyl) and integrated your change for hardlinks. They have both been released to Maven Central using the `ca.vanzyl` groupId instead of `com.spotify`.

https://repo.maven.apache.org/maven2/ca/vanzyl/docker-client/8.16.1/

and

https://repo.maven.apache.org/maven2/ca/vanzyl/dockerfile-maven-plugin/1.4.14/

PR #1160, opened 2019-05-29
implement docker service rollback command

taclano:


taclano:
Poke, Don't understand why it's failing

